PR #55, opened 2015-01-05
Fixed NPE when unsuccessful response is returned from server when buildi...

kpiwko:
...ng the image.


lordofthejars:
Thank you so much :)

Issue #66, opened 2015-01-20
Allow dynamic port binding


lordofthejars:
Allow dynamic port binding so the bind port to an exposed port is assigned automatically from random port. In this way we will be able to run tests in parallel without having port collisions.

lordofthejars:
Dynamic port binding is done by setting one property to true (`PublishAllPorts`). 
To implement this feature we are going to add a new possible value in  `PortBinding` property. Instead of passing an array you can also pass a string with `dynamic`.

```
portBindings: dynamic
```

This means we should take care of having an `instanceof` when getting port binding between an array or an string.

lordofthejars:
Thinking better because `PublishAllPorts`is a property that you can set in configuration file,  the developer is free to not specify any portbind and set this property to true.

PR #72, opened 2015-01-24
resolves issue #66

lordofthejars:
Resolves issue #66 by not trying to remap ports when no port binding. Probably this means that port binding is dynamically assigned by docker server.

### DIFF
--- a/docker/src/main/java/org/arquillian/cube/impl/client/container/RemapContainerController.java
+++ b/docker/src/main/java/org/arquillian/cube/impl/client/container/RemapContainerController.java
@@ -42,29 +42,33 @@ public class RemapContainerController {
 
         Binding binding = BindingUtil.binding(cube.configuration());
 
-        ContainerDef containerConfiguration = container.getContainerConfiguration();
-        List<String> portPropertiesFromArquillianConfigurationFile = filterArquillianConfigurationPropertiesByPortAttribute(containerConfiguration);
+        if (binding.arePortBindings()) {
 
-        Class<?> configurationClass = container.getDeployableContainer().getConfigurationClass();
-        List<PropertyDescriptor> configurationClassPortFields = filterConfigurationClassPropertiesByPortAttribute(configurationClass);
+            ContainerDef containerConfiguration = container.getContainerConfiguration();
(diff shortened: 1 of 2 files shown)


aslakknutsen:
Does this mean we can't do automatic remap of dynamic ports?

lordofthejars:
Remapping is done changing configured AS port by setting the port binding. What we are doing is: see if exposed port is equal to AS port and if it is the case then we change to port bind port. The problem is that AS port is changed before container is created (@BeforeSetup) so we can change that value in AS configuration object. The problem is that port is dynamic and at that time we don't know the port that is going to be assigned, and we cannot do an inspect to the docker container to know the exposed ports because docker container has not started yet.
I think that if you are using dynamic ports we will not be able to do automatic remap. What I think is that dynamic ports should be used for none AS containers, but if we want to use it with AS containers as well we would need a way to change the configuration of Arquillian adapter just after docker container is created and we can inspect the assigned ports.

aslakknutsen:
Yeah, that's what I was thinking. 

Need to see if we can figure out some kinda 'lazy configure' here.

rbattenfeld:
I don't think that dynamic port binding is for non AS containers only valid.

Currently, I am working on a docker proxy service that can probably solve situations like this. I will see the weekend how far I come. Dispatching request is trivial and works fine. 

Besides this, is moving of the lifecycle handling of docker containers before @BeforeSetup too difficult? I believe that this is the right approach for a service like docker anyway. I was able to do this even with my poor arquillian internal know how.

lordofthejars:
Yes first of all now I can't find where I read that PublishAllPorts creates random ports for exposed ones, but assuming that this is right, I am not sure how to modify container configuration before all things.

About the question of events I would prefer that @aslakknutsen answer it because a change on an event can be a problem in how Arquillian works.

rbattenfeld:
I did this already here: https://github.com/rbattenfeld/arquillian-cube/tree/dynamic-ports/docker/src/main/java/org/arquillian/cube/impl

Absolutely, Aslak is the expert especially for this.

I wanted to point out that without moving of the docker life cycle management, cube is limited in supporting docker features heavily. It depends on the used arquillian remote container.

aslakknutsen:
I had an idea on how to improve on the lifecycle change... but the 'solution' has escaped me at the moment. Let me re read some code and see if it magically reappear.. :)

Issue #81, opened 2015-01-29
Binding polling does not correctly set host IP


tnine:
When choosing a polling strategy to ensure that startup is successful, it does not correctly discover the IP to ping.  It chooses the docker image IP, NOT the ip of the host where the port will be exposed.  As a result, the ping will never succeeded for users of boot2docker.   I propose the following change.

Make the ping strategy only ping opened ports on the host docker instance ip, not the image ip and it's ports.

lordofthejars:
Hi yes we know this issue, in fact is related with https://github.com/arquillian/arquillian-cube/issues/52

Let me explain the problem on doing in this way, When you do a port binding docker server just opens the port when the container is just started, this doesn't means that the exposed port (the port of server in our case) is available. So when you execute the ping it simply returns saying that the port is ready to serve incoming messages but this is not true because the server (tomcat, wildfly, ...) is not booted up yet, so the ping returns an ok although in reality the server port is not. And this makes that when arquillian tries to deploy the file it returns that the server is not ready to deploy that file. So we decided to use the gateway as a workaround.

Currently we are improving docker-client so when next version will be released we will be able to execute commands inside the docker container, and then we will be able to ping directly inside the container by calling `ss` command. Until then we know that there is a limitation but we cannot do anymore for now. :)

tnine:
@lordofthejars  I've looked at the strategy PollingAwaitStrategy and that makes sense.  I assumed that was the case, we've done a similar style for process forking tomcat, Cassandra and Elastic Search.  I was thinking of performing the following.

In the AwaitStrategyFactory, where we produce the PollingStrategy, modify the PollingStrategy to perform the following.

If the user has defined a http client, then we know this means we're connecting to an API.  This ip could be localhost, or a remote ip, or a virtual machine, it won't matter, it's routable and can be seen by the client, otherwise our client would completely fail.

Currently we use the IP that has been returned from docker as the guest ip, instead use the IP from the client URI configured in the plugin.  I propose we use the client ip/hostname to verify, then only verify the ports that have been marked as portBindings in the yaml.  Clients would only "care" about these ports, since they've mapped to the host system ports.

I've implemented this locally, and it works well.  This approach introduces the requirement that ports your application utilizes about must be mapped via the portBindings.  This would then work for all all possible deployment scenarios.  Local, virtual machine, or remote.

Thoughts?

tnine:
[...]

Issue #79, opened 2015-01-29
Unable to run a test more than once when using a dependent service


tnine:
I have the following arquillian.xml file.

```

<arquillian xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xmlns="http://jboss.org/schema/arquillian"
  xsi:schemaLocation="http://jboss.org/schema/arquillian
  http://jboss.org/schema/arquillian/arquillian_1_0.xsd">

  <extension qualifier="docker">
      <property name="serverVersion">1.16</property>
      <property name="serverUri">https://192.168.59.105:2376</property>
  
     <property name="dockerContainers">
        cassandra:
          image: spotify/cassandra
          exposedPorts: [9160/tcp]
          await:
            strategy: polling
          env: []
          portBindings: [9160/tcp]
     </property>
    <property name="autoStartContainers">cassandra</property>
    <property name="shouldAllowToConnectToRunningContainers">true</property>
  </extension>

</arquillian>
```

When I run my trivial test below, it runs the first time

```
@RunWith( Arquillian.class)
public class StartupTest {

    @Test
    public void iStartCassandra(){
        assertTrue( "stuff done", true);
    }
}
```

Subsequent invocation results in the following error.

```
Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 1.609 sec <<< FAILURE! - in StartupTest
StartupTest  Time elapsed: 1.609 sec  <<< ERROR!
org.arquillian.cube.spi.CubeControlException: Could not create cassandra
	at org.arquillian.cube.spi.CubeControlException.failedCreate(CubeControlException.java:19)
	at org.arquillian.cube.impl.model.DockerCube.create(DockerCube.java:69)
	at org.arquillian.cube.impl.client.CubeLifecycleController.create(CubeLifecycleController.java:14)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:483)
	at org.jboss.arquillian.core.impl.ObserverImpl.invoke(ObserverImpl.java:94)
	at org.jboss.arquillian.core.impl.EventContextImpl.invokeObservers(EventContextImpl.java:99)
[...]

PR #82, opened 2015-01-29
First pass at updating the "create" command to be idempotent

tnine:
Refactored endpoint and version to be maven properties.  This allows users of boot2docker to override in their maven profile for testing.  This PR will address issues #79  and #81.

### DIFF
--- a/docker-server-stub/pom.xml
+++ b/docker-server-stub/pom.xml
@@ -1,0 +1,32 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.arquillian.cube</groupId>
+        <artifactId>arquillian-cube-parent</artifactId>
+        <version>1.0.0.Final-SNAPSHOT</version>
+    </parent>
+    <artifactId>docker-server-stub</artifactId>
+
+    <properties>
+        <version.spark>2.1</version.spark>
+        <version.jackson>2.5.0</version.jackson>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.sparkjava</groupId>
+            <artifactId>spark-core</artifactId>
+            <version>${version.spark}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <version>${version.jackson}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.github.docker-java</groupId>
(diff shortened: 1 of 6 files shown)


tnine:
This isn't ready to ack yet.  Rather, a work in progress where we can comment on the progress of the the issue.

lordofthejars:
@aslakknutsen we have a first approach of reusing an existing container if it is already started. Now I think we must discuss which is the best strategy to follow. Discard to use an already started container automatically, by using a configuration parameter, ...

It is true that reusing a container it should not be an automatic operation because it can be there something from previous execution. Also let suppose that the already started container have a link to another container, and that second container has just stopped correctly, then we have a problem because the container that is being reused is pointing out to a not created container.

It seems that reusing a container is more complicated that we though (or at least me) initially.

tnine:
So I believe addressed #81 and #79.  However, the 2 original tests from CubeControllerTest.should_enrich_test_with_cube* are failing.  I'm not sure if these fail in the current master with a linux distro, or if they're broken with boot2docker.  I was able to use this version in my dev environment to deploy to tomcat with boot2docker.  Can someone test the latest master in their environment and let me know if the tests are passing?

tnine:
So I believe addressed #81 and #79. However, the 2 original tests from CubeControllerTest.should_enrich_test_with_cube* are failing. I'm not sure if these fail in the current master with a linux distro, or if they're broken with boot2docker. I was able to use this version in my dev environment to deploy to tomcat with boot2docker. Can someone test the latest master in their environment and let me know if the tests are passing?

lordofthejars:
I have just run the tests and they run.

ftest/src/test/resources/arquillian.xml
```diff
@@ -62,7 +65,7 @@
 
     <container qualifier="tomcat_dockerfile">
         <configuration>
-            <property name="host">localhost</property>
+            <property name="host">${docker.tomcat.host}</property>
```
aslakknutsen: host isn't really needed here. It will be auto updated by Cube anyway.

ftest/src/test/resources/arquillian.xml
```diff
@@ -70,14 +73,14 @@
     </container>
     <container qualifier="tomcat_default">
         <configuration>
-            <property name="host">localhost</property>
+            <property name="host">${docker.tomcat.host}</property>
```
aslakknutsen: host isn't really needed here. It will be auto updated by Cube anyway.

ftest/src/test/resources/arquillian.xml
```diff
@@ -70,14 +73,14 @@
...
             <property name="user">admin</property>
             <property name="pass">mypass</property>
         </configuration>
     </container>
     <container qualifier="tomcat">
         <configuration>
-            <property name="host">localhost</property>
+            <property name="host">${docker.tomcat.host}</property>
```
aslakknutsen: host isn't really needed here. It will be auto updated by Cube anyway.

tnine:
Changing this into 2 pull requests.  You can find them here.

tnine:
See #84 for fixing the docker ip issue

Issue #52, opened 2014-12-14
Execute Test On Remote Docker Host ConnectException: Connection timed out: connect


dimitrichkhartishvili:
My remote docker host installed on centos 7 in hyper-v vm. configured like in arquillian cube git documentation.
-H tcp://{hostIp}:2375 -H unix:///var/run/docker.sock (for remote connections)
firewalld disabled.
telnet works from windows hostIp:2375
started docker container ports exposed (8080,9990) works to.
everything works.

running tests from windows machine tests not executed.
Exception

.net.ConnectException: Connection timed out: connect
javax.ws.rs.ProcessingException
	at java.net.DualStackPlainSocketImpl.connect0(Native Method)
	at java.net.DualStackPlainSocketImpl.socketConnect(DualStackPlainSocketImpl.java:79)
	at java.net.AbstractPlainSocketImpl.doConnect(AbstractPlainSocketImpl.java:345)
	at java.net.AbstractPlainSocketImpl.connectToAddress(AbstractPlainSocketImpl.java:206)
	at java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:188)

after some research I found arquillian executes tests again  
         http://172.17.42.1:8080/7c832426-9615-4e47-bf74-458813015fa2

ip 172.17.42.1 is internal docker ip which is not visible from LAN.

but correct url must be

     http://{hostIp}:8080/7c832426-9615-4e47-bf74-458813015fa2

rbattenfeld:
Ok, I am not sure now:-) #64 is closed but gives for this here some ideas... But to make progress, what is now open needs to be discussed?

lordofthejars:
Well we need to implement a call to ss inside container with a while to
know when port is ready to receive incoming messages
El ds., 24 de gen., 2015 a les 17.35 rbattenfeld <notifications@github.com>
va escriure:

rbattenfeld:
Hi Alex, Aslak
I worked on this over the last couple of days. I worked mainly of wildfly execution over published ports. Today, I think I made also Tomcat executable. Here is the problem of the jmx stuff. But I think it works now; a new property solves this, I believe. 

Here is the link the the repo: https://github.com/rbattenfeld/arquillian-cube/tree/dynamic-ports

I prepared the profiles in the ftest module:
- wildfly_dynamic_ports
- tomcat_dynamic_ports

You can see that the configuration is slightely different. For example wildfly_dynamic_ports:

[...]

PR #84, opened 2015-01-30
Issue 81

tnine:
Resolves issue #81. Note that this temporarily breaks the polling strategy. We now poll the docker host IP, NOT the instance ip itself.  As a result, the ports report open even when they are not.  This issue is dependent on a fix for #52 to work correctly with a polling strategy.  A workaround is to use the wait strategy "sleeping" so that users of external docker hosts or boot2docker can run containers.  This strategy is less effective than a poll, and either introduces extra wait, or may not wait long enough.

### DIFF
--- a/docker/src/main/java/org/arquillian/cube/impl/util/BindingUtil.java
+++ b/docker/src/main/java/org/arquillian/cube/impl/util/BindingUtil.java
@@ -1,5 +1,6 @@
 package org.arquillian.cube.impl.util;
 
+import java.net.URI;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -7,6 +8,7 @@ import java.util.Map.Entry;
 import org.arquillian.cube.impl.docker.DockerClientExecutor;
 import org.arquillian.cube.spi.Binding;
 
+import com.github.dockerjava.api.DockerClient;
 import com.github.dockerjava.api.command.InspectContainerResponse;
 import com.github.dockerjava.api.model.ExposedPort;
 import com.github.dockerjava.api.model.HostConfig;
@@ -20,12 +22,21 @@ public final class BindingUtil {
     }
 
     public static Binding binding(DockerClientExecutor executor, String cubeId) {
-        InspectContainerResponse inspectResponse = executor.getDockerClient().inspectContainerCmd(cubeId).exec();
+        InspectContainerResponse inspectResponse = executor.getDockerClient().inspectContainerCmd( cubeId ).exec();
 
         HostConfig hostConfig = inspectResponse.getHostConfig();
+
+        /**
+         * This isn't actually the gateway IP of the host, it's the ip of the gateway within the docker runtime,
+         * which may not be accessible to us.  We need the IP the client uses, that will be routable, and will
+         * allow us to validate the ports in later phases
         String gatewayIp = inspectResponse.getNetworkSettings().getGateway();
 
-        Binding binding = new Binding(gatewayIp);
+         **/
+
+        final String dockerIp = executor.getDockerUri().getHost();
+        Binding binding = new Binding(dockerIp);
+
         for (Entry<ExposedPort, com.github.dockerjava.api.model.Ports.Binding[]> bind : hostConfig.getPortBindings()
                 .getBindings().entrySet()) {
             com.github.dockerjava.api.model.Ports.Binding[] allBindings = bind.getValue();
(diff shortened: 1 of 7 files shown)


ftest/src/test/resources/arquillian.xml
```diff
@@ -14,15 +17,15 @@
...
+              env: [TOMCAT_PASS=mypass, JAVA_OPTS=-Djava.rmi.server.hostname=${docker.tomcat.host}  -Dcom.sun.management.jmxremote.rmi.port=8088 -Dcom.sun.management.jmxremote.port=8089 -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.authenticate=false]
+              portBindings: [8089/tcp, 8088/tcp, 8081->8080/tcp]
             tomcat:
               image: tutum/tomcat:7.0
               exposedPorts: [8089/tcp]
               await:
                 strategy: polling
-              env: [TOMCAT_PASS=mypass, JAVA_OPTS=-Dcom.sun.management.jmxremote.port=8089 -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.authenticate=false]
-              portBindings: [8089/tcp, 8081->8080/tcp]
+              env: [TOMCAT_PASS=mypass, JAVA_OPTS=-Djava.rmi.server.hostname=${docker.tomcat.host}  -Dcom.sun.management.jmxremote.rmi.port=8088 -Dcom.sun.management.jmxremote.port=8089 -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.authenticate=false]
```
aslakknutsen: I'm not clean on why we need this?

aslakknutsen: I retract my previous statement.. nothing to see here :)

docker/src/main/java/org/arquillian/cube/impl/util/BindingUtil.java
```diff
@@ -20,12 +20,21 @@ private BindingUtil() {
...
+         * This isn't actually the gateway IP of the host, it's the ip of the gateway within the docker runtime,
+         * which may not be accessible to us.  We need the IP the client uses, that will be routable, and will
+         * allow us to validate the ports in later phases
         String gatewayIp = inspectResponse.getNetworkSettings().getGateway();
 
-        Binding binding = new Binding(gatewayIp);
+         **/
+
+        final String dockerIp = executor.getDockerUri().getHost();
+        Binding binding = new Binding(dockerIp);
```
lordofthejars: Understood so what we are using here is the ip of the docker server. client(ip1) - docker(ip2) - contianer(ip3) which in my schema is the ip2. So in ping await strategy we are going to ping ip2 and the port defined in port binding.  So if dockerserver is in localhost then it is ok, if it is remote ip then we need another wait strategy which executes an `ss` command and return when the port is available.  Is it correct or ping strategy will work in case of remote host (like boot2docker) as well?

lordofthejars:
Should I accept this PR and https://github.com/arquillian/arquillian-cube/pull/85/files right?

docker/src/main/java/org/arquillian/cube/impl/util/BindingUtil.java
lordofthejars: Sorry to my question but I have a headache with ip and ports in docker. I mean originally we used docker server ip and we had some problems with polling strategy because it returned true immediately not when the container port was ready.

lordofthejars: Also another question should we add some special code in case of using unix socket? Something like if it is unix socket then URI is localhost or getNetworkSewttings().getIp() or something like that.

lordofthejars:
Hi I have accepted the PR but when I try to run hello world test from tomcat next exception is thrown:

```
Caused by: javax.ws.rs.ProcessingException: Already connected
  at org.glassfish.jersey.client.ClientRuntime.invoke(ClientRuntime.java:255)
  at org.glassfish.jersey.client.JerseyInvocation$2.call(JerseyInvocation.java:683)
  at org.glassfish.jersey.internal.Errors.process(Errors.java:315)
  at org.glassfish.jersey.internal.Errors.process(Errors.java:297)
  at org.glassfish.jersey.internal.Errors.process(Errors.java:228)
  at org.glassfish.jersey.process.internal.RequestScope.runInScope(RequestScope.java:424)
```

I am going to investigate but any hints?

aslakknutsen:
@lordofthejars it's caused by a underlying SSL issue; https://github.com/arquillian/arquillian-cube/pull/91

lordofthejars:
Cool merged. This afternoon I will close some issues as well.

PR #86, opened 2015-01-31
Minor corrections to README wording

jamesnetherton:


jamesnetherton:
Good job with arquillian-cube. I was reading trough the README yesterday and thought some minor improvements could be made. Hence this PR.

lordofthejars:
Thank you so much for these fix. Documentation is also so important and has the same value as any bug fix. Thanks.

Issue #78, opened 2015-01-29
boot2docker fails without additional configuration [workaround found]


tnine:
When using boot2docker additional configuration must be performed in order to allow the java client to successfully negotiate the tls protocol.  This was tested with 1.4.1.  Steps to reproduce failure.

1. Execute boot2docker start
1. Set the following properties in the configuration
```
  <extension qualifier="docker">
      <property name="serverVersion">1.16</property>
      <property name="serverUri">https://192.168.59.104:2376</property>
     <!-- container properties --> 
  </extension>

```
1. Attempt to start a container, which will fail

Per the documentation of the client [outlined here](https://github.com/docker-java/docker-java#programmatic), you can create the following file 

```
~/.docker.io.properties
```

With the following content

```
docker.io.url=https://192.168.59.104:2376 [Replace with your docker vm]
docker.io.version=1.16 [Ensure this is the api version you wish to use]
docker.io.dockerCertPath=/Users/apigee/.boot2docker/certs/boot2docker-vm [ Replace with the cert path given to you by the boot2docker command]
```

You are now able to use this with boot2docker.  The Java client will read the properties from the system on initialization.

lordofthejars:
It is important that you have been able to run Arquillian cube with boot2docker because we were not sure about this because we don't have a computer which requires boot2docker. About what you mention, we are rewriting some parts of the docker-client, when we finish it and we have full support for SSL then we will modify arquillian configuration file so you can set these parameters from that file instead of creating a new one.

Thank you very much for using Arquillian Cube.

PR #91, opened 2015-02-01
Default to http, not https based Docker serverUri

aslakknutsen:


lordofthejars:
Oh Cool I have been debugging all the code and I have not seen that it was working with https :( sorry.

tnine:
See #78 for a way to run in the current framework with https. Requires additional external configuration.

Issue #110, opened 2015-02-20
shouldAllowToConnectToRunningContainers stop containers


lordofthejars:
shouldAllowToConnectToRunningContainers stop containers if container was not started previously. Current behavior is:

if container is already running then is reused, and not stopped at the end of the test execution.

if container is not running and shouldAllowToConnectToRunningContainers stop containers is true, then a container is started and stopped.

The correct behavior should be that if shouldAllowToConnectToRunningContainers stop containers is true and container is not started, Cube should start one but it should not stop it.

lordofthejars:
After discussing the issue the expected behaviour will be next:

`shouldAllowToConnectToTunningContainers` is going to be an enum with three possible values:

*NO (default): start/stop all and fail if started. This is how it works when now the parameter is false.
* STARTORCONNECT: start or connect and stop if we started.
* STARTANDLEAVE: start and leave started for reuse.

lordofthejars:
Also `shouldAllowToConnectToRunningContainers` is renamed to `connectionMode`.

PR #111, opened 2015-02-22
resolves issue #110 by adding connection mode

lordofthejars:
resolves issue #110 by adding connection mode attribute for reusing already existing containers.

### DIFF
--- a/docker/src/main/java/org/arquillian/cube/impl/client/CubeConfiguration.java
+++ b/docker/src/main/java/org/arquillian/cube/impl/client/CubeConfiguration.java
@@ -19,13 +19,13 @@ public class CubeConfiguration {
     private static final String DOCKER_CONTAINERS_FILE = "dockerContainersFile";
     private static final String DOCKER_REGISTRY = "dockerRegistry";
     private static final String AUTO_START_CONTAINERS = "autoStartContainers";
-    private static final String SHOULD_ALLOW_TO_CONNECT_TO_RUNNING_CONTAINERS = "shouldAllowToConnectToRunningContainers";
+    private static final String CONNECTION_MODE = "connectionMode";
     private static final String BOOT2DOCKER_PATH = "boot2dockerPath";
 
     private String dockerServerVersion;
     private String dockerServerUri;
     private String dockerRegistry;
-    private boolean shouldAllowToConnectToRunningContainers = false;
+    private ConnectionMode connectionMode = ConnectionMode.STARTANDSTOP;
     private String boot2DockerPath;
     private String username;
     private String password;
@@ -35,8 +35,8 @@ public class CubeConfiguration {
 
     private Map<String, Object> dockerContainersContent;
 
-    public boolean shouldAllowToConnectToRunningContainers() {
-        return shouldAllowToConnectToRunningContainers;
+    public ConnectionMode getConnectionMode() {
+        return connectionMode;
     }
 
     public String getDockerServerUri() {
@@ -133,8 +133,8 @@ public class CubeConfiguration {
             cubeConfiguration.autoStartContainers = ConfigUtil.trim(map.get(AUTO_START_CONTAINERS).split(","));
         }
 
-        if(map.containsKey(SHOULD_ALLOW_TO_CONNECT_TO_RUNNING_CONTAINERS)) {
-            cubeConfiguration.shouldAllowToConnectToRunningContainers = Boolean.parseBoolean(map.get(SHOULD_ALLOW_TO_CONNECT_TO_RUNNING_CONTAINERS));
+        if(map.containsKey(CONNECTION_MODE)) {
+            cubeConfiguration.connectionMode = ConnectionMode.valueOf(ConnectionMode.class, map.get(CONNECTION_MODE));
         }
         return cubeConfiguration;
     }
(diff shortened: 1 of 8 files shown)


docker/src/main/java/org/arquillian/cube/impl/client/CubeConfiguration.java
```diff
@@ -133,8 +133,8 @@ public static CubeConfiguration fromMap(Map<String, String> map) {
             cubeConfiguration.autoStartContainers = ConfigUtil.trim(map.get(AUTO_START_CONTAINERS).split(","));
         }
 
-        if(map.containsKey(SHOULD_ALLOW_TO_CONNECT_TO_RUNNING_CONTAINERS)) {
-            cubeConfiguration.shouldAllowToConnectToRunningContainers = Boolean.parseBoolean(map.get(SHOULD_ALLOW_TO_CONNECT_TO_RUNNING_CONTAINERS));
+        if(map.containsKey(CONNECTION_MODE)) {
+            cubeConfiguration.connectionMode = ConnectionMode.valueOf(ConnectionMode.class, map.get(CONNECTION_MODE));
```
aslakknutsen: Add some better Exception handling then just the default IllegalArgumentException("No Enum constant " + type...) coming from valueOf

docker/src/main/java/org/arquillian/cube/impl/client/CubeSuiteLifecycleController.java
```diff
@@ -100,22 +100,28 @@ private boolean isCubeRunning(String cube) {
...
-                    controlEvent.fire(new StartCube(cubeId));
+                    return null;
+                }
+                controlEvent.fire(new CreateCube(cubeId));
+                controlEvent.fire(new StartCube(cubeId));
+
+                if(connectionMode.isAllowReconnect() && !connectionMode.isStoppable()) {
+                 // If we allow reconnections and containers are none stoppable which means that they will be able to be
+                 // reused in next executions then at this point we can assume that the container is a prerunning container.
+                 controlEvent.fire(new PreRunningCube(cubeId));
```
aslakknutsen: Hmm.. are we missing some else or something here? Why create/start then call PreRunningCube?

aslakknutsen:
We're missing 'don't stop if leave' part don't we? As far as I can tell we always stop/destroy the Cube?

docker/src/main/java/org/arquillian/cube/impl/client/CubeConfiguration.java
lordofthejars: Agreed

lordofthejars:
Hi we are not leaving the don't stop if leave. To not stop a container we only need to set the container as prerunningcube. In case that you set connection mode to STARTORCONNECTXXX means that allowReconnect returns true, and then you can check if container is running. If it is the case then we send the prerunningcontainer event.

In other cases (in all connection modes) we need to start the container, it doesn't matter if we are in STARTANDSTOP or in STARTORCONNECTXXX, because we can be sure that container is not running, so we start it.

Then when the container is started only if connection mode is STARTORCONNECTANDLEAVE we must not stop the container. And what we do is fire the PreRunningEvent. From the point of view of STARTORCONNECTANDLEAVE it doesn't matter if container was started just now or two executions back, he only wants to notify the event to set to Cube that the container should not be stopped.

aslakknutsen:
mm, right. It's DockerCube that just don't call stop if it's in PRE_RUNNING state.

lordofthejars:
So we should change this approach? PreRunningEvent change state of the cube.

lordofthejars:
> The command events are still fired, but the Cube is in the wrong state.. 
why this happens?

aslakknutsen:
AfterSuite is always called; https://github.com/lordofthejars/arquillian-cube/blob/master/docker/src/main/java/org/arquillian/cube/impl/client/CubeSuiteLifecycleController.java#L42

The Cube.stop/destroy methods are always called.. But the DockerCube is in the wrong state so it just returns instead of calling DockerExecutorClient.

lordofthejars:
Well it will not happen anything simply that because it is in prerunning state it won't be stopped which is the desired behaviour. What I tried is to avoid having special cases, in the sense that a STARTORCONNECTANDLEAVE logic doesn't matter if container it has just started in current execution or not, it will leave the container without stopping it.

aslakknutsen:
I was more concerned with the Events being fired for extensions. But extensions should probably rely on BeforeStop/AfterStop and not StopCube. StopCube is always fired, but Before/AfterStop is just fired if it's actually stopped.

So I think technically we're good here. 

Personally I found it a bit odd that the logic is a bit split; We handle PreConnect specially in the CubeSuiteLifecycleHandler but when it comes to stopping or not stopping it's handled in DockerCube just because it's placed in a non stoppable state. 

Wondering if PRE_RUNNING is not a State, but some kinda Mode. the State would be STARTED. You can't stop a Cube in PRE_RUNNING mode, but it's still started. hm..

lordofthejars:
Yes I implemented in this way for one reason and now I can't remember. I just wondered the same as you do, and finally I decided the move there. Probably to have all the same logic in the same place. Of course it would be better to have the start as well there but because we need to know if the container is running and so on and the logic is much complex I decided to move in upper level. Probably you are right that prerunning is a mode and not a state.
So we can accept for this release and figure out how we can do it in next release?

aslakknutsen:
@lordofthejars sure, make a 'improvment' issue on it

PR #141, opened 2015-06-02
Adds TomEE example

lordofthejars:
Adds example of using Apache TomEE with Arquillia Cube.

### DIFF
--- a/docker/ftest-boot2docker/pom.xml
+++ b/docker/ftest-boot2docker/pom.xml
@@ -125,6 +125,27 @@
                 </dependency>
             </dependencies>
         </profile>
+        
+        <profile>
+            <id>tomee-dockerfile</id>
+            <properties>
+                <arquillian.launch>tomee_dockerfile</arquillian.launch>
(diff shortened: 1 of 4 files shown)


docker/ftest-boot2docker/pom.xml
```diff
@@ -125,6 +125,27 @@
...
+        <profile>
+            <id>tomee-dockerfile</id>
+            <properties>
+                <arquillian.launch>tomee_dockerfile</arquillian.launch>
+            </properties>
+            <dependencies>
+                <dependency>
+                    <groupId>org.apache.openejb</groupId>
+                    <artifactId>arquillian-tomee-remote</artifactId>
+                    <version>1.7.1</version>
```
cchacin: TomEE 1.7.2 was released

lordofthejars: Yes I know but currently the docker image is a 1.7.1, so the version should be the same. I will update in next PR. Thank you :+1:

Issue #159, opened 2015-07-20
Makes embeddedPort and containerlessDocker properties optionals


lordofthejars:
Makes embeddedPort and containerlessDocker propreties optionals. And then try to get from the current configuration. 

For example if there is only one port binding or only one container defined, it is not necessary to set there since they can be retrieved from configuration.

PR #160, opened 2015-07-20
resolves issue #159 by resolving containerless properties automatically.

lordofthejars:


aslakknutsen:
I believe this will always create a Cube Controlled Container regardless of there is a 'active' container configuration configured as long as the Containerless Container is on classpath. If no container is configured in arquillian.xml, one is created and all default values are assumed.

Not sure how much of a problem that is per say, but..

lordofthejars:
Yes I understand your point, well probably you will always have container configuration since in theory your service will contain more than one docker instance (1 for service and 1 for db). And in case not then containerless container with default properties will be enough. For now I don't see a problem at all.

Issue #161, opened 2015-07-28
Make autoStartContainers regular expression aware


lordofthejars:
autoStartContainers is a property where you specify which containers must be started before the application/main container is started. Currently it supports a comma-separated list, so if you need to start two containers you can do it as `A, B`. This is ok for a few containers but in cases you want to start more than once, for example one email server, one database server, one key-value server, then it needs more work. Also if you change the name of the container then the test will fail because the container won't be started.

To avoid this scenario we can apart from supporting comma approach using regular expression approach, so you can do something like `^((?!tomee).)*$` which means it will autostart all containers except the one tagged as tomee.

aslakknutsen:
This sounds reasonable.

Another approach we could look into as well is:

* Auto start linked containers to 'managed' container

lordofthejars:
So true I will implement this in a separate issue and be the default
strategy to use.
El dt., 28 de jul., 2015 a les 19.10 Aslak Knutsen <notifications@github.com>
va escriure:

PR #163, opened 2015-07-29
resolves issue #161 by adding regular expression support on autostart

lordofthejars:
resolves issue #161 by adding regular expression support on autostart containers.

### DIFF
--- a/docker/docker/src/main/java/org/arquillian/cube/docker/impl/util/AutoStartOrderUtil.java
+++ b/docker/docker/src/main/java/org/arquillian/cube/docker/impl/util/AutoStartOrderUtil.java
@@ -7,6 +7,8 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.arquillian.cube.docker.impl.client.CubeDockerConfiguration;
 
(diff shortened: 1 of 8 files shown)


docker/docker/src/main/java/org/arquillian/cube/docker/impl/util/AutoStartOrderUtil.java
```diff
@@ -70,6 +86,14 @@
         return new HashSet<>(nodes.values());
     }
 
+    private static boolean isRegularExpressionBased(String[] autoStartContainers) {
+        return autoStartContainers.length == 1;
```
aslakknutsen: what if you only have one named container to start? Any chance a container name could cause a 'fail to parse regexp' or create a regexp that doesn't actually match the name?

docker/docker/src/main/java/org/arquillian/cube/docker/impl/util/AutoStartOrderUtil.java
```diff
@@ -70,6 +86,14 @@
...
+    }
+
+    private static boolean isCommaSeparatedBased(String[] autoStartContainers) {
+        return autoStartContainers.length > 1;
```
aslakknutsen: Any chance we will get a regexp that contain a ',' so it's handled as a exact match instead of an expression?

lordofthejars: I thought about this case but I think that it would be quite strange to give a container a name with ',' maybe not but a bit strange

docker/docker/src/main/java/org/arquillian/cube/docker/impl/util/AutoStartOrderUtil.java
lordofthejars: Yes I thought this but I am not sure how to resolve this in a better way. Maybe adding that if starts with keyword `regexp:` then it is a regular expression. But I tried to be as much clear as possible. But maybe I need to add this approach.

aslakknutsen: yea, a regexp: prefix was my first thought.

Issue #162, opened 2015-07-29
Auto start linked containers to 'managed' container


lordofthejars:
In case that autostart property is not set, Arquillian Cube could start all containers that are required by the master container.

lordofthejars:
@aslakknutsen I have seen that autostart containers are started at BeforeSuite event. We talked that to fix this issue we need to split in two where autostart containers happening, the regexp and comma-separated at beforesuite event and in case of autostart linked containers of managed ones since we need to know which is the managed one at BeforeSetup event.

My question is do you think it has sense to move also the regexp and comma-separated to be started BeforeSetup too and stop them at AfterSuite? In this way we will have all the logic in the same place and not spread it across all the code.

aslakknutsen:
On top of my head I don't think it makes sense to have the linked containers follow the lifecycle of the managed container, e.g. started and stopped with the managed container.. So, possible we should turn it around; Figure out which are managed up front, then autostart the linked with the 'other autostart' ones in BeforeSuite.  Then later start the managed container in BeforeSetup.

?

lordofthejars:
I agree with you, the problem is that in BeforeSuite we don't have the information to get the managed container, we don't have the `event.getDeployableContainer()`.

Maybe I will need to create a common class and call it from different points to not replicate at least the code, but of course we will have two points of starting auto-start containers which I don't like it at all.

aslakknutsen:
True, but in BeforeSuite you would want to start all linked containers for all managed containers correct? So you don't need the individual events with getDeployableContainer. Inject and use the ContainerRegistry directly. If any of the names of the containers in the ContainerRegistry match the name of a Cube, then start the Cubes linked Containers.. Doesn't that work?

lordofthejars:
mmm yes, it is worth to try it. When I finish the integration with docker-java 2.0 I will implement this.

PR #179, opened 2015-09-18
resolves issue #162 by autostarting containers in case of no autostart strategy

lordofthejars:
resolves issue #162 by autostarting containers in case of no autostart strategy


lordofthejars:
@aslakknutsen if you want I merge this PR and we discuss any other addition as an enhancement

aslakknutsen:
Merge for now.We'll revisit if the use case of having both auto and manually registered autostarting at once ever come up in real life.. :)

PR #180, opened 2015-09-25
Renamed HostIP.java to match class name

mikesir87:
When running the builds, I was getting 

```
class HostIp is public, should be declared in a file named HostIp.java
```

This simple file renaming fixes that error.


lordofthejars:
Thank you so much.

Issue #181, opened 2015-10-05
PollingAwaitStrategy: polling problem with ss command on Windows


foxlegend:
Hi there,

When executing an Arquillian test case on Windows with the Cube extension I had a problem with the polling strategy: polling limit was reached and Cube raised an exception as it was unable to connect to my container.
I took a look at my docker daemon logs, and I found such lines:

    time="2015-10-05T14:38:39.352198042Z" level=debug msg="Calling POST /containers/{name:.*}/exec"
    time="2015-10-05T14:38:39.352238525Z" level=info msg="POST /v1.20/containers/jbossas/exec"
    time="2015-10-05T14:38:39.362334001Z" level=debug msg="Calling POST /exec/{name:.*}/start"
    time="2015-10-05T14:38:39.362387345Z" level=info msg="POST /v1.20/exec/100722dc8c81731b53aee9b11ac42a676d5803ff9559a1b5e17db4b5dec6393a/start"
    time="2015-10-05T14:38:39.363428060Z" level=debug msg="starting exec command 100722dc8c81731b53aee9b11ac42a676d5803ff9559a1b5e17db4b5dec6393a in container fd471659bf523c0aaebe933ef157779d7c932f291480a031a275263900a816ec"
    time="2015-10-05T14:38:39.363579706Z" level=debug msg="attach: stdout: begin"
    time="2015-10-05T14:38:39.459238430Z" level=debug msg="Exec task in container fd471659bf523c0aaebe933ef157779d7c932f291480a031a275263900a816ec exited with code 1"
    time="2015-10-05T14:38:39.459315737Z" level=debug msg="attach: stdout: end"
    2015/10/05 14:38:39 http: response.WriteHeader on hijacked connection

I found that the command in `ss.sh` was executed: `ss -lntu | grep '${port}' | grep LISTEN -c`, and the template was built thanks to the `org.arquillian.cube.docker.impl.util.IOUtil` class, with the offending function:

```java
    public static String asStringPreservingNewLines(InputStream response) {
        StringWriter logwriter = new StringWriter();

        try(BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(response))) {

            String line = null;
            while ((line = bufferedReader.readLine()) != null) {
                logwriter.write(line);
                logwriter.write(System.lineSeparator());
            }

            return logwriter.toString();
        } catch (IOException e) {
            throw new RuntimeException(e);
        }
    }
```

[...]

PR #182, opened 2015-10-07
resolves issue #181 by forcing a Unix EOL instead of a platform dependant one

foxlegend:
The pull request proposes to use the ``IOUtils.LINE_SEPARATOR_UNIX`` constant from ``commons-io`` (already available in the module) to specify line ending.


foxlegend:
You're welcome, I'm glad I could help ;)

Issue #169, opened 2015-08-14
Print final configuration variables


lordofthejars:
Since Arquillian Cube autoresolve a lot of things related with docker like ports, ips, ... add a sysout of the final configuration.

Issue #184, opened 2015-10-12
Updates Docker-Java to 2.1.1


lordofthejars:
Updates Docker-Java to 2.1.1

PR #187, opened 2015-10-28
resolves issues #169, #184 and #185

ivannov:
Print the final docker configuration and update the versions of some dependency


docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfiguration.java
```diff
@@ -192,4 +192,23 @@ public static CubeDockerConfiguration fromMap(Map<String, String> map) {
 
         return cubeConfiguration;
     }
+
+    @Override public String toString() {
+        return "CubeDockerConfiguration{" +
```
lordofthejars: Can each option be printed in different line so it will be easy for the reader.

lordofthejars:
Thank you so much Ivan :)

PR #203, opened 2015-11-16
Implements several issues for next version of Cube

lordofthejars:


lordofthejars:
So true, sorry I start coding and coding and I don't remember to document.

aslakknutsen:
Not a fan of the new 'commons' dependencies, but beyond that looks good. :)

lordofthejars:
I know you would said this, but they are not new, they are a dependency from other projects(docker-java) so I am only using them. commons-lang is used by cube-core.

aslakknutsen:
sure, still not a fan. :)

lordofthejars:
not a big fan but commons-lang in our case is really helpful. The only one that I might replace would be the MapUtils class for one of our own. I would do it in near future.

lordofthejars:
https://github.com/arquillian/arquillian-cube/issues/207

aslakknutsen:
:+1:

lordofthejars:
So ready to merge?

aslakknutsen:
yea :)

PR #204, opened 2015-11-17
Minor fixes for test cases.

robertpanzer:
I tried to build arquillian-cube and there seem to be at least these three issues where the test cases fail on my Windows box:

- In `AutomaticResolutionAutoStartParser.java` the content returned is an `ArrayList` instead of a `Set`.
  In this case I think it's valid to only assume that the returned content is some kind of collection.
- In `CubeDockerConfiguration.java` `URI.create()` fails with a Windows path.
  I think `new File(location).toURI()` should work on all platforms.
  This is probably not only an issue with the test cases but also a general problem when having a container configuration file!
- In `CubeConfiguratorTest.java` the class `CubeDockerConfiguration` simply produces another output than expected.
  Additionally if the env variable `DOCKER_HOST` is not set the default docker URI is `http://127.0.0.1:22222`. If it is set it will be `https://...`.

### DIFF
--- a/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/AutomaticResolutionAutoStartParser.java
+++ b/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/AutomaticResolutionAutoStartParser.java
@@ -3,10 +3,10 @@ package org.arquillian.cube.docker.impl.client;
 
 import org.arquillian.cube.docker.impl.util.AutoStartOrderUtil;
 
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
 
 public class AutomaticResolutionAutoStartParser implements AutoStartParser {
 
@@ -30,7 +30,7 @@ public class AutomaticResolutionAutoStartParser implements AutoStartParser {
             }
 
             if (content.containsKey("links")) {
-                Set<String> links = (Set<String>) content.get("links");
+                Collection<String> links = (Collection<String>) content.get("links");
                 for (String link : links) {
                     String[] parsed = link.split(":");
                     String name = parsed[0];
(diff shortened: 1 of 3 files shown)


lordofthejars:
hmmm I will need to check the URI.create thing and apply what you mention here. I thought I had controlled in code since there is a try/catch that takes care of this: https://github.com/arquillian/arquillian-cube/blob/master/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/DockerContainerDefinitionParser.java#L61 but the create can throw an exception as well.

lordofthejars:
Also notice that this: `final URI aaa = new File("c:\\aaa\\bbb").toURI();` returns next: `file:/Users/alexsoto/git/crypto/c:%5Caaa%5Cbbb` which is not valid. I think that since this is an URI (and it is documented as is, user should take care of provide a valid URI. What we can do is provide a WARNING message in docs.

robertpanzer:
The location is is defined here to be the absolute path of a temporary `config.yaml`:
https://github.com/arquillian/arquillian-cube/blob/master/docker/docker/src/test/java/org/arquillian/cube/docker/impl/client/CubeConfigurationTest.java#L122
The location there is sth like: `C:\Users\ROBERT~1.PAN\AppData\Local\Temp\junit637076650332815864\config.yaml`
On my Windows-Box `newFile.toURI()` returns `file:/C:/Users/ROBERT~1.PAN/AppData/Local/Temp/junit637076650332815864/config.yaml` which looks ok to me at first sight.

lordofthejars:
then  I need to fix the test which is wrong in case of running on Windows

robertpanzer:
It's the one I linked to above:
https://github.com/arquillian/arquillian-cube/blob/master/docker/docker/src/test/java/org/arquillian/cube/docker/impl/client/CubeConfigurationTest.java#L112-L135

lordofthejars:
:+1:

lordofthejars:
@robertpanzer I am going to close this PR because since all changes inContainerObject pattern it was easier to copy the changes rather than deal with merge. Sorry :beers:

robertpanzer:
No problem.
My fault, I always tend to put too much into one PR. :wink: 
:beers:

PR #210, opened 2015-11-18
Return ContainerConfig exposedPorts if no HostConfig boundPorts found

### LINKED ISSUE #208: PollingAwaitStrategy throws NPE when no portBinding is provided
To my understanding you don't need to explicitly bind ports to reach a container inside docker it self, only if you need to reach it from the outside. 

Since the polling strategy only needs the internal ports to operate anyway, it should inspect the InspectContainerResponse.ContainerConfig.ExposedPorts, not HostConfig to find the ports exposed. 

```java
Caused by: java.lang.NullPointerException
	at org.arquillian.cube.docker.impl.util.BindingUtil.binding(BindingUtil.java:31)
	at org.arquillian.cube.docker.impl.model.DockerCube.bindings(DockerCube.java:143)
	at org.arquillian.cube.docker.impl.await.PollingAwaitStrategy.await(PollingAwaitStrategy.java:91)
	at org.arquillian.cube.docker.impl.model.DockerCube.start(DockerCube.java:93)
	... 18 more
```

aslakknutsen:
In this scenario boundPort is set to -1.

fixes #208 

I updated the 'ping' polling option to take the -1 boundPort into account. Not sure if there are other places we need to handle this as well?

### DIFF
--- a/docker/docker/src/main/java/org/arquillian/cube/docker/impl/await/PollingAwaitStrategy.java
+++ b/docker/docker/src/main/java/org/arquillian/cube/docker/impl/await/PollingAwaitStrategy.java
@@ -95,6 +95,9 @@ public class PollingAwaitStrategy implements AwaitStrategy {
 
             switch(this.type) {
                 case "ping": {
+                    if(ports.getBindingPort() == -1) {
+                        throw new IllegalArgumentException("Can not use polling of type " + type + " on non externally bound port " + ports.getExposedPort());
+                    }
                     if (!Ping.ping(bindings.getIP(), ports.getBindingPort(), this.pollIterations, this.sleepPollTime,
                             this.timeUnit)) {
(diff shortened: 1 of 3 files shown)


lordofthejars:
Maybe when you use `@HostPort` in ContainerObject https://github.com/arquillian/arquillian-cube/blob/e53596e9545e40c81fdce835ec84d33c05442f08/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/containerobject/CubeContainerObjectTestEnricher.java#L182 where we can decide if we let user to receive a -1 or throw an exception.

aslakknutsen:
Added this: https://github.com/arquillian/arquillian-cube/pull/210/files#diff-9c660a5a29c9fc18b6b22ba056254c8eL181

PR #214, opened 2015-11-23
Avoid ClassCastException in AutomaticResolutionStartParser

robertpanzer:
At least when building on Win7 the links are returned as an ArrayList instead of a Set as expected.
Just expecting a Collection makes the test pass for me on Win.

### DIFF
--- a/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/AutomaticResolutionAutoStartParser.java
+++ b/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/AutomaticResolutionAutoStartParser.java
@@ -3,10 +3,10 @@ package org.arquillian.cube.docker.impl.client;
 
 import org.arquillian.cube.docker.impl.util.AutoStartOrderUtil;
 
+import java.util.Collection;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


lordofthejars:
yes it is fixed locally, but I am not sure why this is happening since a A set is created https://github.com/arquillian/arquillian-cube/blob/master/docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/compose/ContainerBuilder.java#L290

robertpanzer:
This code is actually never called when I debug it.
The ArrayList that I see is created directly by the Yaml parser.

(Maybe because the file is treated as a Cube file and not as a docker-compose yaml?)

lordofthejars:
Right, are you using universe?

robertpanzer:
I'm building arquillian-cube, not an own project.

robertpanzer:
The test case that is failing on my machine is `org.arquillian.cube.docker.impl.client.CubeSuiteLifecycleControllerTest#shouldCreateAndStartAutoContainersWhenNoAutoStartIsProvided()`

lordofthejars:
ok, no worries now it is fixed, but yes probably the YAML is the one who is causing this.

PR #215, opened 2015-11-23
Fix DockerComposeConverterTest when building on Windows.

robertpanzer:
The DockerComposeConverterTest expects that an image name in the docker-compose.yaml is correctly replaced with the content of an environment variable.
Unfortunately the env var `USER` is not available on Windows, so that the test fails.
(`USER` is probably chosen as an arbitrary image name without any further purpose.)

Fixed by explicitly setting the env variable `TESTIMAGENAME` in the configuration of the surefire plugin and using this in the yaml file.

Additionally fixed a typo in a private method name that appears in log messages (`logUsupportedOperations`)

### DIFF
--- a/docker/docker/pom.xml
+++ b/docker/docker/pom.xml
@@ -122,4 +122,18 @@
         
     </dependencies>
 
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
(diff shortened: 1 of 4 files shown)


lordofthejars:
Thank you very much for this fix on Windows systems. Windows it is becoming a headache since we don't have a Windows machine to test everything.

robertpanzer:
You're welcome. :smile: :beers:

PR #240, opened 2015-12-18
Add configuration Object model

### LINKED ISSUE #232: Convert DockerContainerDefinition to proper Object model
Move away from the usage of magic Map<String, Object> everywhere...

aslakknutsen:
fixes #232 

One major thing left;

We used to have Map<String, Object> Cube.configuration() which now is changed to T Cube<T>.getConfiguration(). This was done to allow OpenShiftCube and DockerCube have different configuration internally. The only location where this is actually used is in Containerless Contianer. 

The problem is Containerless Container need the 'BuildImage' part of the Cube configuration, something Openshift only mimiced previously. Containerless container doesn't know about DockerCube or OpenShiftCube specifically, it only operates on Cube. 

Not sure if we should add a Cube.getBuildImage or add some common BaseCubeConfiguration they both can extend, or.. 

Anyway.. You should be able to continue building based on this branch as is, then we can merge etc when done.

Please do a quick review to make sure I didn't change/remove some logic :)

### DIFF
--- a/containerless/src/main/java/org/arquillian/cube/impl/containerless/ContainerlessDockerDeployableContainer.java
+++ b/containerless/src/main/java/org/arquillian/cube/impl/containerless/ContainerlessDockerDeployableContainer.java
@@ -16,6 +16,7 @@ import org.arquillian.cube.spi.event.CubeControlEvent;
 import org.arquillian.cube.spi.event.DestroyCube;
 import org.arquillian.cube.spi.event.StartCube;
 import org.arquillian.cube.spi.event.StopCube;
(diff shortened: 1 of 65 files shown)


lordofthejars:
Looks good to me, the only thing I would do is adding an `isXXX` methods as well so in several places we can avoid doing `getXXX() != null` but that's all, looks really awesome.

aslakknutsen:
I think this should be good to go now. Please give it a review/spin.

aslakknutsen:
There might be some Util methods I've missed around to deal with the old stuff.

PR #259, opened 2016-01-22
add support for port forwarding to openshift; update fabric8 kubernetes client version

rcernich:
Adds support for forwarding to pod ports from the host running the test.  This allows tests to be run on a remote OpenShift instance.


aslakknutsen:
Just trying to understand here, Why does Cube need to setup a proxy?

The usecase, if I understand it is;

* Run Arquillian 'client' in a Container on OpenShift

In this scenario, Arquillian is configured to (probably) point to the same OpenShift instance as it's running in. Then Arquillian Cube starts a few new Containers... For some reason Cube needs to locally forward port X to remote Container.N port Y. ?

rcernich:
This supports running tests locally (i.e. on your laptop) against an image running on a remote instance of OpenShift (i.e. not on your laptop).  Because of that, there is no way to access anything other than a service exposed by a route, as there is no network connectivity to the pod or service IPs.  These tests are effectively running against a remote instance of EAP, which is configured through a pod descriptor and deployed on OpenShift.  What this change does is allow you to configure proxies to the pod ports and those proxies are used to configure the Arquillian remote container.  This allows existing Arquillian tests to be executed on an instance in the cloud.  In this case, the build machine does not need to have Docker or OpenShift installed.

The next step for us is to be able to instantiate an application (e.g. using a template) and run Arquillian tests against it.  @alesj has already done a lot of work to make that happen (e.g. instantiating a template, configuring additional resources, etc.).  The next step for us is to be able to run a separate Arquillian container that can be used to execute tests against the application from within OpenShift (e.g. verifying a HotRod service from a cluster of JDG pods through scale ups/downs).

We have two goals:

1. Be able to verify functionality of the templates and images we produce.
2. Allow users to easily run their existing test suites against their application as manifested by the templates on the images we produce.

In the example I put together here https://github.com/rcernich/jboss-eap-quickstarts/tree/openshift-arquillian/kitchensink illustrates part of this (running existing tests on an image running in OpenShift).  Without port forwarding, the only way to do this would be to have OpenShift running on the build machine, executing with a user with elevated privileges (to bind the HostPort).  With this change, none of that is required.

Hope that helps.

rcernich:
As mentioned in the PR, port forwarding is not much different from what you're doing with mapped ports on Docker.  The same is true of HostPort declarations on Kubernetes pods.  However, both of those cases require that Docker or OpenShift are running on the same machine on which the test is running (or, for the OpenShift case, that the machine has network connectivity to the pod IP space).  This change allows you to do the same test against a remote OpenShift instance (i.e. in the cloud).

lordofthejars:
@rcernich thank you very much for collaborating to Arquillian Cube, for me it looks an awesome new improvement but since @aslakknutsen coded the OpenShift integration, I prefer he takes a look. Also very good explanation on why you need this.

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/PortForwardServerConnection.java
```diff
@@ -0,0 +1,354 @@
...
+                 * client is done and the request stream closes.
+                 */
+                requestComplete.await();
+                /* wait for the response on the error stream. */
+                errorComplete.await();
+            } catch (InterruptedException e) {
+                e.printStackTrace();
+            }
+        } catch (Exception e) {
+            e.printStackTrace();
```
lordofthejars: This printStackTrace will hide a possible exception, is this intentionally done or needs some refactor this part of code.

rcernich: There's no way to get the exception back into the test execution flow, since this is invoked from the socket listener on the local port that's doing the forwarding.  The close on the accepted socket _should_ cause a failure in the client expecting to connect to whatever is on the other side (e.g. the Arquillian EAP remote container should time out or error when trying to connect to the EAP management port because it would see the connection closed; or a servlet test would fail because it can't access the servlet; etc.).

BTW, this code is still a little rough around the edges, so some of this stuff could be replaced with log messages (as opposed to printStackTrace()), but it's a start.

aslakknutsen:
Aha.. Now I get it. I was missing this part in my first review; https://github.com/arquillian/arquillian-cube/pull/259/files#diff-92d2e8d4321dd54fa6d36b60031b8676R52

I thought 'something else' setup the actual proxy(like ssh on OpenShift 2), and could quite understand why we needed ot proxy between two local ports as oppose to just calling the 'targetPort' directly.

aslakknutsen:
But.. Since we currently don't support Route/Service deployments, and technically there is no real way of connecting to a Pod without them(or without a PortForward request), why do we force the user to configure this?

Couldn't we simply make it; If you have bound a 'Container Adapter' to a Cube that is a Pod, then auto setup a PortForward/Proxy?

core/src/main/java/org/arquillian/cube/impl/client/container/CubeContainerLifecycleController.java
```diff
@@ -24,7 +24,8 @@
     @Inject
     private Event<CubeControlEvent> controlEvent;
 
-    public void startCubeMappedContainer(@Observes BeforeStart event, CubeRegistry cubeRegistry,
+    // Must be started before ContainerConfigurationController
+    public void startCubeMappedContainer(@Observes(precedence=100) BeforeSetup event, CubeRegistry cubeRegistry,
```
aslakknutsen: Not a fan of this change as it changes the interaction between the Container lifecycle and the 'hidden' Cube'ification of an Adaptor. BeforeSetup is always called in BeforeSuite, while BeforeStart could technically be called at any point controlled by the User in manual mode. With this change, the Container would be started regardless of what the user say..

Cube+Remote adapter actually works very similar to a Managed container at the moment, this change would alter that.

I do understand why you want it in this case, as you can't setup the PortForward until the Pod is running, but... Since we know the Proxy Port we will use, we can technically configure the adapter to use our proxy port before starting the Proxy.

rcernich:
The user is not required to configure this, but forwarding to the management port is required for deployments and in-container test cases, and forwarding to web port would be required for servlet tests, etc.  It's also possible (probable) that some of these ports won't be declared on the pod (e.g. management port) and would need to be mapped explicitly.

There seems to be some overlap with existing port mapping (e.g. docker -p and pod HostPort), but I like the rename.  Do you want me to make this change and push -f?

@alesj has added support for routes, services, templates, role bindings, etc. as part of https://github.com/jboss-openshift/ce-arq and we're working to see how we can merge this with arquillian-cube.

openshift/openshift/src/main/java/io/undertow/protocols/spdy/SpdyChannelWithoutFlowControl.java
```diff
@@ -0,0 +1,45 @@
+package io.undertow.protocols.spdy;
```
aslakknutsen: This should probably be in some 'org.arquillian.cube.openshift.impl....' package.

core/src/main/java/org/arquillian/cube/impl/client/container/CubeContainerLifecycleController.java
rcernich: Ok.  At the moment, I'm letting the socket pick a random port, so I'd need to figure out a way to allocate a port ahead of time....something to figure out.

openshift/openshift/src/main/java/io/undertow/protocols/spdy/SpdyChannelWithoutFlowControl.java
rcernich: Yes, but I'm overriding a package accessible method. :(  This is only required because the OpenShift SPDY implementation does not support flow control, so once the window is exhausted, we stop forwarding data.  I wish there was an easier way (and maybe there is, but I haven't found it).

aslakknutsen:
> The user is not required to configure this, but forwarding to the management port is required for deployments and in-container test cases, and forwarding to web port would be required for servlet tests, etc. 

We only attempt to Forward the ports that the user actually asks about anyway, e.g. if you target a Remote Adaptor against a Pod, we require that those ports are accessible. So correct, to deploy a Pod the user is not required to configure this, but if the Pod contains a 'Arquillian Container' then the user is required to configure it for the 'Arquillian Container' to work at all. And in the later case I think we could automate most of it. 

> It's also possible (probable) that some of these ports won't be declared on the pod (e.g. management port) and would need to be mapped explicitly.

You mean there is no requirement to define spec.containers[].ports[].containerPort for the PortForward to work?

rcernich:
> > It's also possible (probable) that some of these ports won't be declared on
> > the pod (e.g. management port) and would need to be mapped explicitly.
> 
> You mean there is no requirement to define
> spec.containers[].ports[].containerPort for the PortForward to work?
Correct.

aslakknutsen:
> There seems to be some overlap with existing port mapping (e.g. docker -p and pod HostPort), but I like the rename. Do you want me to make this change and push -f?

Hmm, yea, maybe the whole binding dance with the DockerCube should just be placed behind the same Metadata object. Then we wouldn't require two ways of checking/figuring out which port to use.

What we want to answer in both cases are:

* What IP can I use to communicate with this Cube
* What Port can I use to communicate with OtherPort on this Cube

You don't have to push -f, just push. Makes it easier to follow changes as thye come in. We'll clean up the commits before going upsteam when done. :)

core/src/main/java/org/arquillian/cube/impl/client/container/CubeContainerLifecycleController.java
aslakknutsen: Might be some code in there that is woth looking at: https://github.com/mryan43/arquillian-available-port-extension

'Finding an open port' is probably good enough. Assume it won't be taken over the next few seconds. :)

aslakknutsen:
>> You mean there is no requirement to define
 spec.containers[].ports[].containerPort for the PortForward to work?
> Correct

Well, it doesnt really matter I think. We don't need to know the 'internal Pod port' up front.

Either the Pod uses the standard EAP ports, then the Default Remote Adapter configuration knows which ports these are, and when you 'attempt to configure them' you can setup a Forward to that port.

Else the Pod is configured to use some 'other ports', then the Remote Adapter has to be configured by the user with the updated ManagementPort, and when you 'attempt to configure them' you can setup a Forward to the configured port.

When the Management connection is established, and a Deployment is done, we communicate with the API to see where this is deployed. Via this API we will know what the internal HTTP port is. We can now setup a PortForward to that port.

So essentially just setup a Forward to the Ports 'Arquillian/User' is asking to be used. No need to parse any Specs up front.. 

Or am I missing something?

rcernich:
----- Original Message -----

> > > You mean there is no requirement to define
> > 
> 
> > > spec.containers[].ports[].containerPort for the PortForward to work?
> > 
> 
> > > Correct
> > 
> 
> Well, it doesnt really matter I think. We don't need to know the 'internal
> Pod port' up front.

> Either the Pod uses the standard EAP ports, then the Default Remote Adapter
> configuration knows which ports these are, and when you 'attempt to
> configure them' you can setup a Forward to that port.

> Else the Pod is configured to use some 'other ports', then the Remote Adapter
> has to be configured by the user with the updated ManagementPort, and when
> you 'attempt to configure them' you can setup a Forward to the configured
> port.

> When the Management connection is established, and a Deployment is done, we
> communicate with the API to see where this is deployed. Via this API we will
> know what the internal HTTP port is. We can now setup a PortForward to that
> port.

> So essentially just setup a Forward to the Ports 'Arquillian/User' is asking
> to be used. No need to parse any Specs up front..

> Or am I missing something?
Sounds good...how do I go about that? (Keep in mind, I've only worked in Arquillian for about two weeks now. ;)).

aslakknutsen:
Maybe something like this:

```java
public interface CanCommunicateWith {

	IpPort onPort(int port);

	public interface IpPort {
		String getIp();
		int getPort();
	}
}


CanCommunicateWith com = cube.getMetadata(CanCommunicateWith.class);

IpPort http = com.onPort(8080);
assertEquals(http.getIp(), "some-externally-routed-ip")

IpPort mgm = com.onPort(9990);
assertEquals(mgm.getIp(), "localhost")
assertEquals(mgm.getPort(), locallyPortFOrwardedPort)
```

(Maybe a better name for the interface, but..)

That would allow us to use seperate IPs for different ports as we see fit(implementation wise). E.g. Deploy to a Pod via PortForward but do Test HTTP requests via external bound address.

I think you only need to update to use this new intercace in the following places:

https://github.com/arquillian/arquillian-cube/pull/259/files#diff-92be368c5478e37bad2c6bd1e800eac7R70
https://github.com/arquillian/arquillian-cube/pull/259/files#diff-573baa6a9349e9347d98f5e7d39079aaR37

Then add a impl of this interface to both OpenShift and Docker. The Docker one is fairly static; just return an object that reads bindings() for now. And in the OpenShift version you can do the PortForward thing. Essentially just make a Map of 'requested ports' and 'assigned ports'. Then in the OpenShift cube you can 'start' the PortForwards after the Cube.start() is called, and i guess shut them down before Cube.stop() is called.

openshift/openshift/src/main/java/io/undertow/protocols/spdy/SpdyChannelWithoutFlowControl.java
aslakknutsen: Aa, ok.

rcernich:
Changes are not as drastic as we discussed.  It looked like a lot of work to replace Binding, so I stayed close to the original implementation.  That said, OpenShiftPortProxyController could be moved into cube-core, since it only depends on HasMappedPorts.  (I'm terrible at names, so if you prefer a different name, please suggest an alternative.)

I chose not to automatically forward all ports as this may not be desirable, so it's up to the user to choose which ports they actually need forwarded.  That said, syntax like :9999 will map that port for all containers, to avoid having to explicitly specify it for each container.

I also reworked the PortBinding creation for OpenShift cube objects, since they appeared to be reversed (e.g. container port is synonymous with EXPOSE port and host port is akin to the -p port)

Let me know what you think.

spi/src/main/java/org/arquillian/cube/spi/metadata/HasMappedPorts.java
```diff
@@ -0,0 +1,13 @@
+package org.arquillian.cube.spi.metadata;
+
+
+public interface HasMappedPorts extends CubeMetadata {
+    
+    String getIP();
+
+    Integer forTargetPort(int targetPort);
+
+    void createProxies() throws Exception;
```
aslakknutsen: I would have moved create/destroy to the impl to avoid exposing it in SPI. The control of the creation and destruction of the Mapped Ports should be transparent and can be handled internally within the Cube.

spi/src/main/java/org/arquillian/cube/spi/metadata/HasMappedPorts.java
```diff
@@ -0,0 +1,13 @@
+package org.arquillian.cube.spi.metadata;
+
+
+public interface HasMappedPorts extends CubeMetadata {
+    
+    String getIP();
+
+    Integer forTargetPort(int targetPort);
```
aslakknutsen: I think this needs to return some kind of IP/Port Pair. It would allow to communicate with different ports on the same cube via different interfaces, e.g. externally routed or internally forwarded depending on the port. Maybe URI could do? tcp://ip:port? udp://ip:port?

core/src/main/java/org/arquillian/cube/impl/client/container/ContainerConfigurationController.java
```diff
@@ -64,12 +67,20 @@ public void remapContainer(@Observes BeforeSetup event, CubeRegistry cubeRegistr
                     int containerPort = getDefaultPortFromConfigurationInstance(newConfigurationInstance,
                             configurationClass, configurationClassPortField);
 
-                    PortBinding bindingForExposedPort = null;
-                    if ((bindingForExposedPort = binding.getBindingForExposedPort(containerPort)) != null) {
+                    PortBinding bindingForPort = null;
+                    if ((bindingForPort = binding.getBindingForExposedPort(containerPort)) != null) {
```
aslakknutsen: Could you add a Metadata impl for DockerCube as well that rely on Cube.binding() for now. Then we can at least focus on only one of the 'methods of getting binding data' in this part of the code. Maybe add a @deprecated to Cube.bindings() method.

core/src/main/java/org/arquillian/cube/impl/client/container/ProtocolMetadataUpdater.java
```diff
@@ -27,21 +28,35 @@ public void update(@Observes EventContext<ProtocolMetaData> eventContext, Contai
...
                     String ip = context.getHost();
                     int port = context.getPort();
-                    if(mapped != null && port != mapped.getBindingPort()) {
+                    final String bindingIp;
+                    final Integer bindingPort;
+                    if (mappedPorts != null && mappedPorts.forTargetPort(port) != null) {
+                        bindingIp = mappedPorts.getIP();
+                        bindingPort = mappedPorts.forTargetPort(port);
+                    } else {
+                        PortBinding mapped = binding.getBindingForExposedPort(context.getPort());
```
aslakknutsen: Could you add a Metadata impl for DockerCube as well that rely on Cube.binding() for now. Then we can at least focus on only one of the 'methods of getting binding data' in this part of the code.

spi/src/main/java/org/arquillian/cube/spi/metadata/HasMappedPorts.java
rcernich: I can look into this.

spi/src/main/java/org/arquillian/cube/spi/metadata/HasMappedPorts.java
rcernich: If this were expanded to replace Binding, I think that would be appropriate.  That said, replacing Binding was a lot more work.  I'll look again and see if I can make that work, but it's a pretty significant refactoring effort.

spi/src/main/java/org/arquillian/cube/spi/metadata/HasMappedPorts.java
aslakknutsen: Which means we can kill this class https://github.com/arquillian/arquillian-cube/pull/259/files#diff-2352d8cd76985b787b92036bbf6a7ff1

spi/src/main/java/org/arquillian/cube/spi/metadata/HasMappedPorts.java
aslakknutsen: I think for this issue we can do the following:

* Deprecate bindings()
* Implement a HasMappedPorts wrapper on top of bindings() for DockerCube so we can focus on HasMappedPorts as the single interface for deatling with what we're dealing with here
* Update HasMappedPorts to support IP per Port lookup.

Then we'll deal with updating the rest of Cube around binding later.

core/src/main/java/org/arquillian/cube/impl/client/container/ContainerConfigurationController.java
rcernich: Yep.  That's a good interim solution.  One quick question: what's the difference between bindings() and configuredBindings()?  I added a comments in my latest commit, but I'd like to verify that those comments are correct, i.e. bindings() are the actual bindings for the running container and configuredBindings() are simply what's been configured, e.g. container ports in the pod.json, or EXPOSE elements of a Dockerfile.

spi/src/main/java/org/arquillian/cube/spi/metadata/HasMappedPorts.java
rcernich: I don't see that.  We need to create the proxies after the pod has started, so we need a listener somewhere that does that before running the tests.  What am I missing?

core/src/main/java/org/arquillian/cube/impl/client/container/ContainerConfigurationController.java
aslakknutsen: Correct. configuredBinding() is what we could read out from configuration, e.g. docker-compose/cube format/arquillian.xml/pod.json etc. There is a case for when we need to reconfigure a DeployableContainer to use remapped ports, we need to do that Before the DeploymentContainer is setup, which is Before the Cube is started.

spi/src/main/java/org/arquillian/cube/spi/metadata/HasMappedPorts.java
aslakknutsen:
```java
public class PodCube implements BaseCube {

   HasMappedPorts ports;

   public PodCube() {
     ports = new ForwardedPorts();
     addMetadata(HasMappedPorts.class, ports);
   }

   ..

   public void start() {
     // do normal Cube start
     ports.createProxies();
   }

   public void stop() {
     ports.destroyProxies();

     // do normal Cube stop
   }

   ..
}
```

rcernich: doh!  Thanks!

rcernich:
Hopefully this is better.  It should be pretty easy to remove Binding now as it's only used in DockerCube to generate data for HasPortBindings.  If you're good with this, I can squash and rebase to bring it up to date with master.

aslakknutsen:
This looks great! Excellent work.

rcernich:
did you want me to squash and rebase?

aslakknutsen:
No, that's ok.I've done that already locally. Taking it for a spin :)

aslakknutsen:
Thank you @rcernich ! I updated some failing test cases and fixed a minor order issue (PortMapping called bindings() before state == Start which is not allowd).

rcernich:
Great!  Thanks for polishing it up @aslakknutsen!

Issue #191, opened 2015-11-05
Autodownload Docker Machine


lordofthejars:
It might be interesting that if some kind of flag like `checkDockerExistance` is true, the Arquillian Cube checks if `docker` CLI is on PATH, if not then Cube downloads latest Docker Machine, install it, creates a docker machine with the name specified in `machineName` property and then executes the test.

Probably using Arquillian Spacelift is the best way to achieve this since it has support for caching.

eddumelendez:
Currently working on this and AFAIK the only way to get the last version from github is through the GitHub API using the command below and extracting property *name*

```
curl -s https://api.github.com/repos/docker/machine/releases/latest
```

It is valid for the implementation? are there another way to do that? I thought we could do this:
https://github.com/docker/machine/releases/download/latest/docker-machine_darwin-amd64 instead of https://github.com/docker/machine/releases/download/v0.5.6/docker-machine_darwin-amd64 but didn't work.

aslakknutsen:
Maybe this is the time to start using Arquillian Space lift to deal with the different os's etc; https://github.com/arquillian/arquillian-spacelift

eddumelendez:
Yes currently using it to download, give permissions and run docker-machine :)

aslakknutsen:
Ah, sorry. Had forgotten @lordofthejars original comment.. :)

eddumelendez:
You can see my draft here https://github.com/eddumelendez/arquillian-cube/commit/1df78a0281268b70f4bb9871d2dd7c22dd1f38c6

I will wait the feedback before to submit the PR

aslakknutsen:
@eddumelendez You can just send a pull request. The purpose of a pull request is to give feedback..  It's not set in stone. :)

eddumelendez:
@aslakknutsen yes, just wanted to validate the draft before to send the PR. The next review will be over the PR.

smiklosovic:
@eddumelendez very cool feature, +100 for using Spacelift for that.

eddumelendez:
should this ticket be closed?

lordofthejars:
I think so, thanks :)

PR #266, opened 2016-01-23
Autodownload Docker Machine

eddumelendez:
gh-191

### DIFF
--- a/docker/docker/pom.xml
+++ b/docker/docker/pom.xml
@@ -33,6 +33,11 @@
         </dependency>
 
         <dependency>
+            <groupId>org.arquillian.spacelift</groupId>
+            <artifactId>arquillian-spacelift</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>org.jboss.arquillian.container</groupId>
             <artifactId>arquillian-container-test-impl-base</artifactId> <!-- required to inherit logic of OperatesOnDeploymentAwareProvider -->
             <scope>provided</scope>
--- a/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfiguration.java
+++ b/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfiguration.java
@@ -8,8 +8,12 @@ import java.util.List;
 import java.util.Map;
 import java.util.StringTokenizer;
 
+import org.apache.commons.lang.StringUtils;
 import org.arquillian.cube.docker.impl.client.config.CubeContainers;
+import org.arquillian.cube.docker.impl.model.DockerMachineDistro;
 import org.arquillian.cube.docker.impl.util.ConfigUtil;
+import org.arquillian.cube.docker.impl.util.DockerMachine;
+import org.arquillian.cube.docker.impl.util.HomeResolverUtil;
 
 public class CubeDockerConfiguration {
 
@@ -27,10 +31,13 @@ public class CubeDockerConfiguration {
     public static final String BOOT2DOCKER_PATH = "boot2dockerPath";
     public static final String DOCKER_MACHINE_PATH = "dockerMachinePath";
     public static final String DOCKER_MACHINE_NAME = "machineName";
+    public static final String DOCKER_MACHINE_DRIVER = "machineDriver";
     private static final String AUTO_START_CONTAINERS = "autoStartContainers";
     private static final String DEFINITION_FORMAT = "definitionFormat";
     static final String DIND_RESOLUTION = "dockerInsideDockerResolution";
     private static final String CUBE_ENVIRONMENT = "cube.environment";
+    public static final String DOCKER_MACHINE_CUSTOM_PATH = "dockerMachineCustomPath";
+    public static final String DOCKER_MACHINE_ARQUILLIAN_PATH = "~/.arquillian/machine";
 
     private String dockerServerVersion;
     private String dockerServerUri;
@@ -307,4 +314,16 @@ public class CubeDockerConfiguration {
 
         return content.toString();
     }
+
+    public static String resolveUrl(String machineVersion) {
+        return "https://github.com/docker/machine/releases/download/" + machineVersion + "/" + DockerMachineDistro.resolveDistro();
+    }
+
+    public static String resolveMachinePath(String machineCustomPath, String machineVersion) {
+        if (StringUtils.isBlank(machineCustomPath)) {
+            machineCustomPath = DOCKER_MACHINE_ARQUILLIAN_PATH;
+        }
+        return HomeResolverUtil.resolveHomeDirectoryChar(machineCustomPath + "/" + machineVersion + "/" + DockerMachine.DOCKER_MACHINE_EXEC);
+    }
+
 }
--- a/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfigurator.java
+++ b/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfigurator.java
@@ -2,6 +2,7 @@ package org.arquillian.cube.docker.impl.client;
 
 import java.io.File;
 import java.net.URI;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.logging.Level;
@@ -11,6 +12,7 @@ import org.arquillian.cube.HostUriContext;
 import org.arquillian.cube.docker.impl.util.AbstractCliInternetAddressResolver;
 import org.arquillian.cube.docker.impl.util.Boot2Docker;
 import org.arquillian.cube.docker.impl.util.DockerMachine;
+import org.arquillian.cube.docker.impl.util.GitHubUtil;
 import org.arquillian.cube.docker.impl.util.HomeResolverUtil;
 import org.arquillian.cube.docker.impl.util.Machine;
 import org.arquillian.cube.docker.impl.util.OperatingSystemFamily;
@@ -18,6 +20,12 @@ import org.arquillian.cube.docker.impl.util.OperatingSystemResolver;
 import org.arquillian.cube.docker.impl.util.Top;
 import org.arquillian.cube.impl.util.SystemEnvironmentVariables;
 import org.arquillian.cube.spi.CubeConfiguration;
+import org.arquillian.spacelift.Spacelift;
+import org.arquillian.spacelift.process.Command;
+import org.arquillian.spacelift.process.CommandBuilder;
+import org.arquillian.spacelift.process.ProcessResult;
+import org.arquillian.spacelift.task.net.DownloadTool;
+import org.arquillian.spacelift.task.os.CommandTool;
 import org.jboss.arquillian.config.descriptor.api.ArquillianDescriptor;
 import org.jboss.arquillian.core.api.Instance;
 import org.jboss.arquillian.core.api.InstanceProducer;
@@ -64,6 +72,7 @@ public class CubeDockerConfigurator {
         Map<String, String> config = arquillianDescriptor.extension(EXTENSION_NAME).getExtensionProperties();
         config = resolveSystemEnvironmentVariables(config);
         config = resolveDockerInsideDocker(config);
+        config = resolveDownloadDockerMachine(config);
         config = resolveAutoStartDockerMachine(config);
         config = resolveDefaultDockerMachine(config);
         config = resolveServerUriByOperativeSystem(config);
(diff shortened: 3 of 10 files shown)


docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfiguration.java
```diff
@@ -27,10 +29,14 @@
     public static final String BOOT2DOCKER_PATH = "boot2dockerPath";
     public static final String DOCKER_MACHINE_PATH = "dockerMachinePath";
     public static final String DOCKER_MACHINE_NAME = "machineName";
+    public static final String DOCKER_MACHINE_VERSION = "machineVersion";
+    public static final String DOCKER_MACHINE_DRIVER = "machineDriver";
     private static final String AUTO_START_CONTAINERS = "autoStartContainers";
     private static final String DEFINITION_FORMAT = "definitionFormat";
     static final String DIND_RESOLUTION = "dockerInsideDockerResolution";
     private static final String CUBE_ENVIRONMENT = "cube.environment";
+    private static final String DOCKER_MACHINE_ARQUILLIAN_PATH = "~/.arquillian/machine";
```
smiklosovic: if it is possible, please make this configurable as well so I can put my downloaded machines wherever I want

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfigurator.java
```diff
@@ -87,6 +95,57 @@ private void configure(ArquillianDescriptor arquillianDescriptor) {
...
+
+                Spacelift.task(DownloadTool.class)
+                            .from(machineUrl)
+                            .to(dockerMachineFile)
+                            .execute()
+                            .await();
+                config.put(CubeDockerConfiguration.DOCKER_MACHINE_PATH, dockerMachineFile.getAbsolutePath());
+
+                Command allowExecCmd = new CommandBuilder("chmod", "+x", machineArquillianPath).build();
+                ProcessResult processAllowResult = Spacelift.task(CommandTool.class)
```
smiklosovic: maybe you could check that the chmod exited with exit code 0, there is API for that if you look more into it, via ProcessResult#exitValue and CommandTool#shouldExitWith

eddumelendez:
I think that you can use *dockerMachinePath*

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfiguration.java
eddumelendez: I think that you can use dockerMachinePath

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfigurator.java
smiklosovic: however ... never mind, because if it ever exits with something different then 0, it is going to throw the exception by default

smiklosovic:
Thanks a lot of using Spacelift, if you have any ideas how to improve that tool, feel free to raise them. We will be more than happy to discuss it with you.

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfiguration.java
lordofthejars: What @smiklosovic suggests is that this parameter can be configured from the point of view of caller.

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfigurator.java
```diff
@@ -87,6 +95,57 @@ private void configure(ArquillianDescriptor arquillianDescriptor) {
...
 
+    private Map<String, String> resolveDownloadDockerMachine(Map<String, String> config) {
+        if (config.containsKey(CubeDockerConfiguration.DOCKER_MACHINE_NAME)) {
+            String machineVersion = null;
+            String machineArquillianPath = null;
+            File dockerMachineFile = null;
+            if (config.containsKey(CubeDockerConfiguration.DOCKER_MACHINE_VERSION)) {
+                machineVersion = config.get(CubeDockerConfiguration.DOCKER_MACHINE_VERSION);
+                machineArquillianPath = CubeDockerConfiguration.resolveMachinePath(machineVersion);
+                dockerMachineFile = new File(machineArquillianPath);
```
lordofthejars: I am not sure if this going to work in Windows, I think you should use something like https://github.com/arquillian/arquillian-cube/blob/master/docker/docker/src/main/java/org/arquillian/cube/docker/impl/util/HomeResolverUtil.java

lordofthejars:
Maybe it could be interesting to add a new parameter called automaticUpdate or sth like this which allows to:

1) see if docker machine was downloaded by Cube and not by user (basically check if dockerMachineArquillianPath is empty or not)

2) if docker machine was installed by Cube, check if the version installed is the latest one. If it is ok then don't do anything, but if not let's update it (maintaining the old boot2docker machine).

aslakknutsen:
@lordofthejars Sounds nice. How about creating another issue for that ?

lordofthejars:
sure, I only use this PR for giving some ideas, :)

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfiguration.java
eddumelendez: I have added *dockerMachineCustomPath* in order to use a custom path like /tmp but if it is empty the default value is *~/.arquillian/machine*

eddumelendez:
Thanks for feedback, I have updated the PR. About the new parameter ** automaticUpdate** I will try to implement as a part of other issue.

lordofthejars:
@eddumelendez can you fix the conflict (I think it is because of pom.xml) so it would be really fast to fix it. After that the PR should contain the modification of README as well explaining how it works :)

@aslakknutsen For me everything is correct, if you agree and @eddumelendez fixes minor things, I merge the PR.

docker/docker/pom.xml
```diff
@@ -47,6 +47,11 @@
             <artifactId>shrinkwrap-spi</artifactId>
             <version>1.2.3</version>
         </dependency>
+        <dependency>
+            <groupId>org.arquillian.spacelift</groupId>
+            <artifactId>arquillian-spacelift</artifactId>
+            <version>1.0.0.Alpha8</version>
```
aslakknutsen: Move version to root rom

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfiguration.java
```diff
@@ -307,4 +315,24 @@ private static String resolveFilename(URI uri) {
 
         return content.toString();
     }
+
+    public static String resolveUrl(String machineVersion) {
+        validation(machineVersion, DOCKER_MACHINE_VERSION);
```
aslakknutsen: This force the user to configure a docker-machine version? Any URL that could support 'latest'?

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfiguration.java
```diff
@@ -307,4 +315,24 @@ private static String resolveFilename(URI uri) {
...
+        if (StringUtils.isBlank(machineCustomPath)) {
+            machineCustomPath = DOCKER_MACHINE_ARQUILLIAN_PATH;
+        }
+        return HomeResolverUtil.resolveHomeDirectoryChar(machineCustomPath + "/" + machineVersion + "/" + DOCKER_MACHINE_COMMAND);
+    }
+
+    private static void validation(String attribute, String attributeName) {
+        if (StringUtils.isBlank(attribute)) {
+            throw new IllegalArgumentException("Attribute '" + attributeName +"' is missing.");
```
aslakknutsen: Would be nice to be told why this attribute is important and why i need to specify it. e.g.: Could not attempt docker-machine automatic install, no attributeName defined.

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfiguration.java
eddumelendez: At this time we could use *curl -s https://api.github.com/repos/docker/machine/releases/latest* in order to get the **name** property. For example, *v0.5.6*. Then use the name's value in **machineVersion**.

eddumelendez:
@lordofthejars conflict was solved. 
@aslakknutsen with this update cube will download the latest docker-machine's version.

aslakknutsen:
@lordofthejars with Spacelift now part of Cube, perhaps we should open an issue to rewrite the CommandExecutor stuff to use SpaceLift as well?

lordofthejars:
Completely agree

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfigurator.java
```diff
@@ -87,6 +96,54 @@ private void configure(ArquillianDescriptor arquillianDescriptor) {
         return cubeConfiguration;
     }
 
+    private Map<String, String> resolveDownloadDockerMachine(Map<String, String> config) {
+        if (config.containsKey(CubeDockerConfiguration.DOCKER_MACHINE_NAME)) {
+            String machineVersion = GitHubUtil.getDockerMachineLatestVersion();
```
aslakknutsen: It seems like you will _always_ communicate with GitHub as long as machineName is defined. This needs to be reordered to check if 'docker-machine' exists on path before anything else. Then if it does exist already, move on without doing anything else.

eddumelendez:
@aslakknutsen changes have been applied and PR has been rebased to avoid conflicts in the pom.xml

docker/docker/src/main/java/org/arquillian/cube/docker/impl/util/GitHubUtil.java
```diff
@@ -0,0 +1,36 @@
...
+		WebTarget target = client.target(DOCKER_MACHINE_LATEST_URL);
+		LatestRepository latestRepository = target
+											.request()
+											.accept(MediaType.APPLICATION_JSON_TYPE)
+											.get(LatestRepository.class);
+		return latestRepository.getTagName();
+	}
+
+	public static void main(String[] args) {
+		System.out.println(getDockerMachineLatestVersion());
```
lordofthejars: Can you refactor this main to a JUnit test?

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfigurator.java
```diff
@@ -87,6 +96,60 @@ private void configure(ArquillianDescriptor arquillianDescriptor) {
...
+        }
+        return config;
+    }
+
+    private void printOutput(ProcessResult processResult) {
+        List<String> lines = processResult.output();
+        StringBuilder output = new StringBuilder();
+        for (String line: lines) {
+            output.append(line);
+            output.append("/n");
```
lordofthejars: Can you use System.lineterminator()?

docker/docker/src/main/java/org/arquillian/cube/docker/impl/util/GitHubUtil.java
```diff
@@ -0,0 +1,36 @@
...
+	private static final String DOCKER_MACHINE_LATEST_URL = "https://api.github.com/repos/docker/machine/releases/latest";
+
+	private static Client getClient() {
+		ClientConfig clientConfig = new ClientConfig();
+		clientConfig.register(JacksonJsonProvider.class);
+		return ClientBuilder.newClient(clientConfig);
+	}
+
+	public static String getDockerMachineLatestVersion() {
+		Client client = getClient();
```
lordofthejars: We usually try to avoid using transitive dependencies that other projects add like JAXRS or Jackson (in fact we only use docker-java) but other communications we do with REST (for example in Jenkins) we use HttpUrlConnection, I think @aslakknutsen will agree with me that a refactor of this part would be awesome.

aslakknutsen: Preferred yes. Technically we already depend on all of them transitively via docker-java, and this is a client side only request.

lordofthejars: Yes I know we could use it but since it is only one request and it should be fast to achieve it without using JAXRS client I think it is worth to do it.

aslakknutsen: Agree. It's a simple request.

docker/docker/src/main/java/org/arquillian/cube/docker/impl/util/GitHubUtil.java
eddumelendez: Sorry, forgot to remove this :tired_face:

lordofthejars: no worries, it happens to best families :)

eddumelendez:
Thanks for feedback. PR updated.

lordofthejars:
Thank you very much for contributing and for your patience fixing all the comments we did :) A pleasure to work with you.

Issue #254, opened 2016-01-16
Can I use waiting strategies in docker-compose.yml case?


smiklosovic:
I have Docker compose YAML and I want to use that one but when I do, it fails to start containers because of timeout. So I wanted to use some polling strategy and I can not pass it to docker-compose.yml as they seem to be ignored.

I tried mixed approach - I put them into dockerContainers property but that obviously does not do anything since I have COMPOSE  definition format.

Am I forced to ignore my docker-compose.yml completely and construct something very similar in dockerContainers property where polling strategies can be used successfully?

aslakknutsen:
We were discussing this scenario at some conference, and added support for among others multipe files you can include.. but I forget what the specific fix to this issue was, @lordofthejars ?

lordofthejars:
I think it is not possible. By default it uses `ss` command. Maybe we should be able to extract some properties outside the orchestration without affecting all the instances but only the chosen ones.

aslakknutsen:
@lordofthejars yea, this is where we need to be able to enrich the docker-compose configuraton with Cube specific options.. somehow.

smiklosovic:
Maybe this is offtopic but the reason I am asking is that "normal" waiting strategies are not enough in my case. I have MySQL server in Docker and I am doing some initialization in RUN /my/script.sh when I am setting priviledges to the mysql database, loading sql scripts and so on to the temporary database which is running at port 3307 and once I am done, I kill that instance and I restart that server, now on port 3306.

Cube has in arquillian.xml mysql Docker descriptor in CUBE format and I am exposing 3306 and I am doing port binding as well..

The thing is that when Cube is checking if 3306 is open, it seems to me that it checks the output of the "ss" command executed in container and even that port is really in that ss output, you can not connect to it - the fact that it is portforwarded and in ss command output does not say anything about the fact that underlying service is fully up and running - in my case I am doing additional checks e.g. I am trying to create java.sql.Connection to the server from my side to be sure that it is really accessible.

Maybe you should think about pluggable waiting strategies ...

PR #291, opened 2016-02-15
allows to use custom cube properties to docker-compose

lordofthejars:
allows to use custom cube properties to docker-compose defined containers.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -224,6 +224,9 @@ Let's see valid attributes:
 |definitionFormat
 |Sets the format of content expressed in `dockerContainers` attribute or in file set in `dockerContainersFile`. It can contain two possible values _CUBE_ (default one in case of not set) to indicate that content is written following <<cube-format, Arquillian Cube>> format or _COMPOSE_ to indicate that content is written following <<docker-compose-format, Docker Compose>> format.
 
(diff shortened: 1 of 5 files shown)


README.adoc
```diff
@@ -527,6 +530,31 @@ beforeStop:
...
+        strategy: polling
+      beforeStop:
+        - copy:
+            from: /test
+            to: /tmp";
+</property>
+----
+
+Previous snippet would override `await` strategy to polling and set a `beforeStop` event in cube container named `tomcat` defined in a previous docker-compose definition.
+Also notice that the `image` property here does nothing, since ony CUBE custom properties are overriden.
```
aslakknutsen: Shouldn't this be removed from the example and only explained in text? Are there other Cube specific configurations that we should mention besides Await and BeforeStop ?

lordofthejars: You mean `image`. Yes I though the same but I found a way to enforce the message to reader that only custom properties are updated.

aslakknutsen: Yeah, I understand the point. Just not sure ppl actually read the text :)

lordofthejars: Ok I remove then.

lordofthejars:
Fixes #254

PR #295, opened 2016-02-23
Upgrade to docker-java 3.0.0-RC1

eddumelendez:
See gh-294

### DIFF
--- a/docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/DockerClientExecutor.java
+++ b/docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/DockerClientExecutor.java
@@ -25,7 +25,9 @@ import java.util.regex.Pattern;
 
 import javax.ws.rs.ProcessingException;
 
-import com.github.dockerjava.api.ConflictException;
+import com.github.dockerjava.api.exception.ConflictException;
+import com.github.dockerjava.core.command.ExecStartResultCallback;
+import com.github.dockerjava.core.command.WaitContainerResultCallback;
 import org.apache.http.conn.UnsupportedSchemeException;
 import org.arquillian.cube.TopContainer;
 import org.arquillian.cube.docker.impl.client.CubeDockerConfiguration;
@@ -37,7 +39,7 @@ import org.arquillian.cube.docker.impl.util.BindingUtil;
 import org.arquillian.cube.docker.impl.util.HomeResolverUtil;
 
 import com.github.dockerjava.api.DockerClient;
-import com.github.dockerjava.api.NotFoundException;
+import com.github.dockerjava.api.exception.NotFoundException;
 import com.github.dockerjava.api.command.BuildImageCmd;
 import com.github.dockerjava.api.command.CreateContainerCmd;
 import com.github.dockerjava.api.command.ExecCreateCmdResponse;
@@ -138,17 +140,17 @@ public class DockerClientExecutor {
         dockerUri = URI.create(dockerServerUri);
         dockerServerIp = cubeConfiguration.getDockerServerIp();
 
-        configBuilder.withVersion(cubeConfiguration.getDockerServerVersion()).withUri(dockerUri.toString());
+        configBuilder.withApiVersion(cubeConfiguration.getDockerServerVersion()).withDockerHost(dockerUri.toString());
         if(cubeConfiguration.getUsername() != null) {
-            configBuilder.withUsername(cubeConfiguration.getUsername());
+            configBuilder.withRegistryUsername(cubeConfiguration.getUsername());
         }
 
         if(cubeConfiguration.getPassword() != null) {
-            configBuilder.withPassword(cubeConfiguration.getPassword());
+            configBuilder.withRegistryPassword(cubeConfiguration.getPassword());
         }
 
         if(cubeConfiguration.getEmail() != null) {
-            configBuilder.withEmail(cubeConfiguration.getEmail());
+            configBuilder.withRegistryEmail(cubeConfiguration.getEmail());
         }
 
         if(cubeConfiguration.getCertPath() != null) {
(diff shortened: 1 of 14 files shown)


docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/DockerClientExecutor.java
```diff
@@ -554,10 +556,15 @@ public void pullImage(String imageName) {
...
-        InputStream consoleOutputStream = dockerClient.execStartCmd(execCreateCmdResponse.getId()).withDetach(false)
+                .withAttachStdout(true).withAttachStdin(false).withAttachStderr(false).withTty(false).withCmd(commands)
                 .exec();
+        try {
+            dockerClient.execStartCmd(execCreateCmdResponse.getId()).withDetach(false)
+                    .exec(new ExecStartResultCallback(System.out, System.err)).awaitCompletion();
+        } catch (InterruptedException e) {
+            return "";
+        }
+        InputStream consoleOutputStream = dockerClient.copyArchiveFromContainerCmd(containerId, "/execStartTest.log").exec();
```
lordofthejars: I have seen in the examples of docker-java 3 that now you need to get the `execStartTest.log`. I am not sure the implications that this going to have :(. WhatI have seen in the examples is that you need to call `touch` for creating the file (https://github.com/docker-java/docker-java/blob/9ec46f44374c434a9b6132cd8611e2c8e80b8c69/src/test/java/com/github/dockerjava/netty/exec/ExecStartCmdExecTest.java#L86) but you have not called, maybe it is not mandatory at all. Have you run the ftests to see if everything works as expected?

lordofthejars: This API breaks the backcompatibility from major version to major version which is not good at all for we (the integrators). Have you tried to attach the stdin? https://github.com/docker-java/docker-java/blob/9ec46f44374c434a9b6132cd8611e2c8e80b8c69/src/test/java/com/github/dockerjava/netty/exec/ExecStartCmdExecTest.java#L102

aslakknutsen: I think you need to pass a PipedOutputstream to ExecStartResultCallback as 'System.out', then read the InputStream from that. The execStartTest.log is only added in the Docker-Java test case as a way to verify that the command was actually executed in the container as the command they execute is "echo > execStartTest.log"

aslakknutsen: Or actually, since we read in the whole input stream as a String and return it, we might as well pass in a ByteArrayOutputStream to the ExecStartResultCallback and return byteArrayOutput.toString()

eddumelendez: Using PipedOutputstream got IOException: Pipe not connected. I just use ByteArrayOutputStream and looks like this:

```java
public String execStart(String containerId, String... commands) {
        ExecCreateCmdResponse execCreateCmdResponse = this.dockerClient.execCreateCmd(containerId)
                .withAttachStdout(true).withAttachStdin(true).withAttachStderr(false).withTty(false).withCmd(commands)
                .exec();

        OutputStream outputStream = new ByteArrayOutputStream();
        try {
            dockerClient.execStartCmd(execCreateCmdResponse.getId()).withDetach(false)
                    .exec(new ExecStartResultCallback(outputStream, System.err)).awaitCompletion();
        } catch (InterruptedException e) {
            return "";
        }

        return outputStream.toString();
    }
```

containerless's test was ok but ftest throws errors:

```
Caused by: org.yaml.snakeyaml.parser.ParserException: while parsing a flow sequence
 in 'reader', line 5, column 20:
                  env: [TOMCAT_PASS=mypass, "CATALINA_O ... 
                       ^
expected ',' or ']', but got FlowMappingStart
 in 'reader', line 5, column 136:
     ... PTS=-Djava.rmi.server.hostname=${docker.tomcat.host} -Dcom.sun.m ... 
```

I have not seen `docker.tomcat.host` set in any xml. Am I missing something?

lordofthejars: It doesn't seems a problem with docker-java 3. Have you tried to run the ftest-docker-machine?

smiklosovic:
3.0.0-RC2 was just released

lordofthejars:
@eddumelendez I have seen you updated all the code, does it works now?

eddumelendez:
@lordofthejars I just got error in `ftest` as I said before and the same error for `ftest-boot2docker`. In `ftest-docker-containerobject` got an error in `JSchTest` due to bad private key but doesn't seem to be an error due to docker-java 3. `ftest-docker-compose`, `ftest-docker-machine`, `ftest-standalone`, `ftest-standalone-autostart` runs ok.

lordofthejars:
About test of containerobject, you are right, seems a bug in jsch. What I
will do is update to M2, check that docker-machine example runs correctly
(the one that uses the ping-pong container) and if ok, merge it. Any
thoughts on this Aslak?

lordofthejars:
@eddumelendez can you resolve the merge problem. I think it is a really stupid thing, and update to M2? After that we will merge the PR and start to seeing what is going on :)

eddumelendez:
@lordofthejars upgrading jcabi-ssh from1.5 to 1.5.2 didn't work.

lordofthejars:
@eddumelendez I know, forget about jsch error, I have contacted the maintainers and they told me that should work and they need to take a look.

lordofthejars:
Thank you very much for this contribution, is a major contribution for all the community.

eddumelendez:
Thanks @lordofthejars and sorry at the same time I didn't detect one change when update the docker-java to RC3. I will submit another PR with the little change.

lordofthejars:
It is ok, no worries :) Thanks.

smiklosovic:
:+1:

smiklosovic:
I can't wait to bump from Alpha7 :)

PR #301, opened 2016-02-24
correct config overrides in OpenShiftClientCreator

rcernich:
The overrides were being overridden by the defaults. :(


aslakknutsen:
Thank you :)

Issue #293, opened 2016-02-19
Add ShutdownHook


lordofthejars:
Add some ShutdownHook that when the JVM is killed (by IDE) it sees if any container is still up (started by cube obviously) and kill them

PR #302, opened 2016-02-25
adds shutdown hook to stop unclosed containers

lordofthejars:
fixes issue #293

### DIFF
--- a/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/ForceStopDockerContainersShutdownHook.java
+++ b/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/ForceStopDockerContainersShutdownHook.java
@@ -1,0 +1,32 @@
+package org.arquillian.cube.docker.impl.client;
+
+import org.arquillian.cube.docker.impl.docker.DockerClientExecutor;
+import org.arquillian.cube.spi.Cube;
+import org.arquillian.cube.spi.CubeRegistry;
+import org.jboss.arquillian.core.api.annotation.Observes;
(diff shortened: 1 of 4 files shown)


docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/ForceStopDockerContainersShutdownHook.java
```diff
@@ -0,0 +1,32 @@
...
+        Runtime.getRuntime().addShutdownHook(new Thread() {
+            @Override
+            public void run() {
+                final List<Cube<?>> cubes = cubeRegistry.getCubes();
+                for (Cube cube : cubes) {
+                    // If container is started, and we are exiting we need to stop it.
+                    // Notice that in case of STARTORCONNECT and STARTORCONNECTANDLEAVE the state is PRE_RUNNING
+                    // so they are not going to be stopped
+                    if (Cube.State.STARTED.equals(cube.state())) {
+                        dockerClientExecutor.stopContainer(cube.getId());
```
aslakknutsen: * Why in docker-impl only? Why not a core concern?
* Why use docketClientExecutor.stop.. vs cube.stop?

lordofthejars: So you mean to move this hook to core right? Yes it has sense, and I agree I should call `cube.stop`

aslakknutsen: yes

PR #312, opened 2016-03-07
removes ping operation since it is expensive 

### LINKED ISSUE #305: Remove ping operation
Remove ping operation since it is useless and takes like 1 second to be executed.

lordofthejars:
removes ping operation since it is expensive  and a bit useless in our case. Fixes #305


aslakknutsen:
Do you know why it's expensive? It's just a single minimal REST call.

lordofthejars:
Yes I know but added a delay in my case of 700ms. And it does nothing since
the error will come when you try to create the container, so nit a big deal

PR #313, opened 2016-03-07
updates to commons-lang3 + fixes bug regarding extending resolution

lordofthejars:


aslakknutsen:
The property substitution happens before the file is beeing read as YAML, right? So technically you could substitute whole YAML sections if you wanted to, not just single values.

lordofthejars:
Technically yes. We can change this but I think that in terms of
performance is fastest this approach and also easier to implement.

Issue #310, opened 2016-03-04
InternalServerErrorException: http: Hijack is incompatible with use of CloseNotifier


foxlegend:
Hello there,

I had a new issue when building the image with Cube.
In fact, it seems to be a known issue with docker itself I could not explain.

When building an image, Cube seems to perform the following Docker API queries:

* ping (OK),
* build the image with Dockerfile directory content (OK),
* create the container (OK),
* start the container (OK),
* inspect the container (OK),
* create the await command `ss -lntu | grep … etc.` (OK),
* execute the command (KO).

When the command execution query is sent to my docker host, I obtain:

* the `com.github.dockerjava.api.InternalServerErrorException: http: Hijack is incompatible with use of CloseNotifier` from my project side,
* a `500 Internal Server Error` with message `http: Hijack is incompatible with use of CloseNotifier` from docker host logs.

Additional information I can provide:

* full logs from project: [arquillian_cube_full_logs.txt](https://github.com/arquillian/arquillian-cube/files/158804/arquillian_cube_full_logs.txt),
* docker logs: [docker_full_logs.txt](https://github.com/arquillian/arquillian-cube/files/158809/docker_full_logs.txt)
* arquillian.xml: [arquillian.xml.txt](https://github.com/arquillian/arquillian-cube/files/158812/arquillian.xml.txt) (please note that removing the database image results in the same issue),
* the same issue occured with Docker 1.8.2 and 1.10.0.

I found that the issue was something already "known":

* [Hijack is incompatible with use of CloseNotifier](https://github.com/docker/docker/issues/12845),
* [Close connection after building or pulling](https://github.com/docker/compose/pull/1374),
* [Dockerpy - Making the container with Python](http://harpoon.readthedocs.org/en/latest/docs/building_harpoon/section1/introduce_dockerpy.html#making-the-container-with-python).

It seems there is a bug in go, causing Docker daemon issues when using a recycled connection.

To try a workaround, I shadowed the `DockerClientExecutor` in order to release the `dockerClient` instance properly (`dockerClient.close()`) after the build request (in the `buildImage` method) and create a new instance of `DockerClient`. Everything is working for me now with such a solution.

Do no hesitate if you need more information about the issue (or if I have not been very understandable ;))!

David.

lordofthejars:
Interesting, I have never seen this error :), one question, do you think it is possible to do a JUnit test using `docker-java` to reproduce the problem?

foxlegend:
[...]

PR #321, opened 2016-03-16
resolves issue #310 by rebuilding dockerClient after a Docker build command

foxlegend:
Actually when trying to build an image with Arquillian Cube, docker host throwed
the exception message: http: Hijack is incompatible with use of CloseNotifier.
It seems to be a problem with go and should be fixed in go 1.6 (and maybe in
Docker 1.11.0).
For now, closing and recreating the Docker client after a call to the build Docker
command seems to solves the problem.

I let you have a look to the code. If you have a cleaner solution I could write the enhancement.

I'm sorry for the delay…

### DIFF
--- a/docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/DockerClientExecutor.java
+++ b/docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/DockerClientExecutor.java
@@ -130,6 +130,7 @@ public class DockerClientExecutor {
     private CubeDockerConfiguration cubeConfiguration;
     private final URI dockerUri;
     private final String dockerServerIp;
+    private DockerClientConfig dockerClientConfig;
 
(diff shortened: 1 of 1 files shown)


lordofthejars:
No worries for the delay, I just only asking to know when we could release a new version fixing this :+1:  Thank you very very much for your collaboration.

lordofthejars:
Thank you so much for this workaround.

lordofthejars:
BTW do you think it would be possible to have a hello world project that reproduces it? So it will be a way to see if it is fixed in the future or not.

foxlegend:
Yes, I think it would be possible. I'll create such a project when I have some time ;)

lordofthejars:
Thank you so much, basically to know if it is fixed on docker and go or not
:)

El dv., 18 març 2016 a les 15:54, David Pequegnot (<notifications@github.com>)
va escriure:

PR #325, opened 2016-03-17
introduction of dockerTlsVerify property

smiklosovic:
@lordofthejars @aslakknutsen 

this works for me, both for machine and natively, could you test boot2docker?

It is true by default. That means that you do not need to set this when you are on Mac or Windows. In case you are on Linux, you would have to explicitly set it to false.

There are not tests for it. It would be cool if you could add them where necessary. I do not know how to deal with it.

### DIFF
--- a/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfiguration.java
+++ b/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfiguration.java
@@ -24,6 +24,7 @@ public class CubeDockerConfiguration {
     private static final String PASSWORD =  "password";
     private static final String EMAIL = "email";
     public static final String CERT_PATH = "certPath";
+    public static final String TLS_VERIFY = "tlsVerify";
     private static final String DOCKER_CONTAINERS = "dockerContainers";
     private static final String DOCKER_CONTAINERS_FILE = "dockerContainersFile";
     private static final String DOCKER_CONTAINERS_FILES = "dockerContainersFiles";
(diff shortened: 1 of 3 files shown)


smiklosovic:
however it could be set automatically, if you are or mac or windows it will be true and if you are on linux it would be true only in case you run against machine so you would not need to deal with it at all

lordofthejars:
Can you implement the logic you mention so it can be autnatically set? So
we avoid one parameter.

smiklosovic:
yes, tomorrow

smiklosovic:
@lordofthejars review it please

smiklosovic:
I did not know how to pass it from configurator to executor more property, I can not do that resolution in executor because i do not have access to descriptor to extract properties from there and to that logic in it so I just did public static field which gets its value when doing other resolvings.

lordofthejars:
I get the idea, I will change the OS detection to use our own class basically to try to not tight so much in an external dependency, for just in case some day we want to remove it (probably not but since we have already developed our system :)) The class is https://github.com/arquillian/arquillian-cube/blob/master/docker/docker/src/main/java/org/arquillian/cube/docker/impl/util/OperatingSystemResolver.java and you already have the instance here: https://github.com/arquillian/arquillian-cube/blob/master/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfigurator.java#L59

and instead of public static you can do something similar as we did in `certPath` https://github.com/arquillian/arquillian-cube/blob/master/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfigurator.java#L184 by adding a property in CubeDockerConfiguration.

smiklosovic:
@lordofthejars review please

smiklosovic:
you can have dockerMachine set but when you do not have dockerHost set as well which gets expanded automatically, it will start just docker containers locally.

This is desired behaviour. I have one arquillian.xml for both scenarios and I have machineName set there. I dynamically put the value of docker server uri either to "dockerHost" or to "localhost". 

As I said, if I set it to "localhost", even I do have machineName specified, it will choose native execution path.

PR #314, opened 2016-03-10
fixes #304 by adding clean property.

lordofthejars:

PR #327, opened 2016-03-27
Fixed cleaning on container start

tbarath:
Fixes two issue with PR #314:
* clean property is ignored
* fails with exception in case container was already stopped


lordofthejars:
Thank you very much for your help.

PR #329, opened 2016-03-28
Add support to docker-compose.yml version 2

eddumelendez:
See gh-328

### DIFF
--- a/docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/compose/ComposeBuilder.java
+++ b/docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/compose/ComposeBuilder.java
@@ -1,0 +1,72 @@
+package org.arquillian.cube.docker.impl.docker.compose;
+
+import static org.arquillian.cube.docker.impl.util.YamlUtil.asMap;
+
+import java.nio.file.Path;
+import java.util.Map;
+import java.util.Set;
+
+import org.arquillian.cube.docker.impl.client.config.CubeContainer;
+import org.arquillian.cube.docker.impl.client.config.DockerCompositions;
+import org.arquillian.cube.docker.impl.client.config.Network;
+
+public class ComposeBuilder {
+
+	private DockerCompositions configuration;
+
+	private static final String SERVICES = "services";
+	private static final String NETWORKS = "networks";
+
+	private static final String NETWORK_NAME_SUFFIX = "_default";
+
+	private Path dockerComposeRootLocation;
+
+	ComposeBuilder(Path dockerComposeRootLocation) {
+		this(dockerComposeRootLocation, new DockerCompositions());
+	}
+
+	private ComposeBuilder(Path dockerComposeRootLocation,
+	                       DockerCompositions configuration) {
+		this.dockerComposeRootLocation = dockerComposeRootLocation;
+		this.configuration = configuration;
+	}
+
+	@SuppressWarnings("unchecked")
+	public DockerCompositions build(Map<String, Object> dockerComposeContainerDefinition) {
+		if (dockerComposeContainerDefinition.containsKey(SERVICES)) {
+			Map<String, Object> servicesDefinition = asMap(
+					dockerComposeContainerDefinition, SERVICES);
+			Set<String> keys = servicesDefinition.keySet();
+			for (String key : keys) {
+				ContainerBuilder services = new ContainerBuilder(this.dockerComposeRootLocation);
+				CubeContainer cubeContainer = services.build(asMap(servicesDefinition, key));
+				if (!dockerComposeContainerDefinition.containsKey(NETWORKS)) {
+					String networkName = getDefaultNetworkName();
+					cubeContainer.setNetworkMode(networkName);
+				}
+				this.configuration.add(key, cubeContainer);
+			}
+		}
+		if (dockerComposeContainerDefinition.containsKey(NETWORKS)) {
+			Map<String, Object> networksDefinition = asMap(
+					dockerComposeContainerDefinition, NETWORKS);
+			Set<String> keys = networksDefinition.keySet();
+			for (String key : keys) {
+				NetworkBuilder networks = new NetworkBuilder();
+				Network network = networks.build(asMap(networksDefinition, key));
+				this.configuration.add(key, network);
+			}
+		}  else {
+			String networkName = getDefaultNetworkName();
+			NetworkBuilder networkBuilder = new NetworkBuilder();
+			Network network = networkBuilder.withDefaultDriver().build();
+			this.configuration.add(networkName, network);
+		}
+		return this.configuration;
+	}
+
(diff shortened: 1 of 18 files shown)


docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/compose/DockerComposeConverter.java
```diff
@@ -61,10 +61,19 @@ public CubeContainers convert() {
         CubeContainers cubeContainers = new CubeContainers();
 
         Set<String> names = dockerComposeDefinitionMap.keySet();
-
-        for(String name : names) {
-            CubeContainer cubeContainer = convertContainer(asMap(dockerComposeDefinitionMap, name));
-            cubeContainers.add(name, cubeContainer);
+        boolean isV2 = names.contains("version") && dockerComposeDefinitionMap.get("version").equals("2") && names.contains("services");
```
lordofthejars: it is always better to do something like "2".equals(....) to avoid any NPE

docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/compose/DockerComposeConverter.java
```diff
@@ -61,10 +61,19 @@ public CubeContainers convert() {
...
+        if (isV2) {
+            Map<String, Object> dockerComposeV2DefinitionMap =  asMap(dockerComposeDefinitionMap, "services");
+            Set<String> services = dockerComposeV2DefinitionMap.keySet();
+            for(String name : services) {
+                CubeContainer cubeContainer = convertContainer(asMap(dockerComposeV2DefinitionMap, name));
```
lordofthejars: I think we should need to create a new factory for docker compose 2 changes, notice that there are a lot of things we should adapt. For example: 
[ ] https://docs.docker.com/compose/compose-file/#build
[ ] adding network/volumes
[ ] in docker-compose v2 if no network is provided one is created by default
[ ] you can use build and image at the same time
[ ] we need to implement label support

Info at: https://github.com/docker/compose/releases/tag/1.6.0
Info about networking: https://docs.docker.com/compose/networking/

lordofthejars:
I have added some comments, I am not saying that we need to provide 100% compatibility with docker compose v2 yet, but I have added some checks that I think it might be a must for our first release of v2 support. WDYT? I think we can use the same PR for adding them. Has it sense? Of course we can accept this PR and open new issues for each one new feature.

eddumelendez:
@lordofthejars Agree. I think that can work in the new builder for docker-compose v2. We may can not add the full support but step by step we can go far far away. I will be adding new commits in the next days.

lordofthejars:
Sure, no problem, as I said I am not saying to have 100% compatible, but doing in a different class so we can work on that class independently it would be really helpful, and I think that having network, new build format and labels implemented it would be enough for now.

Notice the important point that now we need to create a network automatically if no network is provided. This is something new and must only occur in case of version 2. Also since we don't have support for network it will also implies adding network support for our internal format. If you can do it then no problem, if you see that it might be a bit complicated, I can do this part :)

eddumelendez:
@lordofthejars I have added a new commit with the new factory class for Docker Compose v2, it is my first draft but I will improve it in the next days.

lordofthejars:
@eddumelendez when you think this is ready, please ping me so we can review it and merge ir :)

lordofthejars:
@eddumelendez No rush, only one thing to take into account now. Using network + docker-compose means `link` is optional. Currently we are using `links` for knowing the dependencies between services. With docker compose 2 now there is the `depends_on` that can be used instead of links to set the order. This is something that in the logics where we are ordering services should take into consideration. https://docs.docker.com/compose/startup-order/

lordofthejars:
Hi @eddumelendez how is going with this? It is not to put pressure it is only a question to know, that's all 😃

eddumelendez:
@lordofthejars it is not really yet. I little busy these days but hope to send the final draft the next week.

lordofthejars:
No worries, and no rush, it was only to know it :) this is not a company hehehe

lordofthejars:
@eddumelendez working on volumes support.

eddumelendez:
@lordofthejars great! Then, I will work in the network support. I have updated the PR.

eddumelendez:
@lordofthejars PR is almost done. I just have one question about volumes. Should it be managed as the below configuration? Currently I have an issue running `ftest-docker-machine-volume`

```
volumes:
  myvolume:
    driver: local
tomcat:
  image: tutum/tomcat:7.0
  exposedPorts: [8089/tcp]
  env: [TOMCAT_PASS=mypass, "CATALINA_OPTS=-Djava.security.egd=file:/dev/./urandom", JAVA_OPTS=-Djava.rmi.server.hostname=dockerServerIp -Dcom.sun.management.jmxremote.rmi.port=8088 -Dcom.sun.management.jmxremote.port=8089 -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.authenticate=false]
  portBindings: [8089/tcp, 8088/tcp, 8081->8080/tcp]
  volumeDriver: myvolume
```

lordofthejars:
What I would do is submitting with only network support, then reciew and
merge and release an alpha to let users start testing it, and meanwhile
implement volumes support. I think that what users expects for testing is
network so we can provide as soon as possible

eddumelendez:
@lordofthejars PR updated.

docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/compose/ComposeBuilder.java
```diff
@@ -0,0 +1,57 @@
...
+			Map<String, Object> servicesDefinition = asMap(
+					dockerComposeContainerDefinition, SERVICES);
+			Set<String> keys = servicesDefinition.keySet();
+			for (String key : keys) {
+				ContainerBuilder services = new ContainerBuilder(this.dockerComposeRootLocation);
+				CubeContainer cubeContainer = services.build(asMap(servicesDefinition, key));
+				this.configuration.add(key, cubeContainer);
+			}
+		}
+		if (dockerComposeContainerDefinition.containsKey(NETWORKS)) {
```
lordofthejars: According to docker compose v2, if there is no network defined, one default is provided (https://docs.docker.com/compose/networking/) So what I would do is:

if no network provide, create one (it means creating a new network definition automatically) with the `current directory_default` and more important iterate over all cube definitions and provide a networkMode field with the name of this network.

lordofthejars:
Nice we are almost done.

eddumelendez:
@lordofthejars I have added the changes to create a default network and attach it in the networkMode

lordofthejars:
Looks good to me :) Let's wait for @aslakknutsen review and we are in :) Thank you so much for your help.

aslakknutsen:
lgtm :fireworks:

Issue #330, opened 2016-03-29
In case docker-machine binary is not installed, Cube fails


smiklosovic:
@eddumelendez @lordofthejars 

We are using Cube in CI environment where we have only native slaves so we do not use docker-machine.

Cube Alpha10 fails on this scenario because it determines if docker-machine is present or not by trying to execute that binary here (1)

The problem is that "false" is returned in case it catches IllegalArgumentException. However, it fails on this:

```
error	29-Mar-2016 12:14:51	Caused by: org.arquillian.spacelift.execution.ExecutionException: Execution of a task failed. Execution of a task failed. Unable to execute task SpawnProcessTask
error	29-Mar-2016 12:14:51		at org.arquillian.spacelift.execution.ExecutionException.prependMessage(ExecutionException.java:71)
error	29-Mar-2016 12:14:51		at org.arquillian.spacelift.execution.impl.FutureBasedExecution.unwrapException(FutureBasedExecution.java:173)
error	29-Mar-2016 12:14:51		at org.arquillian.spacelift.execution.impl.FutureBasedExecution.await(FutureBasedExecution.java:101)
error	29-Mar-2016 12:14:51		at org.arquillian.spacelift.task.os.ProcessBasedExecution.await(ProcessBasedExecution.java:149)
error	29-Mar-2016 12:14:51		at org.arquillian.cube.docker.impl.util.CommandLineExecutor.execCommandAsArray(CommandLineExecutor.java:22)
error	29-Mar-2016 12:14:51		at org.arquillian.cube.docker.impl.util.CommandLineExecutor.execCommand(CommandLineExecutor.java:14)
error	29-Mar-2016 12:14:51		at org.arquillian.cube.docker.impl.util.DockerMachine.isDockerMachineInstalled(DockerMachine.java:82)
error	29-Mar-2016 12:14:51		at org.arquillian.cube.docker.impl.client.CubeDockerConfigurator.resolveDownloadDockerMachine(CubeDockerConfigurator.java:98)
```

Hence you have to catch `ExecutionException` instead of `IllegalArgumentException`.

I would prefferably return `false` on any exception caught.

(1) https://github.com/arquillian/arquillian-cube/blob/master/docker/docker/src/main/java/org/arquillian/cube/docker/impl/util/DockerMachine.java#L84

smiklosovic:
Momentally, workaround is to just provide docker-machine binary to have Cube happy

PR #331, opened 2016-03-29
caught exception in case docker-machine is not invoked successfully

smiklosovic:
tracks #330


lordofthejars:
I think this is a good way to deal it. Catch the global exception since it means something wrong had happened.

PR #332, opened 2016-04-03
Restore ability to connect Docker server without TLS

tbarath:
With upgraded to docker-java 3.0.0-RC3 (PR #295) Cube doesn't accept `http://` and `https://` in `serverUri` parameter anymore. PR #325 added `tlsVerify` parameter but value set by user in arquillian.xml file is overwritten via automatic resolution. The net result is that you can't connect to remote Docker host without TLS.

This PR brings back http and https scheme support and allow manual setting of tlsVerify attribute as well. (Of course tcp and unix schemes remain supported.)

In case the serverUri has http or https scheme, the tlsVerify attribute is set automatically to false or true respectively. Also the URI schema is changed to `tcp://` to make docker-java happy.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -201,7 +201,7 @@ Let's see valid attributes:
 |Version of REST API provided by_Docker_ server. You should check on the _Docker_ site which version of REST API is shipped inside installed _Docker_ service. This field is not mandatory and if it's not set the default provided version from _docker-java_ will be used.
 
 |serverUri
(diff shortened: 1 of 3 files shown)


lordofthejars:
Hi, thank you very much for your help on improving Arquillian Cube and to add support to cover this missing feature of using `http`.

tbarath:
It was my pleasure. Prepare for more to come. :)

lordofthejars:
Thank you very much :)

PR #334, opened 2016-04-17
Add log scanning await strategy

tbarath:
Initial implementation of a new await strategy that scans the container log for specified pattern to detect service startup. The pattern is matched to every log line. It can be a plain text or regular expression.

This strategy can be used when there is no port to ping or connect to the port successfully doesn't mean the service is fully initialized.

I've tested it only on Windows but it should work on other platforms too.

I don't know if this is sound enough to be included in Cube. I just wanted to show the idea. Any criticism are welcomed. :smile:

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -198,7 +198,7 @@ Let's see valid attributes:
 [cols="2*"]
 |===
 |serverVersion
-|Version of REST API provided by_Docker_ server. You should check on the _Docker_ site which version of REST API is shipped inside installed _Docker_ service. This field is not mandatory and if it's not set the default provided version from _docker-java_ will be used.
+|Version of REST API provided by _Docker_ server. You should check on the _Docker_ site which version of REST API is shipped inside installed _Docker_ service. This field is not mandatory and if it's not set the default provided version from _docker-java_ will be used.
 
 |serverUri
 |Uri of _Docker_ server. If the _Docker_ server is running natively on Linux then this will be an URI pointing to _localhost_ docker host but if you are using _Boot2Docker_ or a remote _Docker_ server then the URI should be changed to point to the _Docker_ remote _URI_. It can be a unix socket URI as well in case you are running _Docker_ on Linux (+unix:///var/run/docker.sock+). If the URI has `http://` or `https://` scheme, the `tlsVerify` attribute will be set by Cube to `false` or `true` respectively. Also you can read at <<automatic-resolution, this section>> about automatic resolution of serverUri parameter. Also you can use `DOCKER_HOST` java property or system environment to set this parameter.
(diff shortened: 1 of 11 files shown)


lordofthejars:
Thank you very much for this PR. It looks pretty well and fills the gap of a use case of Arquillian Cube.

Much appreciated.

smiklosovic:
Hi @tbarath , while I very appreciate your effort and your solution can indeed work, I have a strong feeling that this solution is not robust enough. Don't take me wrong, really, but I had very similar problem and in case there is something to ping (in case there is not, your approach is basically the only reasonable one), I would avoid to parse a log whenever I can.

I found parsing a log file in order to see if some service is up to be quite cumbersome in the long run. That logging output can change just by mere updating of the version of that service / container and I just can not depend on it in production.

My solution was to create API which uses libraries which communicates with that services directly - again, in case of port is open, I check, for example in case of MySQL, if I am able to make a JDBC connection and that is my ultimate way how to check that some service is fully up.

I am calling for pluggable service-oriented checks for quite a long time but nobody is working on it and my solution is IP as of now ...

bartoszmajsak:
Hey @smiklosovic if you can draft your idea let's say on [discuss.arquillian.org](http://discuss.arquillian.org) maybe we can all work on it together with @tbarath? Some sort of SPI/Cube extension point where you can simply provide your own logic for the verification might be a starting point.

@lordofthejars what could we inject to such an extension so that we can easily communicate with the container?

smiklosovic:
@bartoszmajsak I think that my approach is completely incompatible with the current source. I have extracted whole Cube runtime so I can use it in a standalone-like applications without Arquillian / JUnit runner at all so I can execute it is Java / Groovy / Scala ... I am just using core Arquillian Manager in non-test context which is able to fire events and mere presence of Cube on classpath is enough to have them started. My code look totally different and it would be basically invented from scratch.

lordofthejars:
@bartoszmajsak it should not be really difficult. Basically nowadays we add an string in the field `strategy` and this string is matched in an internal `switch`.  So what I would do is do a default catch which basically treat this element as a fully classified class with default constructor. Then we instantiate this class and execute a method provided by an interface.

So everyone can implement their own hacks.

smiklosovic:
:+1:

tbarath:
Hi @smiklosovic , you are right that this solution is a little bit fragile and can be cumbersome in some use cases. Still I think it has its place in Cube's Swiss Army knife. It's true you have to find the appropriate patter for every given service/container and it can change between versions but at least it is easy to configure. It doesn't require coding so it is a no-brainer to use for the average users. Although your solution is more robust, it needs much more effort to implement for different services.

Anyway I agree it would be useful to make await strategies pluggable. It would have been my next suggestion if this PR hadn't been accepted. :wink:

Issue #338, opened 2016-04-19
Implement await strategies pluggable


lordofthejars:
It should be interesting to be able to provide to developers a way to implement their own await strategies.
Basically allowing to set a fully qualified class which follows an interface.

With this pluggable system, Cube will inject the current Cube (or binding) object that is going to be deployed as well as the docker host.

PR #340, opened 2016-04-20
implements issue #338 by adding the possibility for custom await strategies

lordofthejars:
implements issue #338 by adding the possibility to register custom await strategies

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -383,6 +383,7 @@ polling:: in this case a polling (with _ping_ or _ss_ command) is executed for 5
 static:: similar to _polling_ but it uses the host ip and specified list of ports provided as configuration parameter. This can be used in case of using _Boot2Docker_.
 sleeping:: sleeps current thread for the specified amount of time. You can specify the time in seconds or milliseconds.
 log:: it looking for a specified pattern in container log to detect service startup. This can be used when there is no port to connect or connecting to the port successfully doesn't mean the service is fully initialized.
+<fullyqualifiedclassname>:: if you specify a fully qualified class name, Arquillian Cube will instantiate the given class. In this way you can implement your own await strategies. There are two rules to follow, the first one is that class must implement `AwaitStrategy` and the second one is that one default constructor must be provided. Optionally you can add setters for types `Cube`, `DockerClientExecutor` or `Await` to inject them into the await strategy.
 
 By default in case you don't specify any _await_ strategy, polling with _ss_ command is used.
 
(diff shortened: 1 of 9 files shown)


README.adoc
```diff
@@ -454,6 +455,48 @@ tomcat:
...
+[source, java]
+.org.arquillian.cube.docker.impl.await.CustomAwaitStrategyImpl
+----
+public class CustomAwaitStrategyImpl implements AwaitStrategy {
+
+  Await params;
+  DockerClientExecutor dockerClientExecutor;
+  Cube<?> cube;
+
+  public void setCube(Cube<?> cube) {
```
bartoszmajsak: Is setter "injection" the only way we ca provide it? How about constructor injection instead?

lordofthejars: I thought the same butI don't like to provide two ways of doing the same thing, or for example you need to check the permutations of all of them since not all the parameters might be mandatory. For example:

```
public Custom(Cube, DockerClient, Await)

or

public Custom(Cube, DockerClient)

or 

public Custom( DockerClient, Await)

and so on
```

bartoszmajsak: Then what about fields?

bartoszmajsak: Anyway, I think that with `@Inject` you can only have one constructor which will be used. Hence no need for any variants check.

aslakknutsen: Why don't we make the SPI await(Cube<?> cube, DockerClientExecutor dockerClientExecutor, Await params) instead?

bartoszmajsak: I don't really know the code base that well, hence it might be an irrelevant question.

Why do we have both `Cube` and `DockerClientExecutor`. Can't we expose clients from the `Cube` itself?

lordofthejars: @aslakknutsen I am not creating the await with all parameters because not all await strategy requires all of them, if you look at current implementations you will see that some of them requires one, two or three parameters, so instead of creating an interface with a method with three parameters, I think it is better to maintain it clean and let the users choose what they want to inject.

lordofthejars: @bartoszmajsak Basically I need to pass both parameters because with `Cube` you can get some information about bindings. But I also pass the DockerClientExecutor because there is where all operations to Docker host are defined (even you can get `DockerClient` which is the base class from docker-java ) the reason is that in every release Docker is adding more and more operations so a user might need one and other user another one, if we make all operations through `Cube` means that we will need to keep update, and possibly some users will open issues because we add more operations on that class. If we find that some operations are repeated a lot, we can add it in `Cube` but at least we left an open door for a workaround (by using `DockerClient`)

lordofthejars: @bartoszmajsak The problem of only having one constructor (`@Inject`) is that then I am sure that there will be users that will add two constructors and will complain. With default constructor + setters seems a better approach for avoiding confusions. 

I agree with you that field injection should be supported too.

aslakknutsen: @lordofthejars I know. But on an SPI level you don't have to care. This is what we provide you to make a decision. What you actually use is up to the impl.

lordofthejars:
@aslakknutsen ok I would like to maintain the method without parameters but let's create this SPI in that way because it seems you and Bartosz feels more comfortable and because it has the advantage of not having to deal with reflection for injecting fields.

lordofthejars:
@aslakknutsen I remember now why we cannot use this approach. Because `AwaitStrategy` is on SPI module, but not `Await` or `DockerClientExecutor`.

lordofthejars:
We merge this?

PR #342, opened 2016-04-26
Add a config option "ignoreCleanup" to keep resources when exiting

jwendell:
Sometimes, mainly when debugging, it's useful to keep pods around
in order to take a closer look at them, their logs, their state, etc.

### DIFF
--- a/openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/CubeOpenShiftConfiguration.java
+++ b/openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/CubeOpenShiftConfiguration.java
@@ -20,6 +20,7 @@ public class CubeOpenShiftConfiguration {
     private static final String DEFINITIONS = "definitions";
     private static final String AUTO_START_CONTAINERS = "autoStartContainers";
     private static final String PROXIED_COTNAINER_PORTS = "proxiedContainerPorts";
+    private static final String IGNORE_CLEANUP = "ignoreCleanup";
 
     private String originServer;
     private String namespace;
@@ -28,6 +29,7 @@ public class CubeOpenShiftConfiguration {
     private String definitionsFile;
     private String[] autoStartContainers;
     private Set<String> proxiedContainerPorts;
+    private boolean ignoreCleanup = false;
 
     public String getOriginServer() {
         return originServer;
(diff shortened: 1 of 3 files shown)


bartoszmajsak:
Thanks for your contribution! @lordofthejars Should we extend tests as part of this PR as well?

lordofthejars:
Thank you so much for the contribution. @bartoszmajsak I think that if in this case there is no test it is not a real disaster since it is the typical test that you mock the client and you only check if Boolean.parseBoolean works as expected

aslakknutsen:
Isn't this the same as ConnectionMode? ConnectionMode is Cube SPI, not Docker specific. World probably be better if they both use the same configuration..

lordofthejars:
Ok, true I have reverted the changes. So basically what it should be done is using ConnectionMode attribute. @jwendell you can read about what @aslakknutsen says in https://github.com/arquillian/arquillian-cube#allow-connecting-to-a-running-container

bartoszmajsak:
But is it really the same? I assume @jwendell would like Cube to take care of the startup of the pods but then either leave them running or export relevant content for further inspection. `STARTORCONNECT` is not really giving this ability if I understand it correctly.

aslakknutsen:
@bartoszmajsak STARTORCONNECTANDLEAVE :)

bartoszmajsak:
didnt-read

jwendell:
Hi guys.

I put this snippet into arquillian.xml file:
    <extension qualifier="cube">
        <property name="connectionMode">STARTORCONNECTANDLEAVE</property>
    </extension>

After that I'm getting an error:
Caused by: io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: POST at: https://SERVER.COM:8443/api/v1/namespaces/cearq-eap64-udbdqlfl194/pods. Message: resourceVersion may not be set on objects to be created. Received status: Status(apiVersion=v1, code=500, details=null, kind=Status, message=resourceVersion may not be set on objects to be created, metadata=ListMeta(resourceVersion=null, selfLink=null, additionalProperties={}), reason=null, status=Failure, additionalProperties={}).

Full stack trace: https://gist.github.com/jwendell/d03cc9cebbd1d93273dad94e7ac56cbe

Am I missing something?

aslakknutsen:
Hmm, no clue on top of my head. I would try to debug the PodCube to see
what's going on..

jwendell:
After fixing the error above, test went successfully but it **did delete** the pod in the end. Is there other setting other than this STARTORCONNECTANDLEAVE flag I should pay attention?

aslakknutsen:
That should be it. I forget where the logic is handled in Cube atm, but possible the OpenShift life cycle is a little bit different and there need some changes in PodCube

jwendell:
Ok, just created issue #346 so that we can discuss this there.

Issue #190, opened 2015-11-04
Explore new network mode in Docker 1.9


lordofthejars:
http://blog.docker.com/2015/11/docker-multi-host-networking-ga/

PR #349, opened 2016-05-05
Adding network support fixing issue #190

lordofthejars:
Adding network support fixing issue #190.

This is an initial implementation that could evolve in following issues.


lordofthejars:
oh shit everything has been screwed

lordofthejars:
It seems that everything has been broken I don't know how. Anyway I have reviewed the PR in GithubDesktop and everything looks fine, even the history is ok but probably since I have done a rebase, some commits must be changed and then it detects as different commit. So what I can do is first you can review https://github.com/arquillian/arquillian-cube/pull/349/commits/d9c4e15640ec4f76dedf5ae2d3caaa93e45e2a32 which is the important one the other were already reviewed.

And then do a git push -f of my master to this master, I know it is dirty but I don't know now how to make everything back to normal

PR #353, opened 2016-05-07
Adds SPI for adding custom autostart order and custom parser

lordofthejars:


spi/src/main/java/org/arquillian/cube/spi/Node.java
```diff
@@ -0,0 +1,102 @@
+package org.arquillian.cube.spi;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class Node {
```
aslakknutsen: This could use some JavaDoc

spi/src/main/java/org/arquillian/cube/spi/AutoStartParser.java
```diff
@@ -0,0 +1,8 @@
+package org.arquillian.cube.spi;
+
+import java.util.Map;
+
+public interface AutoStartParser {
```
aslakknutsen: This could use some JavaDoc

spi/src/main/java/org/arquillian/cube/spi/AutoStartOrder.java
```diff
@@ -0,0 +1,10 @@
+package org.arquillian.cube.spi;
+
+import java.util.List;
+
+public interface AutoStartOrder<T> {
```
aslakknutsen: This could use some JavaDoc

PR #355, opened 2016-05-09
Add configuration option to remove volumes

tbarath:
Add configuration option to remove volumes associated to containers created by Cube when the containers are removed.

### DIFF
--- a/docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/DockerClientExecutor.java
+++ b/docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/DockerClientExecutor.java
@@ -25,8 +25,6 @@ import java.util.regex.Pattern;
 
 import javax.ws.rs.ProcessingException;
 
-import com.github.dockerjava.api.command.CreateNetworkCmd;
-import com.github.dockerjava.api.command.CreateNetworkResponse;
 import org.apache.http.conn.UnsupportedSchemeException;
 import org.arquillian.cube.TopContainer;
 import org.arquillian.cube.docker.impl.client.CubeDockerConfiguration;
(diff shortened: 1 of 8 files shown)


lordofthejars:
Thank you very much for this PR, I like the idea to maintain the environment clean. Only one comment. At this time if you configure to remove volumes, you remove volumes of all containers, but should it has sense to configure this at lower level so you can choose which container you want to remove the volumes and which not?

If you think that removing volumes would have sense to have it not globally but selectively to specific containers, then maybe we should implement as BeforeStopEvent: https://github.com/arquillian/arquillian-cube#before-stop-events

tbarath:
Hi @lordofthejars , I agree with you. It would be useful to configure this on container level. That crossed my mind too. Still I think the global configuration option should be kept. If you have more container and want to configure them in the same way, the global option is more convenient (DRY). What do you think?

About implementing container level configuration. You can delete the volume(s) only after the container is removed so BeforeStop event is not feasible. Implement this would be possible as AfterDestroy event but then we should call `removeVolumeCmd`. Not surprisingly this command requires volume names that are not available after the container is removed. We could save them before container remove but it's more work than required I think. Instead I propose to add a top level option to container configuration:

```yaml
tomcat_default:
  image: tutum/tomcat:7.0
  removeVolumes: true
```
This could be passed as second parameter to method call [here] (https://github.com/arquillian/arquillian-cube/blob/master/docker/docker/src/main/java/org/arquillian/cube/docker/impl/model/DockerCube.java#L153). If not specified the global option is used that is false by default. (I'm inclined to set this true but maybe it is a little bit harsh from a backward compatibility point of view. Although we are in Alpha phase so there is no such promise. :wink:)

Is this solution acceptable for you? If so, I'll update the PR soon.

lordofthejars:
Yes it has sense what you mention. But then this option should be also added as a custom property `cubeSpecificProperties ` so it can be used when users uses Docker-compose format. https://github.com/arquillian/arquillian-cube#cubespecificproperties

So +1 to global and +1 to what you mention here.

tbarath:
@lordofthejars I've updated the PR as we discussed.

docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/DockerClientExecutor.java
```diff
@@ -465,8 +465,8 @@ public void stopContainer(String containerId) {
         this.dockerClient.stopContainerCmd(containerId).exec();
     }
 
-    public void removeContainer(String containerId) {
-        this.dockerClient.removeContainerCmd(containerId).exec();
+    public void removeContainer(String containerId, boolean removeVolumes) {
```
lordofthejars: One thing that I have thought today. Maybe it would be better to pass CubeContainer object like it happens with create. In this way if we ever need to do any other tasks when removing the container, we don't need to expand the method. Maybe for now we can leave as is and in the future if this happens refactor with CubeContainer class. @aslakknutsen what do you think?

PR #360, opened 2016-05-19
adds depends_on support

lordofthejars:


docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/config/CubeContainer.java
```diff
@@ -406,6 +411,31 @@ public void setContainerName(String containerName) {
...
+    public void setDependsOn(Collection<String> dependsOn) {
+        this.dependsOn = dependsOn;
+    }
+
+    public Collection<String> getDependingContainers() {
+        // Depends on has more priority than links
+        if (dependsOn != null && dependsOn.size() > 0) {
+            return Collections.unmodifiableCollection(dependsOn);
+        } else {
+            if (links != null && links.size() > 0) {
```
bartoszmajsak: How about extracting this whole `else` block to a dedicated method with a name explaining what it's supposed to do?

bartoszmajsak: How about extracting this whole `else` block to a dedicated method with a name explaining what it's supposed to do?

docker/docker/src/test/java/org/arquillian/cube/docker/impl/util/AutoStartOrderUtilTestCase.java
```diff
@@ -49,6 +49,28 @@
...
+            "A:\n" +
+            "  dependsOn:\n" +
+            "    - B\n" +
+            "B:\n" +
+            "  image: a\n" +
+            "C:\n" +
+            "  links:\n" +
+            "    - A:A\n" +
+            "  dependsOn:\n" +
+            "    - B\n";
```
bartoszmajsak: Wouldn't it be better for editing to have those as files instead?

lordofthejars: Probably yes, but we started with this approach and since they are more or less simple we continue with this approach. This approach is used un other places as well. Maybe we can open a technical deb issue.

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/config/CubeContainer.java
lordofthejars: ok

docker/docker/src/test/java/org/arquillian/cube/docker/impl/util/AutoStartOrderUtilTestCase.java
bartoszmajsak: C'mon mate :) It's 5 minutes boyscoutting :)

lordofthejars: Well we need to find all the places and change it since it is not only this class. Basically to have something common in all places and do not have both approaches.

Issue #175, opened 2015-08-25
Split README


aslakknutsen:
As we add more and more 'sub modules' to Cube, I think we should split the /README into subsection with links.

e.g. 

Keep /README about the common stuff, e.g. CubeController, Containerless
Move Docker / Docker-Machine / Docker-Compose into /docker/README
'Move' OpenShift into /openshift/README

wdyt?

lordofthejars:
I think it has a lot of sense since the project is growing a lot. Moreover
it will help to organize the chapter of AIA
El dc., 26 d’ag., 2015 a les 1.14 Aslak Knutsen <notifications@github.com>
va escriure:

PR #363, opened 2016-06-10
Polish README

eddumelendez:
See gh-175


lordofthejars:
Thank you very much for this PR. This is something we should do it long time ago.

PR #398, opened 2016-07-18
adds structure for integrating fabric8 kubernetes support

lordofthejars:
This change is [Reviewable](https://reviewable.io/reviews/arquillian/arquillian-cube/398)

### DIFF
--- a/kubernetes/fabric8/pom.xml
+++ b/kubernetes/fabric8/pom.xml
@@ -1,0 +1,14 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<parent>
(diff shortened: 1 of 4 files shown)


aslakknutsen:
I wouldn't call it fabric8. Fabric8 is a very loose term and not really related to what this is.

lordofthejars:
well we can change it if you want but my idea is that inside what now is fabric8 there are some elements that are fabric8 specific, so the idea is move there the project and start moving the parts in their own places. And leave in fabric8 module the parts specific to fabric8. Of course we can do it creating first the final name and create the fabric8 thing at the end. What do you have in mind?

aslakknutsen:
If it is Fabric8 specific then sure. I thought the intent here was the 'Fabric8' low-level Kubernetes assertions. Maybe I spoke too soon. :)

lordofthejars:
Now I think assertions will continue be in fabric8 project and they will be a dependency, but this is something that now we are going to discuss.

In fact there are only some classes specific to fabric8 the other ones are generic to kubernetes but as first start let's stick everything together and then start moving things

lordofthejars:
Ok maybe it is easier in the future to do it another way, create first the kubernetes module and extract to fabric8 parts

PR #401, opened 2016-07-22
400

iocanel:
This change is [Reviewable](https://reviewable.io/reviews/arquillian/arquillian-cube/401)

### DIFF
--- a/containerless/pom.xml
+++ b/containerless/pom.xml
@@ -31,6 +31,11 @@
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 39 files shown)


lordofthejars:
Hi,
Thank you very much for this awesome work, it looks pretty good
I am going to merge it right now so you can continue on work, only one thing when you can (it is not urgent), could you add some lines in README.adoc on how to use it? Just because users of Cube can take advantage of your work as well :)

iocanel:
Of course! I will add something as soon as possible.

Issue #385, opened 2016-07-02
Integration with Drone


lordofthejars:
Integration with Drone. This implies two tasks:

- [x] Add automatically the docker image depending on the browser.
- [x] Replace existing `WebDriver` instance (if any) to a WebDriver that is of type Remote and points to given `dockerhost` value.

PR #402, opened 2016-07-27
Adds cube-drone integration. Issue #385

lordofthejars:
This change is [Reviewable](https://reviewable.io/reviews/arquillian/arquillian-cube/402)

### DIFF
--- a/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeSuiteLifecycleController.java
+++ b/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeSuiteLifecycleController.java
@@ -7,7 +7,6 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.Future;
 
 import org.arquillian.cube.docker.impl.docker.DockerClientExecutor;
-import org.arquillian.cube.docker.impl.util.AutoStartOrderUtil;
 import org.arquillian.cube.spi.ConnectionMode;
 import org.arquillian.cube.spi.CubeConfiguration;
 import org.arquillian.cube.spi.event.CreateCube;
@@ -20,7 +19,6 @@ import org.arquillian.cube.spi.event.lifecycle.AfterAutoStart;
 import org.arquillian.cube.spi.event.lifecycle.AfterAutoStop;
 import org.arquillian.cube.spi.event.lifecycle.BeforeAutoStart;
 import org.arquillian.cube.spi.event.lifecycle.BeforeAutoStop;
-import org.arquillian.cube.spi.event.lifecycle.CubeLifecyleEvent;
 import org.jboss.arquillian.core.api.Event;
 import org.jboss.arquillian.core.api.Instance;
 import org.jboss.arquillian.core.api.annotation.Inject;
@@ -32,16 +30,16 @@ import org.jboss.arquillian.test.spi.event.suite.BeforeSuite;
 public class CubeSuiteLifecycleController {
 
     @Inject
-    private Event<AfterAutoStart> afterAutoStart;
+    private Event<AfterAutoStart> afterAutoStartEvent;
 
     @Inject
-    private Event<BeforeAutoStart> beforeAutoStart;
+    private Event<BeforeAutoStart> beforeAutoStartEvent;
 
     @Inject
-    private Event<BeforeAutoStop> beforeAutoStop;
+    private Event<BeforeAutoStop> beforeAutoStopEvent;
 
     @Inject
-    private Event<AfterAutoStop> afterAutoStop;
+    private Event<AfterAutoStop> afterAutoStopEvent;
 
     @Inject
     private Event<CubeControlEvent> controlEvent;
@@ -53,19 +51,19 @@ public class CubeSuiteLifecycleController {
     private Instance<DockerClientExecutor> dockerClientExecutor;
 
     public void startAutoContainers(@Observes(precedence = 100) BeforeSuite event, CubeConfiguration cubeConfiguration, CubeDockerConfiguration dockerConfiguration) {
-        beforeAutoStart.fire(new BeforeAutoStart());
+        beforeAutoStartEvent.fire(new BeforeAutoStart());
         final DockerAutoStartOrder dockerAutoStartOrder = dockerConfiguration.getDockerAutoStartOrder();
         List<String[]> autoStartSteps = dockerAutoStartOrder.getAutoStartOrder(dockerConfiguration);
         startAllSteps(autoStartSteps, cubeConfiguration.getConnectionMode());
-        afterAutoStart.fire(new AfterAutoStart());
+        afterAutoStartEvent.fire(new AfterAutoStart());
     }
 
     public void stopAutoContainers(@Observes(precedence = -100) AfterSuite event, CubeDockerConfiguration configuration) {
-        beforeAutoStop.fire(new BeforeAutoStop());
+        beforeAutoStopEvent.fire(new BeforeAutoStop());
         final DockerAutoStartOrder dockerAutoStartOrder = configuration.getDockerAutoStartOrder();
         List<String[]> autoStopSteps = dockerAutoStartOrder.getAutoStopOrder(configuration);
         stopAllSteps(autoStopSteps);
-        afterAutoStop.fire(new AfterAutoStop());
+        afterAutoStopEvent.fire(new AfterAutoStop());
     }
 
     private void startAllSteps(List<String[]> autoStartSteps, ConnectionMode connectionMode) {
--- a/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeSuiteLifecycleController.java
+++ b/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeSuiteLifecycleController.java
@@ -18,6 +18,8 @@ import org.arquillian.cube.spi.event.StartCube;
 import org.arquillian.cube.spi.event.StopCube;
 import org.arquillian.cube.spi.event.lifecycle.AfterAutoStart;
 import org.arquillian.cube.spi.event.lifecycle.AfterAutoStop;
+import org.arquillian.cube.spi.event.lifecycle.BeforeAutoStart;
+import org.arquillian.cube.spi.event.lifecycle.BeforeAutoStop;
 import org.arquillian.cube.spi.event.lifecycle.CubeLifecyleEvent;
 import org.jboss.arquillian.core.api.Event;
 import org.jboss.arquillian.core.api.Instance;
@@ -33,6 +35,12 @@ public class CubeSuiteLifecycleController {
     private Event<AfterAutoStart> afterAutoStart;
 
     @Inject
+    private Event<BeforeAutoStart> beforeAutoStart;
+
+    @Inject
+    private Event<BeforeAutoStop> beforeAutoStop;
+
+    @Inject
     private Event<AfterAutoStop> afterAutoStop;
 
     @Inject
@@ -45,6 +53,7 @@ public class CubeSuiteLifecycleController {
     private Instance<DockerClientExecutor> dockerClientExecutor;
 
     public void startAutoContainers(@Observes(precedence = 100) BeforeSuite event, CubeConfiguration cubeConfiguration, CubeDockerConfiguration dockerConfiguration) {
+        beforeAutoStart.fire(new BeforeAutoStart());
         final DockerAutoStartOrder dockerAutoStartOrder = dockerConfiguration.getDockerAutoStartOrder();
         List<String[]> autoStartSteps = dockerAutoStartOrder.getAutoStartOrder(dockerConfiguration);
         startAllSteps(autoStartSteps, cubeConfiguration.getConnectionMode());
@@ -52,6 +61,7 @@ public class CubeSuiteLifecycleController {
     }
 
     public void stopAutoContainers(@Observes(precedence = -100) AfterSuite event, CubeDockerConfiguration configuration) {
+        beforeAutoStop.fire(new BeforeAutoStop());
         final DockerAutoStartOrder dockerAutoStartOrder = configuration.getDockerAutoStartOrder();
         List<String[]> autoStopSteps = dockerAutoStartOrder.getAutoStopOrder(configuration);
         stopAllSteps(autoStopSteps);
--- a/docker/drone/src/main/java/org/arquillian/cube/docker/drone/CubeDroneConfiguration.java
+++ b/docker/drone/src/main/java/org/arquillian/cube/docker/drone/CubeDroneConfiguration.java
@@ -1,0 +1,57 @@
+package org.arquillian.cube.docker.drone;
+
+import java.util.Map;
+
+/**
+ * Configuration for cube drone extension.
+ */
+public class CubeDroneConfiguration {
+
+    /**
+     * Final directory where recordings are moved after execution.
+     */
+    private String finalDirectory;
+    /**
+     * Record mode to decide if not recording, only when failure or always.
+     */
+    private RecordMode recordMode = RecordMode.ALL;
+
+    public boolean isRecordingOnlyFailing() {
+        return recordMode == RecordMode.ONLY_FAILING;
+    }
+
+    public boolean isRecording() {
+        return recordMode != RecordMode.NONE;
+    }
+
+    public RecordMode getRecordMode() {
+        return recordMode;
+    }
+
+    public boolean isVideoOutputDirectorySet() {
+        return finalDirectory != null;
+    }
+
+    public String getFinalDirectory() {
+        return finalDirectory;
+    }
+
+    public static enum RecordMode {
+        ALL, ONLY_FAILING, NONE;
+    }
+
+    public static CubeDroneConfiguration fromMap(Map<String, String> config) {
+        CubeDroneConfiguration cubeDroneConfiguration = new CubeDroneConfiguration();
+
+        if (config.containsKey("recordingMode")) {
+            cubeDroneConfiguration.recordMode = RecordMode.valueOf(config.get("recordingMode"));
+        }
+
+        if (config.containsKey("videoOutput")) {
+            cubeDroneConfiguration.finalDirectory = config.get("videoOutput");
+        }
+
+        return cubeDroneConfiguration;
+    }
+
+}
--- a/docker/drone/src/main/java/org/arquillian/cube/docker/drone/CubeDroneConfiguration.java
+++ b/docker/drone/src/main/java/org/arquillian/cube/docker/drone/CubeDroneConfiguration.java
@@ -16,7 +16,7 @@ public class CubeDroneConfiguration {
      */
     private RecordMode recordMode = RecordMode.ALL;
 
-    public boolean isRecordingOnlyFailing() {
+    public boolean isRecordOnFailure() {
         return recordMode == RecordMode.ONLY_FAILING;
     }
 
--- a/docker/drone/src/main/java/org/arquillian/cube/docker/drone/DroneOverwriteConfigurator.java
+++ b/docker/drone/src/main/java/org/arquillian/cube/docker/drone/DroneOverwriteConfigurator.java
@@ -1,0 +1,30 @@
+package org.arquillian.cube.docker.drone;
+
+import org.jboss.arquillian.config.descriptor.api.ArquillianDescriptor;
+import org.jboss.arquillian.config.descriptor.api.ExtensionDef;
+import org.jboss.arquillian.core.api.annotation.Observes;
+
+import java.util.List;
+import java.util.Map;
+
+public class DroneOverwriteConfigurator {
+
+    public void overwriteDroneConfiguration(@Observes ArquillianDescriptor arquillianDescriptor) {
+
+        final List<ExtensionDef> extensions = arquillianDescriptor.getExtensions();
+
+        final ExtensionDef webdriver = arquillianDescriptor.extension("webdriver");
+
+        if (webdriver != null) {
+
+            webdriver.property("remote", "true");
+
+            final Map<String, String> extensionProperties = webdriver.getExtensionProperties();
+            if (!extensionProperties.containsKey("browser")) {
+                webdriver.property("browser", "firefox");
+            }
+
+        }
+    }
+
+}
--- a/docker/drone/src/main/java/org/arquillian/cube/docker/drone/SeleniumContainers.java
+++ b/docker/drone/src/main/java/org/arquillian/cube/docker/drone/SeleniumContainers.java
@@ -1,16 +1,13 @@
 package org.arquillian.cube.docker.drone;
 
-import org.arquillian.cube.docker.drone.util.SeleniumVersionUtil;
-import org.arquillian.cube.docker.drone.util.VolumeUtils;
+import org.arquillian.cube.docker.drone.util.SeleniumVersionExtractor;
+import org.arquillian.cube.docker.drone.util.VolumeCreator;
 import org.arquillian.cube.docker.impl.client.config.Await;
 import org.arquillian.cube.docker.impl.client.config.CubeContainer;
 import org.arquillian.cube.docker.impl.client.config.Image;
 import org.arquillian.cube.docker.impl.client.config.Link;
 import org.arquillian.cube.docker.impl.client.config.PortBinding;
 
-import java.io.IOException;
-import java.lang.reflect.Array;
-import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Arrays;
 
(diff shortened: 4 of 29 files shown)


docs/drone.adoc
```diff
@@ -0,0 +1,110 @@
+== Cube Docker Drone Integration
+
+=== Drone
+
+The Arquillian Drone extension for Arquillian provides a simple way how to include functional tests for your application with a web-based user interface.
+Arquillian Drone brings the power of WebDriver into the Arquillian framework.
```
bartoszmajsak: I would rephrase this sentence. Maybe something like

> provides a simple way of writing functional tests for your ....

Maybe some additional bits about what Drone simplifies?

docs/drone.adoc
```diff
@@ -0,0 +1,110 @@
...
+
+You can read more about Drone in https://docs.jboss.org/author/display/ARQ/Drone
```
bartoszmajsak: in - at

docs/drone.adoc
```diff
@@ -0,0 +1,110 @@
...
+
+=== Integration with Cube Docker
+
+Arquillian Cube Docker is the Arquillian extension for managing Docker containers from your test.
+
+Selenium offers Docker images for Selenium Standalone Server with Chrome or Firefox installed.
+You can see the images at https://github.com/SeleniumHQ/docker-selenium.
```
bartoszmajsak: how about: You can check available images

docs/drone.adoc
```diff
@@ -0,0 +1,110 @@
...
+
+An example of configuration might look like:
+
+[source, xml]
+.arquillian.xml
+----
+<arquillian>
+    <extension qualifier="cubedrone">
+        <property name="recordingMode">NONE</property>
+</arquillian>
```
bartoszmajsak: XML non-compliant :) Missing `</extension>`

docs/drone.adoc
```diff
@@ -0,0 +1,110 @@
...
+----
+
+Configure the Cube extension:
+
+[source, xml]
+.arquillian.xml
+----
+<extension qualifier="docker">
+    <property name="definitionFormat">COMPOSE</property>
+    <property name="machineName">dev</property>
```
bartoszmajsak: Maybe highlight it's optional? Docker daemon shoud be fine too?

docs/drone.adoc
```diff
@@ -0,0 +1,110 @@
...
+            <groupId>org.jboss.arquillian.extension</groupId>
+            <artifactId>arquillian-drone-bom</artifactId>
+            <version>2.0.0.Final</versio>
+            <type>pom</type>
+            <scope>import</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.arquillian.selenium</groupId>
+            <artifactId>selenium-bom</artifactId>
+            <version>2.53.1</version>
```
bartoszmajsak: Do we need it? Doesn't `arquillian-drone-bom` bring it's own selenium versions? Are they behind?

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeSuiteLifecycleController.java
```diff
@@ -45,13 +53,15 @@
     private Instance<DockerClientExecutor> dockerClientExecutor;
 
     public void startAutoContainers(@Observes(precedence = 100) BeforeSuite event, CubeConfiguration cubeConfiguration, CubeDockerConfiguration dockerConfiguration) {
+        beforeAutoStart.fire(new BeforeAutoStart());
```
bartoszmajsak: I would rename all the `Event<...>` to have an `Event` suffix.

I think it's more readable when you see

```
beforeAutoStartEvent.fire(new BeforeAutoStart());
```

WDYT?

docker/drone/src/main/java/org/arquillian/cube/docker/drone/CubeDroneConfiguration.java
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * Final directory where recordings are moved after execution.
+     */
+    private String finalDirectory;
+    /**
+     * Record mode to decide if not recording, only when failure or always.
+     */
+    private RecordMode recordMode = RecordMode.ALL;
+
+    public boolean isRecordingOnlyFailing() {
```
bartoszmajsak: Maybe `recordOnFailure` would be a bit more readable?

docker/drone/src/main/java/org/arquillian/cube/docker/drone/DroneOverwriteConfigurator.java
```diff
@@ -0,0 +1,30 @@
...
+
+        final ExtensionDef webdriver = arquillianDescriptor.extension("webdriver");
+
+        if (webdriver != null) {
+
+            webdriver.property("remote", "true");
+
+            final Map<String, String> extensionProperties = webdriver.getExtensionProperties();
+            if (!extensionProperties.containsKey("browser")) {
+                webdriver.property("browser", "firefox");
```
bartoszmajsak: Is it default behavior for Drone in general? It falls back to firefox when not specified?

bartoszmajsak: Additionally - what if I specify the browser for which we don't have a docker image yet?

docker/drone/src/main/java/org/arquillian/cube/docker/drone/SeleniumContainers.java
```diff
@@ -0,0 +1,146 @@
...
+import java.lang.reflect.Array;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Arrays;
+
+public class SeleniumContainers {
+
+    private static final String CHROME_IMAGE = "selenium/standalone-chrome-debug:%s";
+    private static final String FIREFOX_IMAGE = "selenium/standalone-firefox-debug:%s";
+    private static final String VNC_IMAGE = "richnorth/vnc-recorder:latest";
```
bartoszmajsak: We could make it extensible so that users can easily add their own containers for those browsers which are not support yet or overwrite existing ones for example.

docker/drone/src/main/java/org/arquillian/cube/docker/drone/SeleniumContainers.java
```diff
@@ -0,0 +1,146 @@
...
+        cubeContainer.setImage(Image.valueOf(image));
+        cubeContainer.setPortBindings(
+                Arrays.asList(PortBinding.valueOf(SELENIUM_BOUNDED_PORT + "-> 4444"))
+        );
+
+        Await await = new Await();
+        await.setStrategy("http");
+        // Doing an http request to selenium node returns a 403 if Jetty is up and running
+        await.setResponseCode(403);
+        await.setUrl("http://dockerHost:" + SELENIUM_BOUNDED_PORT);
```
bartoszmajsak: Shouldn't `"http://dockerHost:"` be configurable?

docker/drone/src/main/java/org/arquillian/cube/docker/drone/VncRecorderLifecycleManager.java
```diff
@@ -0,0 +1,124 @@
...
+            } else {
+                if (GRADLE_REPORT_DIR.exists()) {
+                    return GRADLE_REPORT_DIR.toPath();
+                } else {
+                    if (DEFAULT_LOCATION_OUTPUT_GRADLE.exists()) {
+                        return DEFAULT_LOCATION_OUTPUT_GRADLE.toPath();
+                    } else {
+                        if (DEFAULT_LOCATION_OUTPUT_MAVEN.exists()) {
+                            return DEFAULT_LOCATION_OUTPUT_MAVEN.toPath();
+                        }
```
bartoszmajsak: Can change make this waterfall to somewhat more elegant solution?

docker/drone/src/main/java/org/arquillian/cube/docker/drone/util/VolumeUtils.java
```diff
@@ -0,0 +1,52 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+/**
+ * Volume utils.
+ */
+public class VolumeUtils {
```
bartoszmajsak: I'm not a big fan of `Utils` classes. Maybe `VolumeCreator` would be a better fit here?

docker/drone/src/main/java/org/arquillian/cube/docker/drone/util/SeleniumVersionUtil.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Set;
+import java.util.jar.Attributes;
+import java.util.jar.Manifest;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Utility class for Selenium.
+ */
+public class SeleniumVersionUtil {
```
bartoszmajsak: `SeleniumVersionExtractor`
and then `.fromClassPath` ?

docker/drone/src/main/java/org/arquillian/cube/docker/drone/VncRecorderLifecycleManager.java
```diff
@@ -0,0 +1,124 @@
...
+ */
+public class VncRecorderLifecycleManager {
+
+    private static final File DEFAULT_LOCATION_OUTPUT_MAVEN = new File("target");
+    private static final File DEFAULT_LOCATION_OUTPUT_GRADLE = new File("build");
+
+    private static final File MAVEN_REPORT_DIR = new File(DEFAULT_LOCATION_OUTPUT_MAVEN, "surefire-reports");
+    private static final File GRADLE_REPORT_DIR = new File(DEFAULT_LOCATION_OUTPUT_GRADLE, "test-results");
+
+    Cube vnc;
```
bartoszmajsak: Doesn't have to be package protected?

docker/drone/src/main/java/org/jboss/arquillian/drone/webdriver/factory/DockerRemoteWebDriverFactory.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @Inject
+    Instance<SeleniumContainers> seleniumContainersInstance;
+
+    @Override
+    public void destroyInstance(RemoteWebDriver remoteWebDriver) {
+        try {
+            remoteWebDriver.quit();
+        } catch(WebDriverException e) {
+            log.log(Level.WARNING, "@Drone {0} has been already destroyed and can't be destroyed again.", remoteWebDriver.getClass()
```
bartoszmajsak: Why do we `log.log` at some places where in other ones we simply `System.out.println` instead?

docker/drone/src/test/java/org/arquillian/cube/docker/drone/VncRecorderLifecycleManagerTest.java
```diff
@@ -0,0 +1,165 @@
...
+    Cube cube;
+
+    @Mock
+    After after;
+
+    @Mock
+    SeleniumContainers seleniumContainers;
+
+    @Test
+    public void shouldStartVncBeforeEachTestByDefault() {
```
bartoszmajsak: I don't see a correlation between "each test" and "starting vnc" - here we only test if it starts vnc. Where do we verify if that happens for every test? Some assumption I don't see here?

docker/drone/src/test/java/org/arquillian/cube/docker/drone/VncRecorderLifecycleManagerTest.java
```diff
@@ -0,0 +1,165 @@
...
+
+        VncRecorderLifecycleManager vncRecorderLifecycleManager = new VncRecorderLifecycleManager();
+        vncRecorderLifecycleManager.vnc = cube;
+        vncRecorderLifecycleManager.stopRecording(after,
+                testResult,
+                CubeDroneConfiguration.fromMap(conf),
+                seleniumContainers
+                );
+
+        assertThat(new File(destination, "org_arquillian_cube_docker_drone_VncRecorderLifecycleManagerTest_shouldMoveRecordingVideo.flv").exists(), is(true));
```
bartoszmajsak: This assertion with `assertj` would look way more readable :)

docker/drone/src/test/java/org/arquillian/cube/docker/drone/VncRecorderLifecycleManagerTest.java
```diff
@@ -0,0 +1,165 @@
...
+                testResult,
+                CubeDroneConfiguration.fromMap(conf),
+                seleniumContainers
+        );
+
+        assertThat(new File(destination, "org_arquillian_cube_docker_drone_VncRecorderLifecycleManagerTest_shouldDiscardRecordingIfConfiguredInOnlyFailingAndPassedTest.flv").exists(), is(false));
+    }
+
+    @Test
+    public void shouldMoveRecordingIfConfiguredInOnlyFailingAndFailedTest() throws IOException, NoSuchMethodException {
```
bartoszmajsak: Somehow this 'should_move_recording_if_configured_in_only_failing_and_failed_test' reads better for my taste.

Shall we try to follow this convention?

In addition it will make it easier to find a test method in the stack trace :)

docker/ftest-drone/src/test/java/org/arquillian/cube/drone/TodoBrowserTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @HostPort(containerName = "helloworld", value = 80)
+    int port;
+
+    @Test
+    public void shouldShowHelloWorld() throws MalformedURLException, InterruptedException {
+        URL url = new URL("http", ip, port, "/");
+        webDriver.get(url.toString());
+        final String message = webDriver.findElement(By.tagName("h1")).getText();
+        assertThat(message, is("Hello world!"));
```
bartoszmajsak: I assume this will record the video. How about some verification end-to-end that there really is non-zero file created? :_

docker/drone/src/main/java/org/arquillian/cube/docker/drone/util/SeleniumVersionUtil.java
```diff
@@ -0,0 +1,77 @@
...
+                }
+
+            }
+
+        } catch (Exception e) {
+            logger.log(Level.WARNING, "Exception {0} occurred while resolving selenium version and latest image is going to be used.", e.getMessage());
+            return SELENIUM_VERSION;
+        }
+
+        if (versions.size() == 0) {
```
bartoszmajsak: `versions.isEmpty()` ?

docker/drone/src/main/java/org/arquillian/cube/docker/drone/DroneOverwriteConfigurator.java
aslakknutsen: Drone default is the htmlunit driver.

docker/drone/src/main/java/org/arquillian/cube/docker/drone/VncRecorderLifecycleManager.java
```diff
@@ -0,0 +1,124 @@
...
+
+            this.vnc = vncContainer;
+        }
+    }
+
+    public void stopRecording(@Observes After afterTestMethod, TestResult testResult, CubeDroneConfiguration cubeDroneConfiguration, SeleniumContainers seleniumContainers) {
+
+        if (this.vnc != null) {
+
+            if (cubeDroneConfiguration.isRecordingOnlyFailing() && testResult.getStatus() == TestResult.Status.FAILED) {
```
bartoszmajsak: how about extracting this condition to method named
`shouldRecordOnlyOnFailure`

docker/ftest-drone/src/test/java/org/arquillian/cube/drone/TodoBrowserTest.java
lordofthejars: Well the idea of ftest is provide an example on how a developer would write a test with cube drone. Adding this assertion is something that should be done in another test

docker/drone/src/test/java/org/arquillian/cube/docker/drone/VncRecorderLifecycleManagerTest.java
lordofthejars: Maybe I need to rename the test. This scenario should be in another test more focused on integration test rather than unit test.

docker/drone/src/main/java/org/jboss/arquillian/drone/webdriver/factory/DockerRemoteWebDriverFactory.java
lordofthejars: This is something that I have seen in several Arquillian projects. When it is something related to configuration we use sysout. For the other just log.

docker/drone/src/main/java/org/arquillian/cube/docker/drone/VncRecorderLifecycleManager.java
lordofthejars: It is ok to do it in this way, my idea is packaged because it facilitates the testing thing and not protected because its intend is not to be accessible by a possible extend class.

docker/drone/src/main/java/org/arquillian/cube/docker/drone/VncRecorderLifecycleManager.java
lordofthejars: Any idea that it is easy to follow and fast to execute? I thought on Chain Pattern but maybe it is an overkill for just 5 `ifs`without any logic rather than returning a string.

docker/drone/src/main/java/org/arquillian/cube/docker/drone/SeleniumContainers.java
lordofthejars: It is not necessary since Cube takes care of this at runtime.

docker/drone/src/main/java/org/arquillian/cube/docker/drone/SeleniumContainers.java
lordofthejars: For now only these two are provided. I would say to wait until someone requests it, since I am not sure that someone wants to use a non official browser docker image. I think that the common scenario is this one.

docker/drone/src/main/java/org/arquillian/cube/docker/drone/DroneOverwriteConfigurator.java
lordofthejars: If yo specify an image that does not exists it throws an exception. User has configured something wrong and the exception notifies about this.

Drone by default uses htmlUnit but since this has no sense in this integration I decided that in case that no `browser` property is set, `firefox` is used. This allow user to not having to configure anything from drone perspective. (See arquillian.xml from ftest)

docs/drone.adoc
lordofthejars: Yes it is requirement from Drone. Drone does not specify the selenium version to use. For this reason that bon was created

docs/drone.adoc
lordofthejars: I can do it but in this case I am assuming that reader has read the first "chapter" first, but I will change it.

docker/drone/src/main/java/org/arquillian/cube/docker/drone/VncRecorderLifecycleManager.java
bartoszmajsak: I'm not sure if we should adjust anything in the code to facilitate testing needs to be honest. We could always improve design to make it testable instead of doing such a thing or `@VisibleForTesting` on some weird constructors which is quite often. Just my .2$

docker/ftest-drone/src/test/java/org/arquillian/cube/drone/TodoBrowserTest.java
bartoszmajsak: Then we need some more tests :)

docs/drone.adoc
lordofthejars: Done

docker/drone/src/main/java/org/arquillian/cube/docker/drone/SeleniumContainers.java
lordofthejars: https://github.com/arquillian/arquillian-cube/issues/403

docker/drone/src/main/java/org/arquillian/cube/docker/drone/VncRecorderLifecycleManager.java
lordofthejars: Well this approach is something that it is extensively used. For example Adam Bien talks a lot  about it and I have used for long time and it works. Also this is for unit tests. I prefer this approach rather than for example adding a setter.

bartoszmajsak: Not a part of this PR anyway, but I don't really agree with it as it brings a risk of a poor design just to satisfy testing. If it's hard to wire things together then maybe it's time to re-think how we build our components :)

But let's take it offline.

docs/drone.adoc
MatousJobanek: Well, not exactly. Arquillian-drone-bom manages a selenium version as well: https://repo1.maven.org/maven2/org/jboss/arquillian/extension/arquillian-drone-bom/2.0.0.Final/arquillian-drone-bom-2.0.0.Final.pom
The version defined in this BOM is that one the Drone should be compatible with.
The reason why the selenium-bom was created is that the Selenium's release cadence is much higher then the Drone's one. So, it can happen that the newest version of some browser is not supported by selenium version defined in Drone, then using selenium-bom the developer can easily specify selenium version that he really needs.

lordofthejars: Yes true, I have not explained correctly, but checking all the docs it uses
this approach. I think it is the correct one to follow in our examples too,
isn't it?

El dt., 2 ag. 2016 a les 15:54, Matous Jobanek (<notifications@github.com>)
va escriure:

MatousJobanek: which documents and examples do you mean? Just to know so I can fix it or just to have an overview. :-)

docs/drone.adoc
```diff
@@ -0,0 +1,124 @@
...
+<dependencies>
+    <dependency>
+        <groupId>org.arquillian.cube</groupId>
+        <artifactId>arquillian-cube-docker-drone</artifactId>
+        <scope>test</scope>
+    </dependency>
+    <dependency>
+        <groupId>org.jboss.arquillian.extension</groupId>
+        <artifactId>arquillian-drone-webdriver-depchain</artifactId>
+        <version>2.0.0.Final</version>
```
MatousJobanek: If you use BOM then you don't have to specify version for the depchain - it is managed by the BOM

docs/drone.adoc
lordofthejars: http://arquillian.org/guides/functional_testing_using_graphene/

MatousJobanek: Ah, that one... I gonna fix it there. Thanks :-)

docker/drone/src/main/java/org/jboss/arquillian/drone/webdriver/factory/DockerRemoteWebDriverFactory.java
```diff
@@ -0,0 +1,83 @@
...
+        RemoteWebDriver remoteWebDriver = new RemoteWebDriver(createSeleniumUrl(), getDesiredCapabilities());
+        return remoteWebDriver;
+    }
+
+    private DesiredCapabilities getDesiredCapabilities() {
+
+        final SeleniumContainers seleniumContainers = seleniumContainersInstance.get();
+        switch (seleniumContainers.getBrowser()) {
+            case "firefox": return DesiredCapabilities.firefox();
+            case "chrome": return DesiredCapabilities.chrome();
```
MatousJobanek: Probably stupid question - with this, there is no way to pass parameters to the browsers/Drone itself, isn't it? I mean using arquillian.xml... 
Because of my environment problems I cannot get your ftest-drone run, so I cannot try it - that's why I'm asking.

lordofthejars: We are using `browser` drone property to get the browser to use. The problem is that there is only images for Firefox and Chrome so it is a protection against users wants to use with another browser.

MatousJobanek: Yeah, I noticed that you parse the arq.xml for the browser property. What I meant are other properties that are handled in FirefoxDriverFactory or ChromeDriverFactory (that you have override). Eg. profile for firefox or additional chromeOptions for chrome that I'll need to support: https://issues.jboss.org/browse/ARQ-2037

lordofthejars: Right, so what we can do is implement in Drone (for what I understand they have not implemented yet in Drone) and then add support in Cube Drone. Or we can just add as cubedrone property.

PR #405, opened 2016-07-29
Issue #404. Adds Arquillian Graphene support.

lordofthejars:
This change is [Reviewable](https://reviewable.io/reviews/arquillian/arquillian-cube/405)

### DIFF
--- a/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerExtension.java
+++ b/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerExtension.java
@@ -31,13 +31,21 @@ public class CubeDockerExtension implements LoadableExtension {
 
         // Arquillian Container integration
         // Only register if container-test-spi is on classpath
-        if(Validate.classExists("org.jboss.arquillian.container.test.spi.client.deployment.AuxiliaryArchiveAppender")) {
+        if(Validate.classExists("org.jboss.arquillian.container.test.spi.client.deployment.AuxiliaryArchiveAppender")
+                && doesNotContainStandaloneExtension()) {
             builder.observer(DockerServerIPConfigurator.class);
             builder.observer(CubeDockerAutoStartConfigurator.class);
         } else {
             // Arquillian in Standalone mode
             builder.observer(StandaloneAutoStartConfigurator.class);
         }
+    }
+
+    private boolean doesNotContainStandaloneExtension() {
+        final boolean junitStandalone = Validate.classExists("org.jboss.arquillian.junit.standalone.JUnitStandaloneExtension");
+        final boolean testngStandalone = Validate.classExists("org.jboss.arquillian.testng.standalone.TestNGStandaloneExtension");
+
+        return !junitStandalone && !testngStandalone;
     }
 
 }
--- a/docker/drone/src/main/java/org/arquillian/cube/docker/graphene/location/DockerCubeCustomizableURLResourceProvider.java
+++ b/docker/drone/src/main/java/org/arquillian/cube/docker/graphene/location/DockerCubeCustomizableURLResourceProvider.java
@@ -1,0 +1,335 @@
+package org.arquillian.cube.docker.graphene.location;
+
+import org.arquillian.cube.docker.drone.SeleniumContainers;
+import org.arquillian.cube.docker.impl.client.CubeDockerConfiguration;
+import org.arquillian.cube.docker.impl.client.config.CubeContainer;
+import org.arquillian.cube.docker.impl.client.config.DockerCompositions;
+import org.arquillian.cube.docker.impl.client.config.PortBinding;
+import org.jboss.arquillian.core.api.Instance;
+import org.jboss.arquillian.core.api.annotation.Inject;
+import org.jboss.arquillian.graphene.spi.configuration.GrapheneConfiguration;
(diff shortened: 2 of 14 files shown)


docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerExtension.java
```diff
@@ -40,4 +41,11 @@ public void register(ExtensionBuilder builder) {
         }
     }
 
+    private boolean doesNotContainStandaloneExtension() {
+        final boolean junitStandalone = Validate.classExists("org.jboss.arquillian.junit.standalone.JUnitStandaloneExtension");
+        final boolean testngStandalone = Validate.classExists("org.jboss.arquillian.testng.standalone.TestNGStandaloneExtension");
```
bartoszmajsak: What about other test runners?

docker/drone/src/main/java/org/arquillian/cube/docker/graphene/location/DockerCubeCustomizableURLResourceProvider.java
```diff
@@ -0,0 +1,335 @@
...
+            for (Map.Entry<String, CubeContainer> cubeContainerEntry : containers) {
+
+                // need to skip vnc and selenium container
+                final SeleniumContainers seleniumContainers = seleniumContainersInstance.get();
+                if (isSeleniumContainers(cubeContainerEntry, seleniumContainers)) {
+                    continue;
+                }
+
+                final CubeContainer cubeContainer = cubeContainerEntry.getValue();
+                final Collection<PortBinding> portBindings = cubeContainer.getPortBindings();
```
bartoszmajsak: Shouldn't we simply return collection (or empty collection if not defined otherwise?). This would simplify the code a bit.

docker/drone/src/main/java/org/arquillian/cube/docker/graphene/location/DockerCubeCustomizableURLResourceProvider.java
```diff
@@ -0,0 +1,335 @@
...
+            return this;
+        }
+
+        public URL build() throws MalformedURLException {
+
+            if (protocol == null || protocol.isEmpty()) {
+                throw new IllegalArgumentException("Protocol cannot be null or empty");
+            }
+
+            if (host == null || host.isEmpty()) {
```
bartoszmajsak: `String.isEmpty()` only in JDK6 and above :\ I assume we don't support JDK5 anymore as probably other libraries we use are not.

docker/ftest-graphene/pom.xml
```diff
@@ -0,0 +1,79 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>org.jboss.arquillian.junit</groupId>
+            <artifactId>arquillian-junit-standalone</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.arquillian.graphene</groupId>
+            <artifactId>graphene-webdriver</artifactId>
+            <version>2.1.0.Final</version>
```
bartoszmajsak: Let's make it a property in the parent if used in other modules too. `<version.graphenec>` or sth similar

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerExtension.java
lordofthejars: I think there is only support for these two test runners

docker/drone/src/main/java/org/arquillian/cube/docker/graphene/location/DockerCubeCustomizableURLResourceProvider.java
lordofthejars: I thought the same, the problem is that this class and his aggregator classes contains a lot of `Collections`, and not all of them are used always. So I thought the same of you, but then we are going to instantiate a lot of `ArrayList` that are not going to be used, but they need to be destroyed. So I decided to do that kind of approach instead of creating all the lists. 

Another option could be only create an `ArrayList` for these two arguments, but I don't like this since there are some attributes initialized and anothers not.

docker/drone/src/main/java/org/arquillian/cube/docker/graphene/location/DockerCubeCustomizableURLResourceProvider.java
lordofthejars: In fact minimal is JDK7

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerExtension.java
bartoszmajsak: Or are those the only one "tested/covered" somehow?

lordofthejars: Well I think Arquillian can only work with JUnit and TestNG.

bartoszmajsak: Hmmm... what about other test runners we have? Like spock for example?

lordofthejars: True.

Issue #403, opened 2016-07-27
Add configuration parameter in Drone integration for customizing image


lordofthejars:
Add configuration parameter in Drone integration for customizing browser docker image by setting the docker image name or a Dockerfile location

Create custom browser image property called `browserImage` to use custom browser image instead of the official one.

Create custom browser dockerfile property called `browserDockerfileDirectory` to build custom browser image instead of the official one.

PR #406, opened 2016-08-01
Issue #403. Adds customizable browser images for cube-drone.

lordofthejars:
This change is [Reviewable](https://reviewable.io/reviews/arquillian/arquillian-cube/406)

### DIFF
--- a/docker/drone/src/main/java/org/arquillian/cube/docker/drone/CubeDroneConfiguration.java
+++ b/docker/drone/src/main/java/org/arquillian/cube/docker/drone/CubeDroneConfiguration.java
@@ -20,12 +20,12 @@ public class CubeDroneConfiguration {
      */
     private String browserImage = null;
     /**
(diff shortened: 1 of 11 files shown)


docker/drone/src/main/java/org/arquillian/cube/docker/drone/CubeDroneConfiguration.java
```diff
@@ -15,6 +15,17 @@
...
+     * Docker image to be used as custom browser image instead of the official one.
+     */
+    private String browserImage = null;
+    /**
+     * Dockerfiel location to be used to built custom docker image instead of the official one.
+     * This property has preference over browserImage.
+     *
+     * @see org.arquillian.cube.docker.drone.CubeDroneConfiguration#browserImage
+     */
+    private String browserDockerfileDirectory = null;
```
bartoszmajsak: I would name it `Location` instead of `Directory`. Wouldn't it be better?

Issue #287, opened 2016-02-08
Any idea how to incorporate Recorder to Cube?


smiklosovic:
I am just wondering if there is any specific feature you would like to see in Recorder when added to classpath while Cube is in charge.

Maybe there could be events listening on After/Before Cube Started/Stopped/Created or something similar which would appear in the final log.

Maybe logs from container could be appended to the final report as well? I can imagine that e.g. you would get dump of catalina.out or basically any file you want so you would know what happened there.

Any other idea is highly welcome!

aslakknutsen:
Yea, good question. Adding some of the configuration and I guess the lifecycle events would be a start.

This issue is semi related: https://github.com/arquillian/arquillian-cube/issues/271

lordofthejars:
Basically it means creating the events in SPI and modify this class https://github.com/arquillian/arquillian-cube/blob/master/core/src/main/java/org/arquillian/cube/impl/client/CubeLifecycleController.java to fire the events right?

And add only this

```
@Inject
    private Event<CubeStarted> startedEvent;
```

lordofthejars:
Wait Wait this is what we want right? https://github.com/arquillian/arquillian-cube/blob/master/docker/docker/src/main/java/org/arquillian/cube/docker/impl/model/DockerCube.java#L99 So we do not need to do anything. It is already there.

aslakknutsen:
It might be easier to have as a Observer that observes the After/Before Cube events. That would allow us to dynamically add the Observer only if Recorder is on classpath and not bind Cube to the Reporter/Recorder

lordofthejars:
Yes agreed but this observer should be in its own module, but the events
are already developed. It is a matter of implementing observers

aslakknutsen:
I think we should do as we do with containers. As it to Cube core as a provided dependency. Recorder is added by user. It's so so many ways to configure recorder already

lordofthejars:
yes, so it is a new module in Cube which if users add in the classpath then
everything is recorded.

El dt., 9 febr. 2016 a les 17:45, Aslak Knutsen (<notifications@github.com>)
va escriure:

aslakknutsen:
I was arguing the opposite, keep it in core. A single observer that is
loaded if recorder is on class path. Core it self does not exist recorder
as a dependency.

That will allow the user to configure Cube and add recorder on will and not
have to add Cube-recorder as well.

lordofthejars:
[...]

PR #412, opened 2016-08-05
[issue #287] adds Arquillian Recorder Reporter integration

lordofthejars:
This change is [Reviewable](https://reviewable.io/reviews/arquillian/arquillian-cube/412)

### DIFF
--- a/core/src/main/java/org/arquillian/cube/impl/client/CubeExtension.java
+++ b/core/src/main/java/org/arquillian/cube/impl/client/CubeExtension.java
@@ -9,6 +9,7 @@ import org.arquillian.cube.impl.client.enricher.CubeControllerProvider;
 import org.arquillian.cube.impl.client.enricher.CubeIDResourceProvider;
 import org.arquillian.cube.impl.client.enricher.HostIpTestEnricher;
 import org.arquillian.cube.impl.client.enricher.HostPortTestEnricher;
+import org.arquillian.cube.impl.reporter.TakeCubeInformation;
 import org.jboss.arquillian.container.test.spi.client.deployment.AuxiliaryArchiveAppender;
 import org.jboss.arquillian.core.spi.LoadableExtension;
 import org.jboss.arquillian.test.spi.TestEnricher;
@@ -31,7 +32,7 @@ public class CubeExtension implements LoadableExtension {
 
         // Arquillian Container integration
         // Only register if container-test-spi is on classpath
-        if(Validate.classExists("org.jboss.arquillian.container.test.spi.client.deployment.AuxiliaryArchiveAppender")) {
+        if (Validate.classExists("org.jboss.arquillian.container.test.spi.client.deployment.AuxiliaryArchiveAppender")) {
             builder.observer(ProtocolMetadataUpdater.class)
                    .observer(CubeContainerLifecycleController.class)
                    .observer(ContainerConfigurationController.class)
@@ -39,8 +40,13 @@ public class CubeExtension implements LoadableExtension {
             builder.service(AuxiliaryArchiveAppender.class, CubeAuxiliaryArchiveAppender.class);
         }
         // Only register if container-test-impl is on classpath
-        if(Validate.classExists("org.jboss.arquillian.container.test.impl.enricher.resource.OperatesOnDeploymentAwareProvider")) {
+        if (Validate.classExists("org.jboss.arquillian.container.test.impl.enricher.resource.OperatesOnDeploymentAwareProvider")) {
             builder.service(ResourceProvider.class, CubeIDResourceProvider.class);
+        }
+
+        // Only if recorder-reporter is in classpath we should provide reporting capabilities.
+        if (Validate.classExists("org.arquillian.recorder.reporter.ReporterExtension")) {
+            builder.observer(TakeCubeInformation.class);
         }
     }
 
--- a/core/src/main/java/org/arquillian/cube/impl/reporter/TakeCubeInformation.java
+++ b/core/src/main/java/org/arquillian/cube/impl/reporter/TakeCubeInformation.java
@@ -1,0 +1,43 @@
+package org.arquillian.cube.impl.reporter;
+
+import org.arquillian.cube.spi.Cube;
+import org.arquillian.cube.spi.CubeRegistry;
+import org.arquillian.cube.spi.event.lifecycle.AfterAutoStop;
+import org.arquillian.cube.spi.metadata.CanReportMetrics;
+import org.arquillian.recorder.reporter.PropertyEntry;
+import org.arquillian.recorder.reporter.Reportable;
+import org.arquillian.recorder.reporter.event.PropertyReportEvent;
+import org.jboss.arquillian.core.api.Event;
+import org.jboss.arquillian.core.api.annotation.Inject;
+import org.jboss.arquillian.core.api.annotation.Observes;
+
+import java.util.List;
+
+/**
+ * Observer that is executed before suite and prints all metrics reported by registered cubes.
+ */
+public class TakeCubeInformation {
+
+    @Inject
+    Event<PropertyReportEvent> propertyReportEvent;
+
+    public void generatReportEntries(@Observes AfterAutoStop event, CubeRegistry cubeRegistry) {
+
+        if (cubeRegistry == null) {
+            return;
+        }
+
+        final List<Cube<?>> reportableCubes = cubeRegistry.getByMetadata(CanReportMetrics.class);
+
+        for (Cube cube: reportableCubes) {
+            final CanReportMetrics metadata = (CanReportMetrics) cube.getMetadata(CanReportMetrics.class);
+            final Reportable report = metadata.report();
+
+            if (report instanceof PropertyEntry) {
+                propertyReportEvent.fire(new PropertyReportEvent((PropertyEntry) report));
+            }
+        }
+
+    }
+
+}
--- a/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/metadata/ReportMetrics.java
+++ b/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/metadata/ReportMetrics.java
@@ -1,0 +1,102 @@
+package org.arquillian.cube.docker.impl.client.metadata;
+
+import org.arquillian.cube.docker.impl.client.config.CubeContainer;
+import org.arquillian.cube.docker.impl.model.DockerCube;
+import org.arquillian.cube.spi.metadata.CanReportMetrics;
+import org.arquillian.recorder.reporter.Reportable;
+import org.arquillian.recorder.reporter.model.entry.GroupEntry;
+import org.arquillian.recorder.reporter.model.entry.KeyValueEntry;
+import org.arquillian.recorder.reporter.model.entry.table.TableCellEntry;
+import org.arquillian.recorder.reporter.model.entry.table.TableRowEntry;
+
+import java.security.Key;
+import java.util.Collection;
+
+/**
+ * Reporting metrics capabilities for Docker Cube.
+ */
+public class ReportMetrics implements CanReportMetrics {
+
+    private DockerCube dockerCube;
+
+    public ReportMetrics(DockerCube dockerCube) {
+        this.dockerCube = dockerCube;
+    }
+
+    @Override
+    public Reportable report() {
+
+        GroupEntry groupEntry = new GroupEntry(dockerCube.getId());
+
+        boolean error;
+        switch(dockerCube.state()) {
+            case START_FAILED:
+            case CREATE_FAILED:
+            case STOP_FAILED:
+            case DESTORY_FAILED:
+                error = true;
+            break;
+
+            default: error = false;
+        }
+
+        KeyValueEntry errorKeyValue = new KeyValueEntry("Error during lifecycle", Boolean.toString(error));
+        groupEntry.getPropertyEntries().add(errorKeyValue);
+
+        KeyValueEntry startingTime = new KeyValueEntry("Starting Time", String.format("%s ms", dockerCube.getStartingTimeInMillis()));
+        KeyValueEntry stoppingTime = new KeyValueEntry("Stopping Time", String.format("%s ms", dockerCube.getStoppingTimeInMillis()));
+
+        groupEntry.getPropertyEntries().add(startingTime);
+        groupEntry.getPropertyEntries().add(stoppingTime);
+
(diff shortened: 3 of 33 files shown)


core/src/main/java/org/arquillian/cube/impl/reporter/TakeCubeInformation.java
```diff
@@ -0,0 +1,43 @@
...
+import org.jboss.arquillian.core.api.Event;
+import org.jboss.arquillian.core.api.annotation.Inject;
+import org.jboss.arquillian.core.api.annotation.Observes;
+
+import java.util.List;
+
+/**
+ * Observer that is executed before suite and prints all metrics reported by registered cubes.
+ */
+public class TakeCubeInformation {
```
bartoszmajsak: I think we can find a better name :)

core/src/main/java/org/arquillian/cube/impl/client/CubeExtension.java
```diff
@@ -31,17 +32,22 @@ public void register(ExtensionBuilder builder) {
...
         // Only register if container-test-impl is on classpath
-        if(Validate.classExists("org.jboss.arquillian.container.test.impl.enricher.resource.OperatesOnDeploymentAwareProvider")) {
+        if (Validate.classExists("org.jboss.arquillian.container.test.impl.enricher.resource.OperatesOnDeploymentAwareProvider")) {
             builder.service(ResourceProvider.class, CubeIDResourceProvider.class);
         }
+
+        // Only if recorder-reporter is in classpath we should provide reporting capabilities.
+        if (Validate.classExists("org.arquillian.recorder.reporter.ReporterExtension")) {
+            builder.observer(TakeCubeInformation.class);
+        }
```
bartoszmajsak: As a general thought - I'm not sure if this way of validation is the best one. If we change package names or shuffle classes around, or even change their names this will be all broken. We should think about some other way of detection in the nearest future.

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/metadata/ReportMetrics.java
```diff
@@ -0,0 +1,102 @@
...
+        this.dockerCube = dockerCube;
+    }
+
+    @Override
+    public Reportable report() {
+
+        GroupEntry groupEntry = new GroupEntry(dockerCube.getId());
+
+        boolean error;
+        switch(dockerCube.state()) {
```
bartoszmajsak: If that's an enum we can use `EnumSet.of(....).contains(dockerCube.state())` instead of switch.

docker/docker/src/main/java/org/arquillian/cube/docker/impl/model/DockerCube.java
```diff
@@ -139,10 +155,14 @@ public void stop() throws CubeControlException {
...
         try {
             lifecycle.fire(new BeforeStop(id));
+
+            long currentTime = System.currentTimeMillis();
             try {
                 executor.stopContainer(id);
             } catch(NotFoundException e) {
             } catch (NotModifiedException e) {}
+            this.stoppingTimeInMillis = System.currentTimeMillis() - currentTime;
+
```
bartoszmajsak: Just a thought - can't we collect metrics through events instead of keeping them as a part of the object?

docker/drone/src/main/java/org/arquillian/cube/docker/drone/util/VideofileDestination.java
```diff
@@ -0,0 +1,62 @@
...
+import java.io.File;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+public class VideoFileDestination {
+
+    private static final File DEFAULT_LOCATION_OUTPUT_MAVEN = new File("target");
+    private static final File DEFAULT_LOCATION_OUTPUT_GRADLE = new File("build");
+
+    private static final File MAVEN_REPORT_DIR = new File(DEFAULT_LOCATION_OUTPUT_MAVEN, "surefire-reports");
```
bartoszmajsak: What if we run tests using failsafe?

docker/recorder/src/main/java/org/arquillian/cube/docker/impl/client/reporter/TakeDockerEnvironment.java
```diff
@@ -0,0 +1,189 @@
...
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Class that reports generic Docker information like orchestration or docker version.
+ */
+public class TakeDockerEnvironment {
```
bartoszmajsak: Why `Take`? Does it really explain what is the responsibility for the test?

docker/recorder/src/main/java/org/arquillian/cube/docker/impl/client/reporter/TakeDockerEnvironment.java
```diff
@@ -0,0 +1,189 @@
...
+            screenshotEntry.setWidth(bufferedImage.getWidth());
+            screenshotEntry.setHeight(bufferedImage.getHeight());
+            screenshotEntry.setSize(String.valueOf(imageFile.length()));
+            screenshotEntry.setMessage("Docker Composition before executing tests.");
+
+            return screenshotEntry;
+        } catch (IOException e) {
+            log.log(Level.WARNING, String.format("Docker compositions schema could not be generated because of %s.", e));
+        }
+
```
bartoszmajsak: I don't know the API here but shouldn't we close `bufferedImage` here?

docker/recorder/src/main/java/org/arquillian/cube/docker/impl/client/reporter/TakeDockerEnvironment.java
```diff
@@ -0,0 +1,189 @@
...
+        return null;
```
bartoszmajsak: Maybe some null object instead of just `null`?

docker/recorder/src/test/java/org/arquillian/cube/docker/impl/client/reporter/TakeDockerEnvironmentTest.java
```diff
@@ -0,0 +1,134 @@
...
+        final PropertyEntry propertyDockerInfoEntry = rootEntries.get(0);
+        assertThat(propertyDockerInfoEntry).isInstanceOf(GroupEntry.class);
+
+        GroupEntry dockerInfoEntry = (GroupEntry) propertyDockerInfoEntry;
+        final List<PropertyEntry> propertyEntries = dockerInfoEntry.getPropertyEntries();
+
+        assertThat(propertyEntries).hasSize(5);
+
+        assertThat(propertyEntries.get(0)).isEqualToComparingFieldByField(new KeyValueEntry("Version", "1.1.0"));
+        assertThat(propertyEntries.get(1)).isEqualToComparingFieldByField(new KeyValueEntry("OS", "linux"));
```
bartoszmajsak: Don't we have `equals` for `KeyValueEntry`? If so - we can use assertion for lists, something list `containsExactly`. Then we don't even have to assert the size.

core/src/main/java/org/arquillian/cube/impl/reporter/TakeCubeInformation.java
lordofthejars: Basically I followed the same convention used in Recorder integrations. They use this Take + <blabla>. We can change it but this is the reason why I used this name

core/src/main/java/org/arquillian/cube/impl/client/CubeExtension.java
lordofthejars: Yes but it is used all over Arquillian project.https://github.com/arquillian/arquillian-core/blob/c2ae438796b4bab56a86c6246982ecfe038f4a78/core/spi/src/main/java/org/jboss/arquillian/core/spi/LoadableExtension.java#L46 What we can do is add a new method that receives a Class object and inside to the getFullQualifiedName thing an call the String one, so at least we are not breaking anything and in future we can use the Class approach.

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/metadata/ReportMetrics.java
lordofthejars: But then we will have a list of if/else

docker/docker/src/main/java/org/arquillian/cube/docker/impl/model/DockerCube.java
lordofthejars: I thought the same but notice that there are some before/after events. So if we took there it means that we could add to the time, the duration of the execution of events thrown there. So I decided that the most real time was doing in that way, although as you said it seems event would be better. WDYT?

docker/drone/src/main/java/org/arquillian/cube/docker/drone/util/VideofileDestination.java
lordofthejars: Good point that I didn't take into consideration.

docker/recorder/src/main/java/org/arquillian/cube/docker/impl/client/reporter/TakeDockerEnvironment.java
lordofthejars: Yes just to follow same convention used in other recorder integrations (graphene, drone, ...)

docker/recorder/src/test/java/org/arquillian/cube/docker/impl/client/reporter/TakeDockerEnvironmentTest.java
lordofthejars: No sadly the Recorder entries has no equals implemented. We can do it as a separate PR in Recorder, for this reason now it is in this way.

docker/recorder/src/main/java/org/arquillian/cube/docker/impl/client/reporter/TakeDockerEnvironment.java
lordofthejars: Yes I think try with resources would be the best way here.

lordofthejars: BufferedImage has no close method nor implements `AutoClosable` so it is not required

core/src/main/java/org/arquillian/cube/impl/client/CubeExtension.java
bartoszmajsak: Let's think about it. Shall we creat a jira issue/card for it so we won't forget to address it?

core/src/main/java/org/arquillian/cube/impl/reporter/TakeCubeInformation.java
bartoszmajsak: Ok, clear.

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/metadata/ReportMetrics.java
bartoszmajsak: I don't see where you would need an if-else. You simply do something like

```java
error = EnumSet.of(START_FAILED, CREATE_FAILED, STOP_FAILED, DESTORY_FAILED)
               .contains(dockerCube.state());
```

or am I missing something?

docker/docker/src/main/java/org/arquillian/cube/docker/impl/model/DockerCube.java
bartoszmajsak: Fair point. Not sure what is best. I think I have to spend a bit more time reading the code to make better justification. If you feel like this is good way to go I'm fine with that.

docker/recorder/src/main/java/org/arquillian/cube/docker/impl/client/reporter/TakeDockerEnvironment.java
bartoszmajsak: Ok clear, I was just looking at the code here, not anywhere from the IDE.

docker/recorder/src/test/java/org/arquillian/cube/docker/impl/client/reporter/TakeDockerEnvironmentTest.java
bartoszmajsak: Then maybe we should bring them? WDYT?

core/src/main/java/org/arquillian/cube/impl/client/CubeExtension.java
lordofthejars: We should add a card as community tasks

bartoszmajsak: use 'help wanted' label then

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/metadata/ReportMetrics.java
lordofthejars: Right

docker/recorder/src/test/java/org/arquillian/cube/docker/impl/client/reporter/TakeDockerEnvironmentTest.java
lordofthejars: Yes let's do it after Cube is merge, if not we are going to spend all the time trying to fix things of other extensions without releasing Cube, and for now we can live without that changes.

bartoszmajsak: Sure, that's what I was thinking about. Not as a part of this PR, but in general.

docker/docker/src/main/java/org/arquillian/cube/docker/impl/model/DockerCube.java
lordofthejars: Yes as I said I know it is not the best way in terms of code, but at least it is the best in terms of accuracy. I can try to put in higher priority the event which will not be the same times but really close.

lordofthejars: But as you said when we start adding more and more metrics we could find another kind of integrations. This is a simple one. Maybe an integration with http://metrics.dropwizard.io/3.1.0/

lordofthejars:
You can configure a custom directory too. We can put them in target/videos if you think it is better. In that case it was only a decision, but we can change

PR #414, opened 2016-08-06
Kubernetes

iocanel:
This change is [Reviewable](https://reviewable.io/reviews/arquillian/arquillian-cube/414)

### DIFF
--- a/docs/kubernetes.adoc
+++ b/docs/kubernetes.adoc
@@ -1,0 +1,194 @@
+== Kubernetes
+
+The kubernetes extension helps you write and run integration tests for your Kubernetes/Openshift application.
+
+=== Overview
+
+This extension, wll create and manage a temporary namespace for your tests, apply all Kubernetes resources required to create your environment and once everything is ready it will run your tests. The tests will be enriched with resources required to access services. Finally when testing is over it will cleanup everything.
+
+This extension, will not mutate neither your containers *(by deploying, reconfiguring etc)* nor your Kubernetes resources, and takes a black box approach to testing.
+
+=== Modules
+The main modules of this extension are the following:
+
+- Kubernetes (default feature set)
+- Openshift (work in progress)
+- Fabric8 Microservices Platform (Fabric8 label and annotation support)
+
+=== Features
+- Hybrid *(in or out of Kubernetes/Openshift)*
+- Advanced namespace management
+- Dependency management *(for maven based projects)*
+- Auto align with Docker Registry
+- Enrichers for:
+    - Kubernetes/Openshift client
+    - Pods
+    - Replication Controllers
+    - Services
+- Integration with Fabric8 Modules:
+    - link:https://fabric8.io/guide/mavenPlugin.html[Fabric8 Maven Plugin]
+    - link:https://fabric8.io/guide/fabric8DevOps.html[Microservices Platform]
+- "Bring your own client" support
+
+
+=== Configuring the extension
+
+The plugin can be configured using the traditional arquillian.xml, via system properties or evnironment variables (in that particular order).
+Which means that for every supported configuration parameter, the arquillian.xml will be looked up first, if it doesn't contain an entry, the system properties will be used.
+If no result has been found so far the environment variables will be used.
+
+**Note:** When checking for environment variables, property names will get capitalized, and symbols like "." will be converted to "_".
+For example **foo.bar.baz** will be converted to **FOO_BAR_BAZ**.
+
+
+[width="80%"]
+|=========================================================================================================================================
+| Option                              | Type         | Env | Description                                                                  |
+| kubernetes.master                   | URL          | Any | The URL to the Kubernetes master                                             |
+| kubernetes.domain                   | String       | OSE | Domain to use for creating routes for services                               |
+| docker.registry                     | String       | Any | The docker registry                                                          |
+| namespace.use.existing              | String       | Any | Don't generate a namespace use the specified one instead                     |
+| namespace.lazy.enabled              | Bool (true)  | Any | Should the specified namespace be created if not exists, or throw exception? |
+| namespace.cleanup.enabled           | Bool (true)  | Any | Flag to destroy the namespace after the end of the test suite                |
+| namespace.cleanup.confirm.enabled   | Bool (false) | Any | Flag to ask for confirmation to delete the namespace                         |
+| namespace.cleanup.timeout           | Long         | Any | Time to wait when cleaning up the namespace                                  |
+| env.init.enabled                    | Bool (true)  | Any | Flag to initialize the environment (apply kubernetes resources)              |
+| env.config.url                      | URL          | Any | URL to the Kubernetes JSON (defaults to classpath resource kubernetes.json)  |
+| env.config.resource.name            | String       | Any | Option to select a different classpath resource (ohter than kubernetes.json) |
+| env.dependencies                    | List         | Any | Whitespace separated list of URLs to more dependency kubernetes.json         |
+| wait.timeout                        | Long (5mins) | Any | The total ammount of time to wait until the env is ready                     |
+| wait.poll.interval                  | Long (5secs) | Any | The poll interval to use for checking if the environment is ready            |
+| wait.for.service.list               | Long (5secs) | Any | Explicitly specify a list of service to wait upon                            |
+| wait.for.service.connection.enabled | Bool (false) | Any | Flag to specify if we should wait for an actual connection to the service    |
+| wait.for.service.connection.timeout | Long (10secs)| Any | The amount of time we should wait for each socket connection.                |
+| ansi.logger.enabled                 | Bool (true)  | Any | Flag to enable colorful output                                               |
+| kubernetes.client.creator.class.name| Bool (true)  | Any | Fully qualified class name of a kubernetes client creator class (byon)       |
+|=========================================================================================================================================
+
+=== Namespaces
+
+The default behavior of the extension is to create a unique namespace per test suite. The namespace is created Before the suite is started and destroyed in the end.
+For debugging purposes, you can set the **namespace.cleanup.enabled** to false and keep the namespace around.
+
+In other cases you may find it useful to manually create and manage the environment rather than having **arquillian** do that for you.
+In this case you can use the **namespace.use.existing** option to select an existing namespace. This option goes hand in hand with **env.init.enabled** which can be
+used to prevent the extension from modifying the environment.
+
+### Creating the environment
+After creating or selecting an existing namespace, the next step is the environment preparation. This is the stage where all the required Kubernetes configuration will be applied.
+Out of the box, the extension will use the classpath and try to find a resource named **kubernetes.json** or **kubernetes.yaml***. The name of the resource can be changed using the **env.config.resource.name**.
+Of course it is also possible to specify an external resource by URL using the **env.config.url**.
+
+Either way, it is possible that the kubernetes configuration used, depends on other configurations. It is also possible that your environment configuration is split in multiple files.
+To cover cases like this the **env.dependencies** is provided which accepts a space separated list of URLs.
+
+**Note:** Out of the box mvn urls are supported, so you can use values like: **mvn:my.groupId/artifactId/1.0.0/json/kubernetes** (work in progress)
+
+**Also:** If your project is using maven and dependencies like the above are expressed in the pom, the will be used *automatically*. (work in progress)
+
+
+=== Readiness and waiting
+Creating an environment does not guarantee its readiness. For example a Docker image may be required to get pulled by a remote repository and this make take even several minutes.
+Running a test against a Pod which is not Running state is pretty much pointless, so we need to wait until everything is ready.
+
+This extension will wait up to **wait.timeout** until everything is up and running. Everything? It will wait for all Pods and Service *(that were created during the test suite initialization)* to become ready.
+It will poll them every **wait.poll.interval** milliseconds. For services there is also the option to perform a simple "connection test"  by setting the flag **wait.for.service.connection.enabled** to true.
+In this case it will not just wait for the service to ready, but also to be usable/connectable.
+
+=== Immutable infrastructure and integration testing
+
+As mentioned in the overview, this extension will not try to deploy your tests, inside an application container.
+It doesn't need nor want to know what runs inside your docker containers, nor will try to mess with it.
+It doesn't even need to run inside Kubernetes (it can just run in your laptop and talk to the kubernetes master).
+
+So what exactly is your test case going to test?
+
+The test cases are meant to consume and test the provided services and assert that the environment is in the expected state.
+
+The test case may obtain everything it needs, by accessing the Kubernetes resources that are provided by the plugin as @ArquillianResources (see resource providers below).
+
+=== Resource Providers
+
+The resource providers available, can be used to inject to your test cases the following resources:
+
+- A kubernetes client as an instance of KubernetesClient
+- Session object that contains information (e.g. the namespace) or the uuid of the test session.
+- Pods *(by id or as a list of all pods created during the session)*
+- Replication Controllers *(by id or as a list of all replication controllers created during the session)*
+- Services *(by id or as a list of all services created during the session)*
+
+Here's a small example:
+
+[source, java]
+.ExampleTest.java
+----
+    @RunWith(Arquillian.class)
+    public class ExampleTest {
+
+     @ArquillianResource
+     KubernetesClient client;
+
+     @ArquillianResource
+     Session session;
+
+      @Test
+      public void testAtLeastOnePod() throws Exception {
+       assertThat(client).pods().runningStatus().filterNamespace(session.getNamespace()).hasSize(1);
+      }
+    }
+----
+
+The test code above, demonstrates how you can inject an use inside your test the *KubernetesClient* and the *Session* object. It also demonstrates the use of **kubernetes-assertions** which is a nice little library based on [assert4j](http://assertj.org) for performing assertions on top of the Kubernetes model.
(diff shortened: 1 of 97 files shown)


kubernetes/fabric8/src/main/java/org/arquillian/cube/kubernetes/fabric8/impl/visitor/SecretsAndServiceAccountVisitor.java
```diff
@@ -0,0 +1,126 @@
...
+                if (SecretKeys.isSecretKey(key)) {
+                    SecretKeys keyType = SecretKeys.fromValue(key);
+                    for (String name : Secrets.getNames(value)) {
+                        Map<String, String> data = new HashMap<>();
+
+                        Secret secret = null;
+                        try {
+                            secret = client.secrets().inNamespace(configuration.getNamespace()).withName(name).get();
+                        } catch (Exception e) {
+                            // ignore - probably doesn't exist
```
lordofthejars: Should we add a log line?

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/ConfigurationRegistar.java
```diff
@@ -0,0 +1,54 @@
...
+        Map<String, String> config = arquillianDescriptor.extension(EXTENSION_NAME).getExtensionProperties();
+        configurationProducer.set(DefaultConfiguration.fromMap(config));
+        configureProtocolHandlers(config);
+    }
+    
+    private static void configureProtocolHandlers(Map<String, String> conf) {
+        Set<String> handlers = new LinkedHashSet<>();
+        handlers.addAll(Strings.splitAndTrimAsList(System.getProperty(JAVA_PROTOCOL_HANDLER, ""), " "));
+        handlers.addAll(Strings.splitAndTrimAsList(conf.containsKey(PROTOCOL_HANDLERS) ? conf.get(PROTOCOL_HANDLERS) : DEFAULT_MAVEN_PROTOCOL_HANDLER, " "));
+        System.setProperty(JAVA_PROTOCOL_HANDLER, Strings.join(handlers, " "));
```
lordofthejars: I see that this property is set but never clear. Is this normal or safe? I am thinking in case of splitting tests in parallel in same VM for example.

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/enricher/PodResourceProvider.java
```diff
@@ -0,0 +1,24 @@
...
+public class PodResourceProvider extends AbstractKubernetesResourceProvider {
+
+    @Override
+    public boolean canProvide(Class<?> type) {
+        return Pod.class.isAssignableFrom(type);
+    }
+
+    @Override
+    public Object lookup(ArquillianResource resource, Annotation... qualifiers) {
+        return getClient().pods().inNamespace(getSession().getNamespace()).withName(getName(qualifiers)).get();
```
lordofthejars: Don't know the API deeply but is it possible in that train of calls to get any null?

sundrio: No, its not possible to get an NPE here (it would defeat the whole purpose of the DSL).

kubernetes/fabric8/src/main/java/org/arquillian/cube/kubernetes/fabric8/impl/visitor/SecretsAndServiceAccountVisitor.java
sundrio: I think that the whole try/catch should go. No checked execeptions are thrown from this piece of code.

lordofthejars:
Overall it looks good to me, I have done some comments but are more questions rather than bugs.

Just two things which I think it might be important, one to fix in this PR another one that we would do it in the future.

* First thing that could be added in this PR is next thing: In Cube we usually create an `ftest` module that basically is like a _hello world_ app so a user can see how he could use in real world. Can you provide something like this? I know that the project has tests, but having a module with the example makes easier to start.

* Second, currently as we talked during the meeting that k8s, fabric8 integration will be as is without any integration with cube spi or api. For example cuberegistry does not contain anything related to kubernetes (pod, replicationcontroller, ...) or they don't implement `Cube` interface, but no worries because this is something to do it late in another PR, since the important thing is having now everything together. See that this integration with spi or api are important to reuse the reporter capabilities or chaos logic provided by Cube.

So for me if you answer the questions and provide an ftest module of kubernetes, we will be able to merge it.

bartoszmajsak:
Another thing which we might bring to master before releasing is to make sure we have license headers everywhere. This can be easily automated using IDEA or Maven plugin. I know it does not bring that much value code-wise, but it's  good to have.

docs/kubernetes.adoc
```diff
@@ -0,0 +1,194 @@
+== Kubernetes
+
+The kubernetes extension helps you write and run integration tests for your Kubernetes/Openshift application.
+
+=== Overview
+
+This extension, wll create and manage a temporary namespace for your tests, apply all Kubernetes resources required to create your environment and once everything is ready it will run your tests. The tests will be enriched with resources required to access services. Finally when testing is over it will cleanup everything.
```
bartoszmajsak: `wll` - typo

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/api/AnnotationProvider.java
```diff
@@ -0,0 +1,16 @@
+package org.arquillian.cube.kubernetes.api;
+
+import java.util.Map;
+
+/**
+ * Created by iocanel on 8/1/16.
+ */
```
bartoszmajsak: I'm not sure if we should keep this kind of headers.

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/Constants.java
```diff
@@ -0,0 +1,77 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ *  implied.  See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+package org.arquillian.cube.kubernetes.impl;
+
+public class Constants {
+
+    public static final String KUBERNETES_MASTER = "kubernetes.master";
+    public static final String KUBERNETES_NAMESPACE = "kubernetes.namespace";
```
bartoszmajsak: We have the same constants in `Configuration` - is it needed?

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/enricher/PodResourceProvider.java
lordofthejars: Ok cool

lordofthejars:
>Another thing which we might bring to master before releasing is to make sure we have license headers everywhere. This can be easily automated using IDEA or Maven plugin. I know it does not bring that much value code-wise, but it's good to have.

We don't have this headers in any cube project :P

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/ShutdownHook.java
```diff
@@ -0,0 +1,30 @@
...
+
+    private final Runnable delegate;
+
+    public ShutdownHook(Runnable delegate) {
+        this.delegate = delegate;
+    }
+
+    @Override
+    public void run() {
+        delegate.run();;
```
bartoszmajsak: No need for double `;;` ;)

bartoszmajsak:
> We don't have this headers in any cube project :P

Then it's time to bring it in ;) It's an automatic job anyway

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/SuiteListener.java
```diff
@@ -0,0 +1,52 @@
...
+    private DefaultSession session;
+
+    public void start(@Observes(precedence = 100) BeforeSuite event, Configuration configuration, Logger logger) {
+        session = new DefaultSession(configuration.getSessionId(), configuration.getNamespace(), logger);
+        session.init();
+        sessionProducer.set(session);
+        controlEvent.fire(new Start(session));
+    }
+
+    public void stop(@Observes(precedence = -100) AfterSuite event, Logger logger) {
```
bartoszmajsak: Not sure if we need `Logger` here. Is it injected btw? Didn't know that it can be done this way hehe.

lordofthejars: Well if is produced somewhere you can do it. But I am not sure if in core it is done, FWIR it is not done in Cube

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/log/AnsiLogger.java
```diff
@@ -0,0 +1,32 @@
...
+public class AnsiLogger implements Logger {
+
+    @Override
+    public void info(String msg) {
+        System.out.println(msg);
+    }
+
+    @Override
+    public void warn(String msg) {
+        System.out.println( ansi().fg(YELLOW).a(msg).reset() );
```
bartoszmajsak: Oh that is cool :) If we gonna stick to sysout maybe we can align it across the modules. The only thing is that we bring additional dep to the project, but I guess it's not really that crucial from the "client" perspective.

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/requirement/KubernetesRequirement.java
```diff
@@ -0,0 +1,37 @@
...
+import com.squareup.okhttp.Response;
+import io.fabric8.kubernetes.client.DefaultKubernetesClient;
+import io.fabric8.kubernetes.client.HttpClientAware;
+import io.fabric8.kubernetes.client.KubernetesClient;
+import org.arquillian.cube.spi.requirement.Requirement;
+import org.arquillian.cube.spi.requirement.UnsatisfiedRequirementException;
+
+import java.io.IOException;
+
+//TODO: The kubernetes client currently doesn't expose a method to do a version check. An issue has been raised, but until its done we do the work here. See https://github.com/fabric8io/kubernetes-client/issues/477.
```
bartoszmajsak: @lordofthejars let's create an issue here out of it and reference to the linked one. Otherwise it might get lost.

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/log/AnsiLogger.java
lordofthejars: I think that in this case it is because the AnsiLogger, but as you said we could spread to other pieces of cube.

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/requirement/KubernetesRequirement.java
lordofthejars: Ok

kubernetes/kubernetes/src/test/java/org/arquillian/cube/kubernetes/impl/MockTest.java
```diff
@@ -0,0 +1,130 @@
...
+        System.setProperty(Config.KUBERNETES_TRUST_CERT_SYSTEM_PROPERTY, "true");
+        System.setProperty(Config.KUBERNETES_NAMESPACE_SYSTEM_PROPERTY, "arquillian");
+        System.setProperty(Constants.NAMESPACE_TO_USE, "arquillian");
+        System.setProperty(Constants.NAMESPACE_CLEANUP_ENABLED, "true");
+        System.setProperty(Constants.ENVIRONMENT_CONFIG_URL, MockTest.class.getResource("/test-kubernetes-1.json").toString());
+    }
+
+    @AfterClass
+    public static void tearDownClass() throws IOException {
+        //MOCK.destroy();
```
bartoszmajsak: We don't do that?

bartoszmajsak:
Thanks @iocanel (or @sundrio? I'm confused now ;))!

Great job.

I've shared few minor things and if we can pull out some simple `ftest` (which we can later also use in the documentation itself) that would be enough from my point of view to bring this to upstream.

There are few other small things which I have noted down, but that's not for this PR anyway - don't want to grow it even more.

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/Constants.java
iocanel: No, let's remove them.

jwendell:
Not commenting on the patch itself, I just want to say that it didn't break our tests (we are using arquillian-cube at https://github.com/jboss-openshift/ce-arq).

Also, I've tested it using latest kubernetes-client, it worked fine for our purposes as well. Here's the trivial change I had to do: https://github.com/jwendell/arquillian-cube/commit/7213d23ed678bea291dad2b338b7e6278cfe2898

bartoszmajsak:
Thanks @jwendell for your help!

@iocanel is it safe to bump `okhttp`?

lordofthejars:
@iocanel One question, could you use this https://github.com/micro-kube/micro-kube too in ftest? Or what is the best way to run the ftest, I comment this because it will be nice if in the README of ftest user can see one way to install k8s and run the ftest.

iocanel:
Back from vacation!

Regarding okhttp, I think I had an issue with the mock tests. I can have a second look if needed.

@lordofthejars: Currently all the kubernetes/openshift bits assume that the kubernetes environment exists. If you are suggesting that we should bootstrap minikube/nanocube/minishift if available. Yeah that would make a great addition (and we now have the annotations to support it too). Do you want me to add it as part of this PR?

bartoszmajsak:
Hey @iocanel, welcome back to reality!

Since @lordofthejars is on holidays ATM I can share my $ .2. 

If this PR is now buildable on linux using simple command such as `mvn clean install` we should aim to close it. The enhancement can follow in the new one.

lordofthejars:
Is it possible to provide a ftest project before merging it?

iocanel:
So, I just rebased and I hit a couple of issues with compilation. It seems that its just a matter of adding a couple of `@RequiresDockerMachine` here and there.

I'll fix that, add an ftest and will update the PR ASAP.

In the meantime is it possible to add some kind of feedback mechanism for the incoming PRs? CirceCI maybe?

lordofthejars:
The truth is that this is something we need to discuss and do, we were
thinking in talking with you to know hiw it is done in fabric8

iocanel:
In the fabric8 land we use both a circleci and a dedicated jenkins instance.
CircleCI is used just for feedback and the jenkins instance to test and approve incoming pull requests. 

For now we don't have a kubernetes env to use for our tests, but if we had something like minicube integration, that would solve all our issues.

I'd say that for now, even a simple circleci integration would help.

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/ConfigurationRegistar.java
iocanel: AFAIR, this is something that can only be set once. Resetting or clearing the property doesn't have an actual effect from the protocol handlers perspective.

Not sure how it will work out in parallel testing, but I assume the way to go, is either use it as is, or override via JVM arguments (in which case this code will have no effect).

kubernetes/ftest-kubernetes/src/test/resources/arquillian.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0"?>
+<arquillian xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xmlns="http://jboss.org/schema/arquillian"
+    xsi:schemaLocation="http://jboss.org/schema/arquillian
+    http://jboss.org/schema/arquillian/arquillian_1_0.xsd">
+
+    <extension qualifier="kubernetes">
```
lordofthejars: I think that if no attribute is used, you don't need to add it, but leave as is, it doesn't matter, just to let you know :)

kubernetes/ftest-kubernetes/src/test/java/HelloWorldTest.java
```diff
@@ -0,0 +1,28 @@
...
+    @Named("hello-world-service")
+    @ArquillianResource
+    Service helloWorld;
+
+    @Test
+    public void shouldShowHelloWorld() throws IOException {
+        assertNotNull(helloWorld);
+        assertNotNull(helloWorld.getSpec());
+        assertNotNull(helloWorld.getSpec().getPorts());
+        assertFalse(helloWorld.getSpec().getPorts().isEmpty());
```
lordofthejars: Looks awesome to me 💃 
One question if we want to use fabric8 assertion mechanism, should be possible as well by just adding the dependency right?

iocanel: Exactly!

lordofthejars:
@bartoszmajsak looks good to me. I merge the PR, yahooooooooo. Really happy.

PR #419, opened 2016-08-09
Host

jwendell:
This change is [Reviewable](https://reviewable.io/reviews/arquillian/arquillian-cube/419)

### DIFF
--- a/openshift/openshift/pom.xml
+++ b/openshift/openshift/pom.xml
@@ -12,6 +12,7 @@
 	<properties>
 		<version.openshift_restclient>1.0.0-SNAPSHOT</version.openshift_restclient>
 		<version.jgit>4.0.0.201506090130-r</version.jgit>
+		<version.alpn>8.1.9.v20160720</version.alpn>
 	</properties>
 	<dependencies>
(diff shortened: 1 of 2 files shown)


lordofthejars:
Currently we are reworking the openshift solution to be aligned with fabric8 library. You can see the PR https://github.com/arquillian/arquillian-cube/pull/414 which should be ready to merge, it only needs to be reviewed. If you want feel free to review it too.

rcernich:
@jwendell please review #414 to verify it doesn't break anything in ce-arq.

jwendell:
ping

so, the new openshift solution makes this patch obsolete? (I just reached this bug)

jwendell:
just pulled the latest changes (like upgrading to newer undertow) but the issue is still there...

lordofthejars:
@jwendell so we need to merge this issue right?

jwendell:
imo, yes.

lordofthejars:
then it is easy, I merge it, and we see since we can always revert the change.

lordofthejars:
@jwendell did it work?

PR #430, opened 2016-08-13
Improve Docker Machine Requirement

eddumelendez:
This commit check docker machine requirement using `docker-machine ls`
rather than `docker-machine ip`. Also, fix the issue in
ftest-standalone-autostart, which can not start the machine due to the
execution of `docker-machine ip` command before to auto-start the
machine.

---
This change is [Reviewable](https://reviewable.io/reviews/arquillian/arquillian-cube/430)


lordofthejars:
Thank you very much for taking care of this.

Issue #442, opened 2016-09-22
ArquillianKubeResourceProvider should also work with javax.inject.Named


iocanel:
Currently, they require a custom annotation `org.arquillian.cube.kubernetes.annotations.Named`.
Users that read examples easily confuse that with `javax.inject.Named`. 

I think that both should be supported, to make the life of users easier.

lordofthejars:
This is something that we usually don't use in Arquillian extensions, but if you think it has sense we could provide support for both.

Maybe what we could do is add in `@ArquillianResource` an attribute named `name` to inject an specific resource.What do you think @bartoszmajsak @aslakknutsen

iocanel:
Just to clarify what I am proposing is to use reflection instead and support any kind of Named annotation without being coupled to any. 

It's just a safety net for the users. FWIW, I think that guice uses the exact same approach too.

PR #443, opened 2016-09-22
#442: AbstractKubernetesResourceProvider now uses reflection to read …

iocanel:
…the @Named annotation so that it works with 3rd party annotations too (i.e. javax.inject.Named).

---
This change is [Reviewable](https://reviewable.io/reviews/arquillian/arquillian-cube/443)


lordofthejars:
Although I would do this as @MatousJobanek said, LGTM

iocanel:
Applied suggestions and squashed.

lordofthejars:
Thank you so much.

PR #446, opened 2016-09-25
port-forwarder improvements

spolti:


jwendell:
It seems GH doesn't allow review of/comment on a single commit :/

"Could you elaborate on which circumstances would it overflow?" refers to the commit "Keep the buffers equal to avoid overflow".

"I'm wondering if we could have some low defaults and expose these options in a builder so that users of this class could override them with higher values." refers to commit "Increase the default IO threads.".

spolti:
The overflow can occurs if the connection buffer has the buffer more than 8196 before it gets upgraded.

Regarding the buffer pool size, I believe it can be configurable, but this setup looks enough to most scenarios.

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/arquillian/arquillian-cube/446#-:-KSah0K2t9i3nIgV_QEx:b-ugtqrj)*

lordofthejars:
Everyone feels comfortable with this PR?

lordofthejars:
@spolti @jwendell the review process is still pending according to code-review. Is it really at this stage or we can merge it?

jwendell:
@lordofthejars I think @rcernich will take a look at them.

BTW, do you want we to fill an issue for each bug fixed in those commits (and amend commit messages saying the issue they are closing)?

lordofthejars:
@jwendell If you could do it, it would be perfect for our release notes.

Thanks

spolti:
Looks like #450 was merged, I'll rebase it and then update this PR or create a new one.

lordofthejars:
@spolti ok just ping me when ready. 

Btw are you interested in releasing a new alpha after this change?

spolti:
#460

spolti:
@lordofthejars feel free to merge it and realease a new beta version.

lordofthejars:
@spolti merged. Let me wait until we finish one PR regarding reports and I'll release a new alpha.

Issue #451, opened 2016-09-28
Add ability to map container name to pod name


rcernich:
Some existing test suites use container names that can't be manifest in certain environments (e.g. upper case letters in k8s).  Allowing users to manually specify the mapping between container name (arq) and container name (runtime) would alleviate this issue.

rcernich:
Added support for a "cubeId" property on the container configuration.  This should default to the container name if no cubeId is specified.

jwendell:
I suggest you guys to put something like "Closes #NNN", or "Fixes #NNN" into the commit messages so that github automatically closes the issues when the PR is merged :)

PR #452, opened 2016-09-28
Allow custom mapping between container ID and cube ID (#451)

rcernich:
This is now supported by adding a "cubeId" property to the container config, e.g:

<property name="cubeId">alternativeId</property>

### DIFF
--- a/core/src/main/java/org/arquillian/cube/impl/client/container/ContainerConfigurationController.java
+++ b/core/src/main/java/org/arquillian/cube/impl/client/container/ContainerConfigurationController.java
@@ -37,7 +37,7 @@ public class ContainerConfigurationController {
             return;
         }
 
-        Cube<?> cube = cubeRegistry.getCube(container.getName());
+        Cube<?> cube = cubeRegistry.getCube(ContainerUtil.getCubeIDForContainer(container));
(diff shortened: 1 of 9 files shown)


pom.xml
```diff
@@ -62,6 +62,8 @@
     </issueManagement>
 
     <properties>
+        <maven.compiler.source>1.8</maven.compiler.source>
```
lordofthejars: Can you change compiler version? Cube should be 1.7 compliant, not 1.8 for now :)

rcernich: Sure.  I think I was getting errors compiling with 1.7 jdk.  Let me test again...

rcernich: Not sure why I was getting an error.  Everything appears to be fine now.

rcernich:
I've removed those compiler settings from the commit.  That said, this seems to be causing problems with our tests, which I'm investigating atm.

lordofthejars:
@rcernich Do you prefer I merge this PR and if you find why this happens open a new PR or I just wait until you find why this happening.

rcernich:
This should be good to go now.  Thanks!

lordofthejars:
Thanks.

PR #455, opened 2016-09-29
Kubernetes cleanup

iocanel:
Most of these stuff were part mentioned in the review of the initial kubernetes pr.

The rest are misc cleaning up and preparation for aligning cube-kubenetes with cube-openshift.

### DIFF
--- a/kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/api/Configuration.java
+++ b/kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/api/Configuration.java
@@ -39,15 +39,18 @@ public interface Configuration {
 
     String ANSI_LOGGER_ENABLED = "ansi.logger.enabled";
 
-        // Non-config constants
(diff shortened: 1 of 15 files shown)


kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/api/Configuration.java
```diff
@@ -9,9 +9,16 @@
 
 public interface Configuration {
 
-    String KUBERNETES_MASTER = "kubernetes.master";
-    String KUBERNETES_NAMESPACE = "kubernetes.namespace";
-    String KUBERNETES_DOMAIN = "kubernetes.domain";
+    //Deprecated Property Names: {
+        String KUBERNETES_MASTER = "kubernetes.master";
```
lordofthejars: One question why these properties are deprecated? I mean don't we need kubernetes.master in any case?

iocanel: in favor of `master.url`, which I propose as the common new property name between cube-kubernetes and cube-openshift.

Issue #365, opened 2016-06-14
Add support for context in docker-compose v2


lordofthejars:
Add support for context in docker-compose v2 https://docs.docker.com/compose/compose-file/#build

lordofthejars:
Also implies resolving git repos https://docs.docker.com/compose/compose-file/#/context

PR #464, opened 2016-10-03
Add support for context in docker-compose v2

eddumelendez:
See gh-365

### DIFF
--- a/docker/docker/pom.xml
+++ b/docker/docker/pom.xml
@@ -97,6 +97,11 @@
             <artifactId>snakeyaml</artifactId>
         </dependency>
 
+		<dependency>
+			<groupId>org.eclipse.jgit</groupId>
+			<artifactId>org.eclipse.jgit</artifactId>
+		</dependency>
+
     </dependencies>
 
     <build>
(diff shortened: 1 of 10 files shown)


pom.xml
```diff
@@ -275,6 +276,12 @@
                 <version>${version.undertow}</version>
             </dependency>
 
+			<dependency>
+				<groupId>org.eclipse.jgit</groupId>
+				<artifactId>org.eclipse.jgit</artifactId>
+                <version>${version.jgit}</version>
```
lordofthejars: I think in version tag there is a mix between spaces and tabs.

docker/docker/pom.xml
```diff
@@ -97,6 +97,11 @@
             <artifactId>snakeyaml</artifactId>
         </dependency>
 
+		<dependency>
+			<groupId>org.eclipse.jgit</groupId>
+			<artifactId>org.eclipse.jgit</artifactId>
```
lordofthejars: I would add this dependency as provided and add in documentation that if you want to use `context` with git repository you need to add the jgit dependency. I said this because I think that almost all users will not use git feature, maybe `context` but without git. Then we are adding a dependency with all transitive dependencies to all users.

eddumelendez:
@lordofthejars I have added changes for this PR.

lordofthejars:
Thank you very much for your support and your work.

Issue #470, opened 2016-10-06
Couldn't create a service with mentioned networks in docker-compose.yml version 2


dipak-pawar:
docker-compose - 
```yml
version: '2'
services:
  pingpong:
    image: jonmorehouse/ping-pong
    ports:
        - "8080:8080"
    networks:
        - front-tier

networks:
  front-tier:
```

network inspect logs https://gist.github.com/dipak-pawar/b57c28ee999b02f7f438ec0198e28eac

It's creating container with default NetworkMode as `default` & networks as `bridge`
It should create a container with network mentioned in docker-compose.yml & start container.

dipak-pawar:
@lordofthejars I am working on this.

bartoszmajsak:
I assume it works using `docker-compose` cli?

bartoszmajsak:
If so label it as a bug, assign yourself and happy hacking! Thanks.

dipak-pawar:
@bartoszmajsak yes it works using docker-compose cli. I couldn't add label. Looks like need some permission

lordofthejars:
Thanks Dipak, good work finding the bug. No worries about label. If you
wamt to fix the bug go ahead. File that manges this is called
ComposeBuilder which is used in DockerComposeConverter

Alec

PR #472, opened 2016-10-12
 [issue #470] 

dipak-pawar:
This Fixes 
1.naming convention of docker-compose services & networks.
2.assigning network mentioned in services to container if mentioned otherwise default.


docker/docker/pom.xml
```diff
@@ -98,6 +98,11 @@
         </dependency>
 
         <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
```
bartoszmajsak: `<scope>test</scope>`?

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/DockerContainerDefinitionParser.java
```diff
@@ -69,20 +69,21 @@ private static DockerCompositions convert(URI uri, DefinitionFormat definitionFo
...
+            String fileContent = "";
             if(uri.isAbsolute()) {
                 content = IOUtil.asStringPreservingNewLines(uri.toURL().openStream());
             } else {
-                String fileContent = uri.toString();
+                fileContent = uri.toString();
                 content = IOUtil.asStringPreservingNewLines(new FileInputStream(fileContent));
             }
-            return convert(content, definitionFormat);
+            return convert(content, definitionFormat, fileContent);
```
bartoszmajsak: Why do we need both `content` and `fileContent`? Isn't it file location?

lordofthejars:
Change the PR title to show what is fixing. Now it is a simple number. You can use the constructor [issue #470] and the description.

docker/docker/src/test/java/org/arquillian/cube/docker/impl/docker/compose/DockerComposeConverterTest.java
```diff
@@ -21,212 +22,246 @@
 
 public class DockerComposeConverterTest {
```
dipak-pawar: yes I have run all tests belonging to this test class.

dipak-pawar:
@alexsoto, Tests was creating network but failing due change in network configuration. I have updated tests to new logic. Don't forget to do , mvn install as ftest-docker-compose-v2 is depends on snapshots in project.

lordofthejars:
@dipak-pawar why have you modified these tests? They are running correctly.

dipak-pawar:
@lordofthejars , Following are the reasons:
1.We couldn't access container name using arquillian-resource as actual container is tomcat & our service is running by name rootDir_tomcat.
2.changed port as container is running on custom network, So getting connection refused for 8081.

lordofthejars:
But still I don't understand the reason, in my computer using the master
branch before your change, everything works and a network called frontend
is created, the tests are executed without any problem.

Look:

docker network ls

NETWORK ID          NAME                DRIVER              SCOPE
1080531f9ec6        bridge              bridge              local
de8138d3ec7e        front-tier          bridge              local
80ff8d2201d8        host                host                local
9cbbc4ab1229        none                null                local

docker network inspect de81

[
    {
        "Name": "front-tier",
        "Id":
"de8138d3ec7e42d3f860796040273017e6a9b98269dc76c496a47285f91d1c18",
        "Scope": "local",
        "Driver": "bridge",
        "EnableIPv6": false,
        "IPAM": {
            "Driver": "default",
            "Options": null,
            "Config": [
                {
                    "Subnet": "172.18.0.0/16",
                    "Gateway": "172.18.0.1/16"
                }
            ]
        },
        "Internal": false,
        "Containers": {},
        "Options": {},
        "Labels": {}
    }
]

And after the test which passes correctly

docker network ls

NETWORK ID          NAME                DRIVER              SCOPE
1080531f9ec6        bridge              bridge              local
80ff8d2201d8        host                host                local
9cbbc4ab1229        none                null                local

So for this reason I don't understand this PR at all bcause it seems to work


El dj., 13 oct. 2016 a les 16:05, Dipak Pawar (<notifications@github.com>)
va escriure:

dipak-pawar:
@lordofthejars In master branch it's creating network but service is not using created network which is bug.
If you look at output of docker network inspect de81 it should include service container. currently its empty

lordofthejars:
ah true, but then your PR fixes this right? But I cannot see the relation
ship with the port modification you did. Why in frontend network cannot use
8081?

El dj., 13 oct. 2016 a les 16:29, Dipak Pawar (<notifications@github.com>)
va escriure:

dipak-pawar:
@lordofthejars need to figure out. I'll figure it out. No worries.

lordofthejars:
Ok, great. And Thanks for taking care of this, now I understand exactly what is happening, at first I thought the problem was that network was not created, but the problem is that container is not attached

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/DockerContainerDefinitionParser.java
```diff
@@ -69,20 +69,21 @@ private static DockerCompositions convert(URI uri, DefinitionFormat definitionFo
...
             if(uri.isAbsolute()) {
                 content = IOUtil.asStringPreservingNewLines(uri.toURL().openStream());
             } else {
-                String fileContent = uri.toString();
-                content = IOUtil.asStringPreservingNewLines(new FileInputStream(fileContent));
+                fileLocation = uri.toString();
+                content = IOUtil.asStringPreservingNewLines(new FileInputStream(fileLocation));
             }
-            return convert(content, definitionFormat);
+            return convert(content, definitionFormat, fileLocation);
```
dipak-pawar: @lordofthejars I didn't get what do you want to say?

bartoszmajsak:
Closed by #480

Issue #473, opened 2016-10-12
Add Support for multiple networks for a single service mentioned in docker-compose.yml in version 2


dipak-pawar:
version: '2'
services:
  tomcat:
    env_file: envs
    image: tutum/tomcat:7.0
    ports:
      - "8089:8089"
      - "8088:8088"
      - "8081:8080"
    networks:
      - front-tier
      - back-tier
networks:
  front-tier:
  back-tier.

Currently we can start service with single network. but not using all networks. provide support for multiple networks.

PR #480, opened 2016-10-14
[issue #473]Connects multiple network.

dipak-pawar:
Adds Logic to connect multiple networks.
1.It'll connect to single network & create container.
2.After creation of container container will connect to all networks mentioned in dockerContainers property.
3.after that container will start.


bartoszmajsak:
Closed by #480

PR #489, opened 2016-10-19
Network topology Implementation

dipak-pawar:
1. Adds network topology implementation.
2.Resolves null pointer exception for connect to network method.
2.Adds missing logic of networks in composebuilder.

### DIFF
--- a/docker/recorder/src/main/java/org/arquillian/cube/docker/impl/client/reporter/TakeDockerEnvironment.java
+++ b/docker/recorder/src/main/java/org/arquillian/cube/docker/impl/client/reporter/TakeDockerEnvironment.java
@@ -1,5 +1,7 @@
 package org.arquillian.cube.docker.impl.client.reporter;
 
+import com.github.dockerjava.api.command.InspectContainerResponse;
+import com.github.dockerjava.api.model.ContainerNetwork;
 import com.github.dockerjava.api.model.Statistics;
 import com.github.dockerjava.api.model.Version;
 import com.mxgraph.layout.hierarchical.mxHierarchicalLayout;
@@ -10,6 +12,7 @@ import org.arquillian.cube.docker.impl.client.CubeDockerConfiguration;
 import org.arquillian.cube.docker.impl.client.config.CubeContainer;
 import org.arquillian.cube.docker.impl.client.config.DockerCompositions;
 import org.arquillian.cube.docker.impl.client.config.Link;
+import org.arquillian.cube.docker.impl.client.config.Network;
 import org.arquillian.cube.docker.impl.docker.DockerClientExecutor;
 import org.arquillian.cube.spi.Cube;
 import org.arquillian.cube.spi.CubeRegistry;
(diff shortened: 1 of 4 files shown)


docker/recorder/src/main/java/org/arquillian/cube/docker/impl/client/reporter/TakeDockerEnvironment.java
```diff
@@ -186,6 +201,59 @@ private GroupEntry createDockerCompositionSchema(CubeDockerConfiguration cubeDoc
...
+            final Map<String, Object> insertedVertex = new HashMap<>();
+                for (Map.Entry<String, CubeContainer> container: containers.entrySet()) {
+                    final String containerId = container.getKey();
+                    Object containerName = graph.insertVertex(parent, null , containerId, 0, 0, 80, 30);
+                    final CubeContainer cubeContainer = container.getValue();
+                    Set<String> nwList = new HashSet<>();
+                    if (cubeContainer.getNetworkMode() != null) {
+                        nwList.add(cubeContainer.getNetworkMode());
+                    } else {
+                        InspectContainerResponse inspect = executor.inspectContainer(containerId);
```
lordofthejars: In case of not setting `networkMode` in `docker-compose`, would docker host assign automatically a networkMode? I ask it because you are querying docker host about this.

dipak-pawar: for docker-compose if you are not setting networkmode docker host assign networkmode as parentDir_default automatically.

lordofthejars: ok, so my question is what is the value in cube in case of not setting in our `docker-compose` file? If there is a value there it is because docker-host automatically assigns it, if not it will be null so this query is not required.

dipak-pawar: If you look at [this](https://github.com/dipak-pawar/arquillian-cube/blob/39bdd2d60bdffc00e083f71f55ab8b6e0c394f4a/docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/compose/ComposeBuilder.java#L50) we are assigning value of networkmode if there is no networks in docker-compose.yml.

but in case of CUBE format we are not setting it to cubecontainer, so networkmode is null for cubecontainer. but as there is no networkmode mentioned in containerdefinition, docker-host is assigning it as default. So we are getting it null as we are not setting it, docker host is setting it by default.
So to get default networkmode, we need to query this.

lordofthejars: right this is what I was wondering if docker host assign one automatically or not :) thanks.

lordofthejars:
I am going to merge this PR since the failure is not related with this logic.

Issue #486, opened 2016-10-18
Add cleanBuildImage attribute


lordofthejars:
Add cleanBuildImage attribute in docker module which by default is true. 

If this attribute is true, it removes any image that has been built during test process after the container instance is removed. If it is false then nothing is done.

This logic should resides in its own observer of `AfterDestroy` (https://github.com/arquillian/arquillian-cube/blob/17d1180b9c413dea72d9c64314937510d5bb48a4/spi/src/main/java/org/arquillian/cube/spi/event/lifecycle/AfterDestroy.java) event and should remove the image that was created for given `cubeId` only if the flag is true.

The method signature should look like:

`public void removeImage(@Observes AfterDestroy event, CubeRegistry cubeRegistry, CubeDockerConfiguration config) {}`

PR #492, opened 2016-10-20
[Issue #486] CleanBuildImage attribute support.

dipak-pawar:


docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfiguration.java
```diff
@@ -65,6 +67,7 @@
     private boolean removeVolumes = true;
     private AutoStartParser autoStartContainers = null;
     private DockerAutoStartOrder dockerAutoStartOrder = null;
+    private boolean cleanBuildImage = true;
```
bartoszmajsak: Cosmetic thing but I would move it next to `removeVolumes`. Somehow logically belongs there.

docker/docker/src/test/java/org/arquillian/cube/docker/impl/client/DockerImageControllerTest.java
```diff
@@ -0,0 +1,79 @@
...
+        bind(ApplicationScoped.class, CubeDockerConfiguration.class, dockerConfiguration);
+
+        final LocalCubeRegistry localCubeRegistry = new LocalCubeRegistry();
+        localCubeRegistry.addCube(cube);
+        bind(ApplicationScoped.class, CubeRegistry.class, localCubeRegistry);
+
+        bind(ApplicationScoped.class, DockerClientExecutor.class, executor);
+
+        fire(new AfterDestroy(CUBE_ID));
+        Mockito.verify(executor).removeImage("jonmorehouse/ping-pong", false);
```
bartoszmajsak: Isn't Mockito verifying it implicitly? @lordofthejars

bartoszmajsak: Another thing - what about integration test to interact with docker?

core/src/main/java/org/arquillian/cube/impl/client/CubeLifecycleController.java
```diff
@@ -31,7 +31,7 @@ public void changeToPreRunning(@Observes PreRunningCube event, CubeRegistry regi
         validateAndGet(registry, event.getCubeId()).changeToPreRunning();
     }
 
-    private Cube<?> validateAndGet(CubeRegistry registry, String cubeId) {
+    public static Cube<?> validateAndGet(CubeRegistry registry, String cubeId) {
```
bartoszmajsak: Why static? Maybe we should extract it to a validator class and inject in all the components where needed?

docker/docker/src/test/java/org/arquillian/cube/docker/impl/client/DockerImageControllerTest.java
lordofthejars: First thing,  Yes I think Mockito verifies automatically, but I always prefer to put it to help reader understand the test, if not for me it is more difficult to understand the test.

About integration test, remove image is just a call to docker-java function. We could create an integration test about this, but instead of doing an specific integration test for dockerexecutor, we usually depends on ftest, but in this specific case I am not sure if it has sense or not.

docker/docker/src/test/java/org/arquillian/cube/docker/impl/client/DockerImageControllerTest.java
```diff
@@ -0,0 +1,79 @@
...
+    @Before
+    public void before(){
+        when(cube.getId()).thenReturn(CUBE_ID);
+        when(cube.configuration()).thenReturn(config);
+        when(config.getImage()).thenReturn(new Image("jonmorehouse/ping-pong", null));
+    }
+
+    @Test
+    public void should_remove_docker_image() {
+        DockerClientExecutor executor = Mockito.mock(DockerClientExecutor.class);
```
lordofthejars: In this case image should not be removed because it is not built by cube.

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/DockerImageController.java
```diff
@@ -0,0 +1,34 @@
...
+            Cube cube = validateAndGet(registry, event.getCubeId());
+            CubeContainer config = (CubeContainer) cube.configuration();
+            String imageRef;
+            if (config.getImage() != null) {
+                imageRef = config.getImage().toImageRef();
+            } else {
+                imageRef = event.getCubeId() + ":latest";  // building image from Dockerfile.
+            }
+            DockerClientExecutor executor = dockerClientExecutorInstance.get();
+            executor.removeImage(imageRef, false);
```
lordofthejars: You need to remove the image if and only if `containerConfiguration.getBuildImage()` is not null. In other cases the image has not been built by cube so it should not be removed.

docs/configuration.adoc
```diff
@@ -71,6 +71,10 @@ Let's see valid attributes:
 |removeVolumes
 |Boolean to set if Cube should also remove the volumes associated with a container when removing the container. By default is true. Can be overwritten on container level.
 
+|cleanBuildImage
+|Boolean to set if Cube should also remove the images on which container is build after removing the container. By default is true. Can be overwritten on container level.
```
lordofthejars: No it cannot be overwritten on container level, if you set to true all built images are removed and if false no built images are removed.

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/DockerImageController.java
```diff
@@ -0,0 +1,34 @@
...
+
+
+public class DockerImageController {
+
+    @Inject
+    private Instance<DockerClientExecutor> dockerClientExecutorInstance;
+
+    public void removeImage(@Observes AfterDestroy event, CubeRegistry registry, CubeDockerConfiguration configuration) {
+        if (configuration.isCleanBuildImage()) {
+            Cube cube = validateAndGet(registry, event.getCubeId());
```
lordofthejars: You don't need  to do a validateAndGet, I would do directly a get and if cube == null then log a warning message and ignore that cube.

docker/docker/src/test/java/org/arquillian/cube/docker/impl/client/DockerImageControllerTest.java
dipak-pawar: To test removeImage as a part of integration tests we couldn't remove image which is used by container as it is used by container. It doesn't make sense to remove any random image from host. In this we are simply calling docker-java's removeImage. @bartoszmajsak If we have to test it we have to create new image & then remove it. I don't think it make sense to implement it here as docker-java already have same [test](https://github.com/docker-java/docker-java/blob/master/src/test/java/com/github/dockerjava/core/command/RemoveImageCmdImplTest.java#L55)

docker/docker/src/test/java/org/arquillian/cube/docker/impl/client/DockerImageControllerTest.java
```diff
@@ -0,0 +1,79 @@
...
+    CubeContainer config;
+
+    @Override
+    protected void addExtensions(List<Class<?>> extensions) {
+        extensions.add(DockerImageController.class);
+        super.addExtensions(extensions);
+    }
+
+    @Before
+    public void before(){
```
bartoszmajsak: I would suggest avoiding this kind of usage of `@Before`. I would rather prefer to duplicate mocking setup in each and every test, so when I read such a test I instantly know what is going on. With before I not only have to remember it, but also jump to it when I start my code reading exercise from the test method itself.

docker/docker/src/test/java/org/arquillian/cube/docker/impl/client/DockerImageControllerTest.java
bartoszmajsak: I'm not talking about testing `docker-java` bare bone functionality of removing image, but our flow of using things. 

> To test removeImage as a part of integration tests we couldn't remove image which is used by container as it is used by container. 

Sure, but you can always find a way to verify it. e.g. have one test doing this stuff and then next one which is executed `@InSequence` just verifies that the particular image is not there.

On the other hand if unit test with mocks gives us confidence... then I'm fine with it.

lordofthejars: @bartoszmajsak yes so what you mean is use same approach but using real DockerClientExecutor instead of the mocked one and then validate at the end that the built image is removed, or you think more like an ftest like module?

bartoszmajsak: Yes, more a functional test... but again - only if it gives us more confidence that what unit tests can give. I'm not sure, hence I opened the discussion

lordofthejars: For me, this is not complicated stuff so mocking would be enough, but it is my impression.

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/DockerImageController.java
```diff
@@ -0,0 +1,34 @@
...
+            CubeContainer config = (CubeContainer) cube.configuration();
```
lordofthejars: Surround with instanceof code, a cube might not return a CubeContainer all the time, for example in near future might return a PodConfiguration. Although probably you are not going to mix things, it might occur, and I prefer to avoid ClassCastExceptions.

Issue #483, opened 2016-10-17
Refactor Openshift extension on top of Kubernetes


jwendell:
Making it extends kubernetes avoids code duplication and reflects what in fact Openshift is: an enhanced Kubernetes.

lordofthejars:
Yes this is something that @iocanel wants to work as well i two steps first one:

https://github.com/arquillian/arquillian-cube/issues/445

And second one by aligning kubernetes objects to Cube interfaces too.

@iocanel @jwendell go ahead, if you need anything please feel free to ping me or @bartoszmajsak

iocanel:
If there are no objections I am on it.

jwendell:
@iocanel no problem. I had started to play with it, didn't get too far, but surprisingly that was not a huge refactoring as I was thinking. I managed to run a simple "hello world" test with a small set of changes.

Here is what I have done so far:
https://github.com/jwendell/arquillian-cube/commit/38d2bcc2c95907e92089ee25af7c1dfcfb920278

Feel free to use it, or some part of it, or discard it at all.

The next step I was about to do was to move the port forwarding code from openshit to kubernetes, as this is a kubernetes feature. If you want, I can do this separately and provide a PR with just this change.

iocanel:
@jwendell since you started, then please ho for it :-)

lordofthejars:
@jwendell happy to hear about that, when you finish this movement, we can start planning to do https://github.com/arquillian/arquillian-cube/issues/468 https://github.com/arquillian/arquillian-cube/issues/467 and https://github.com/arquillian/arquillian-cube/issues/465

jwendell:
for reference, my working can be tracked here: https://github.com/jwendell/arquillian-cube/commits/483
comments are welcome!

lordofthejars:
@jwendell @iocanel I have reviewed that there are several issues interrelated between them about integration of kubernetes/openshift (or even a PR https://github.com/arquillian/arquillian-cube/pull/493) how we can proceed about this?

iocanel:
My main issue is that currently there are tests that are broken in the openshift extension so I can't really do any kind of refactoring (I can't know if I broke something or not).

So we can either fix tests and refactor later, or ignore tests, refactor and fix later if needed.

iocanel:
@lordofthejars: fwiw i created #561 which for now seems to build / test with no issues.

lordofthejars:
I can close this issue since PR is merged isn't it?

Issue #445, opened 2016-09-22
Align the way kubernetes and openshift client is configured


iocanel:
So right now we have two modules that are configured with a completely different way.

They use different ways to pass configuration to the plugin and also configuration property names.
For example:

arquillian-cube-kubernetes is using `kubernetes.master` while arquillian-cube-kubernetes is using `originServer`. 

I think that we should have a common name to define common things across these modules. Maybe the arquillian guys can provide some details about possible naming conventions being used.

Next step is to provide a common set of properties (where applicable) that will work in parallel with the existing ones (which we will mark as deprecated).

Now regarding the possible ways of defining configuration the kubernetes module is using the ` system properties > environment variable > arquillian.xml` which I believe that should also be used in the openshift side.

lordofthejars:
I think that here 

> arquillian-cube-kubernetes is using kubernetes.master while arquillian-cube-kubernetes is using originServer.

you mean arquillian-cube-openshift. In theory in the future, after kubernetes part is fully integrated with cube model, openshift part should contain only things specific to openshift.

PR #493, opened 2016-10-20
WIP - Move PortForward implementation to kubernetes

jwendell:
It's not specific to Openshift, so, it doesn't have to live there.
# 

This is being done as part of #483.
I think this deserves being a separated commit because it's huge.

Tagged as WIP because it's probably better not merge it right now, perhaps it makes sense as part of #483. However it would be good to have an overall feedback or even an early review.

Tests are working.

A required change for an existing user of this feature is to move the definition in arquillian.xml:
 property "proxiedContainerPorts" from "openshift" qualifier
to
 property "pod.forward.ports" in "kubernetes" qualifier

(thought of a better name for "proxiedContainerPorts", following kubernetes convention of having dots, but this of course can be changed)


jwendell:
cc @rcernich

rcernich:
@jwendell, you might want to verify that having dots in the property name still allows those properties to be configured through arquillian.xml.

iocanel:
I think that we can also have configuration objects extend one an other so that we don't have to worry if the user adds config to "kubernetes" or "openshift".

jwendell:
Indeed, but this is a thing for #483 or even #445 :)

iocanel:
Definitely!

jwendell:
@iocanel so, do you have any opinion on *this* patch?

iocanel:
I am not 100% sure but I think that there is a logical gap.

I am not very familiar with the openshift extension, so please do correct me if I am wrong :-)

I think that the openshift plugin was mostly targeting tests that involve a single pod, and for that case it made sense to define the ports inside the configuration.

The kubernetes extension, is taking a more `system testing` approach, where it may spin up multiple pods. So defining the ports inside the configuration file, doesn't help so much here (can't tell which port is for which pod etc). I guess that it doesn't harm either, but we need something more flexible here, but I think that we need something more flexible.

Also in most cases, the pod name is not known in advance, which breaks the syntax (pod:source:target). And hardcoding that local port, also seems problematic.

So I think that:

- The local port needs to be somehow injected to the test, rather than preconfigured.
- The pod needs to be selected (labels? random pod of service X? first pod of service X).

Here's a possible approach:
https://github.com/iocanel/arquillian-cube/commit/1c89c49e9ed713512a12d690c13d83f7add86618

jwendell:
I think we should not make assumptions on how openshift or kubernetes extensions must be used. Openshift is a layer on top of kubernetes. So, whatever can be done with Kubernetes can also be done with Openshift. The opposite is not true, by definition.

In our use case, we are using Port Forward (which is a kubernetes feature, therefore is also available on openshift) to deploy a JBoss EAP server into a remote container. For instance, see this [example](https://github.com/jboss-openshift/ce-testsuite/blob/master/eap/eap64/src/test/resources/arquillian.xml#L57).

So, arquillian will make a deployment at localhost:9990, but this port is mapped (proxied) to the remote container (that could be a kubernetes or openshift), via Port Forward code. This definition is on the same file, within the proper [extension qualifier](https://github.com/jboss-openshift/ce-testsuite/blob/master/eap/eap64/src/test/resources/arquillian.xml#L44).

So, that's how it works for us.

This patch just move things from openshift extension to kubernetes one.

iocanel:
I don't question that this will work, I am just saying that the `kubernetes` extension covers some `more generic` use cases and it would be nice to cover those too.

jwendell:
So, you propose the `@PortForward` annotation in addition of current state (having port mapping done in arquillian.xml)?

Also, note that port forward is not tied to url's or web services. A MySQL server pod (inside kubernetes or openshift) can have its port (usually 3306) exposed and we can make use of that to forward it locally so that a MySQL client can connect to localhost:3306 and in fact access the remote MySQL server.

iocanel:
I suggested `@PortForward` as a more flexible way of doing port mapping, but I don't mind if its going to be added in addition or as a replacement of what we currently have.

jwendell:
`@PortForward` would work for tests that somehow uses those forwarded ports.

In our use case we need the ports to be forwarded as soon as the container is up, so that arquillian can deploy the tests/artifacts there.

So, the workflow would be:
 1. Start pods defined in kubernetes.json or similar
 1. Start forwarding ports (localhost:9990 => container:9990, etc)
 1. Arquillian deploys tests/artifacts in "local" EAP, using localhost:9990
 1. Start running the tests

iocanel:
I guess that we need both approaches.

iocanel:
@jwendell: I think that we can push this and further improve that in follow up PRs. wdyt?

jwendell:
@iocanel, I'm talking to @rcernich , it seems this isn't the way to go...

This port forward feature doesn't make sense for the kubernetes extension use case (system testing).

So, I believe we should close this for now, and proceed with the refactoring of the openshift extension on top of kubernetes without this change...

Rob has some thoughts about how could that be done, but I believe this conversation belongs to #483.

iocanel:
Port forwarding as a concept does make sense for system testing too. I could say its *required* in many cases.

What does not fit very well, to system tests, is the way one defines the port forwarding rules. But this is something we can add later on (e.g. bring in the @PortForward annotation).Then users can decide which approach serves them best.

lordofthejars:
@iocanel @jwendell @rcernich how is going this? Now I see some conflicts in this PR.

jwendell:
This is "on hold" as I have other priorities ATM.

Talking with @rcernich last week, this question was raised: Why not having this port forward feature implemented directly in kubernetes-client? Makes sense, right?

iocanel:
@jwendell: I've already raised an issue in the kubernetes-client project. The main issue is that kubernetes-client is using okhttp and its a little bit tricky to implement using okhttp.

lordofthejars:
I was just asking to know the status :) For what I see and since portforwarding is something that affects also kubernetes, yes it has sense to put there too.

lordofthejars:
What is going to happen with this PR? It is already done in https://github.com/arquillian/arquillian-cube/pull/561

iocanel:
This has been implemented for a while now. We can close this one.

Issue #498, opened 2016-10-22
should_report_network_topology_of_docker_containers fails


iocanel:
https://github.com/arquillian/arquillian-cube/blob/f8b0ac239314976bb5c84d88ff8dbc8888a823c2/docker/recorder/src/test/java/org/arquillian/cube/docker/impl/client/reporter/TakeDockerEnvironmentTest.java#L284

It seems that no behavior is defined for this test case.

lordofthejars:
What you mean by no behaviour, there is some assertions that asserts that the information is correctly created.

iocanel:
I mean that the mock doesn't seem to define any behavior for the inspect method, which is indirectly called (resulting in an NPE).

lordofthejars:
No, we are using the object just for capturing the result inside. https://github.com/arquillian/arquillian-cube/blob/f8b0ac239314976bb5c84d88ff8dbc8888a823c2/docker/recorder/src/test/java/org/arquillian/cube/docker/impl/client/reporter/TakeDockerEnvironmentTest.java#L293

iocanel:
Does the test work for you?

dipak-pawar:
@iocanel test is not working. Let me check what's wrong with it.

dipak-pawar:
Submitted [PR](https://github.com/arquillian/arquillian-cube/pull/501) for above issue. Thanks @iocanel

iocanel:
You may have to use something like the `@RequiresDocker` annotation so that the test will not fail on environments where docker is not available (e.g. circleci).

dipak-pawar:
@iocanel there is bug in test code. No need to mention @RequiresDocker annotation for this tests as  we are providing networkMode from dockerContainers [properties](https://github.com/arquillian/arquillian-cube/blob/master/docker/recorder/src/test/java/org/arquillian/cube/docker/impl/client/reporter/TakeDockerEnvironmentTest.java#L48), and you will never reach [this](https://github.com/arquillian/arquillian-cube/blob/master/docker/recorder/src/main/java/org/arquillian/cube/docker/impl/client/reporter/TakeDockerEnvironment.java#L223) scenario.

lordofthejars:
@iocanel fixed and merged :)

PR #500, opened 2016-10-24
Fix [Issue #498]

dipak-pawar:


dipak-pawar:
Added fix in https://github.com/arquillian/arquillian-cube/pull/501

PR #503, opened 2016-10-26
[issue #499] Copy waitforit script to /tmp and fallback strategy

### LINKED ISSUE #499: Await polling strategy should not make assumptions about the container environment
I've been experimenting with cube and WildFly. Many of the examples I found reference the 'polling' await strategy. I encountered some problems with this:

Neither `ss` or `ping` is guaranteed to be available in the running container. Indeed, in the official WildFly base image, `ss` is not installed.

When using the 'wait-for-it' option, cube copies the script into '/'. There's no guarantee that the running container user has access to this location. Also, the `chmod` that [applies execute permissions](https://github.com/arquillian/arquillian-cube/blob/308287ec56a091d473309f21dcd3dff9f6a4fdf5/docker/docker/src/main/java/org/arquillian/cube/docker/impl/await/PollingAwaitStrategy.java#L137) to the script, incorrectly assumes that the `WORKDIR` is '/', which is not the case with the WildFly image (or the majority of JBoss base images).

lordofthejars:
This PR fixes #499 bug and improves how ping works by:

1. Copying Wait-For-It script from classpath location and not downloading from internet, so in cases of not having internet connection or for future customizations we don't relay on external site
2. Copy `wait-for-it` to `/tmp` to avoid any workspace problem inside docker containers.
3. Implementing fallback strategy so if _ss_ is not found in the container it automatically changes ping strategy to _wait-for-it_.

### DIFF
--- a/docker/docker/src/main/java/org/arquillian/cube/docker/impl/await/PollingAwaitStrategy.java
+++ b/docker/docker/src/main/java/org/arquillian/cube/docker/impl/await/PollingAwaitStrategy.java
@@ -1,6 +1,10 @@
 package org.arquillian.cube.docker.impl.await;
 
 import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
@@ -9,27 +13,23 @@ import java.util.logging.Logger;
 
 import org.arquillian.cube.docker.impl.client.config.Await;
 import org.arquillian.cube.docker.impl.docker.DockerClientExecutor;
-import org.arquillian.cube.docker.impl.util.HomeResolverUtil;
 import org.arquillian.cube.docker.impl.util.Ping;
 import org.arquillian.cube.impl.util.IOUtil;
 import org.arquillian.cube.spi.Cube;
 import org.arquillian.cube.spi.metadata.HasPortBindings;
 import org.arquillian.cube.spi.metadata.HasPortBindings.PortAddress;
-import org.arquillian.spacelift.Spacelift;
-import org.arquillian.spacelift.task.net.DownloadTool;
 
 public class PollingAwaitStrategy extends SleepingAwaitStrategyBase {
 
     private static final Logger log = Logger.getLogger(PollingAwaitStrategy.class.getName());
     private static final String MESSAGE = "Service is Up";
-    private static final String WAIT_FOR_IT_SCRIPT_DOWNLOAD_LOCATION = "https://raw.githubusercontent.com/vishnubob/wait-for-it/master/wait-for-it.sh";
-    private static final String WAIT_FOR_IT_SCRIPT_DIRECTORY = HomeResolverUtil.resolveHomeDirectoryChar("~/.arquillian");
-    private static final String WAIT_FOR_IT_SCRIPT = WAIT_FOR_IT_SCRIPT_DIRECTORY + "/wait-for-it.sh";
+    private static final String WAIT_FOR_IT_SCRIPT = "wait-for-it.sh";
 
     public static final String TAG = "polling";
 
     private static final int DEFAULT_POLL_ITERATIONS = 10;
     private static final String DEFAULT_POLL_TYPE = "sscommand";
+    public static final String CONTAINER_DIRECTORY = "/tmp";
 
     private int pollIterations = DEFAULT_POLL_ITERATIONS;
     private String type = DEFAULT_POLL_TYPE;
@@ -38,9 +38,12 @@ public class PollingAwaitStrategy extends SleepingAwaitStrategyBase {
     private Cube<?> cube;
     private List<Integer> ports = null;
 
+    // To avoid having to copy the script for all ports, state is saved.
+    private boolean alreadyCopiedWaitForIt = false;
+
     public PollingAwaitStrategy(Cube<?> cube, DockerClientExecutor dockerClientExecutor, Await params) {
         super(params.getSleepPollingTime());
-        
+
         this.cube = cube;
         this.dockerClientExecutor = dockerClientExecutor;
 
@@ -97,9 +100,16 @@ public class PollingAwaitStrategy extends SleepingAwaitStrategyBase {
 
                 break;
                 case "sscommand": {
-                    if(!Ping.ping(dockerClientExecutor, cube.getId(), resolveCommand("ss", port),
-                            this.pollIterations, this.getSleepTime(), this.getTimeUnit())) {
-                        return false;
+                    try {
+                        if (!Ping.ping(dockerClientExecutor, cube.getId(), resolveCommand("ss", port),
+                                this.pollIterations, this.getSleepTime(), this.getTimeUnit())) {
+                            return false;
+                        }
+                    } catch (UnsupportedOperationException e) {
+                        // In case of not having ss command installed on container, it automatically fall back to waitforit approach
+                        if (! executeWaitForIt(portBindings.getInternalIP(), port)) {
+                            return false;
+                        }
                     }
                 }
                 break;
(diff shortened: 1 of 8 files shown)


lordofthejars:
@jamesnetherton I have implemented this PR for fixing #499 feel free to review it :)

jamesnetherton:
I gave the changes a quick test, but they didn't work for me.

I did not see the `UnsupportedOperationException` get thrown, so the fallback to 'wait-for-it' never happened. `ss` is definately not installed in my image because I see `sh: ss: command not found` in the build output. 

On reflection, I think copying the wait-for-it script to `/tmp` is not enough. The problem is with which user owns the file and which user performs the `chmod` operation. E.g for the WildFly base image:

```
[jboss@1f673467be14 ~]$ whoami
jboss
[jboss@1f673467be14 ~]$ ls -l /tmp/
total 8
drwxr-xr-x. 2 jboss jboss   17 Oct 26 08:41 hsperfdata_jboss
drwxr-xr-x. 2 root  root    16 Sep  7 01:17 hsperfdata_root
-rwx------. 1 root  root   827 Sep  6 14:02 ks-script-VNahoJ
-rw-r--r--. 1 root  root  3657 Oct 26 08:41 wait-for-it.sh
```

Since the script is owned by root, when the `chmod` happens (presumably run as the jboss user) we get:

```
chmod: changing permissions of '/tmp/wait-for-it.sh': Operation not permitted
sh: /tmp/wait-for-it.sh: Permission denied
```

lordofthejars:
This is going to be harder than expected :)

jamesnetherton:
Out of interest - is there any specific reason why we do these port availability checks inside of the container?

Why not just have some Java logic to poll the exposed ports from the host? Then you can remove the dependency on shell scripts and other binaries.

lordofthejars:
You can still do what you mention, but doing inside the container is faster in terms of performance. Maybe not when starting one container but yes when several containers are started in parallel.

lordofthejars:
For example you can use health check await strategy.

lordofthejars:
@jamesnetherton I think now it should work perfectly :)

jamesnetherton:
Much better, thanks!

docker/docker/src/main/java/org/arquillian/cube/docker/impl/await/PollingAwaitStrategy.java
```diff
@@ -78,14 +86,14 @@ public boolean await() {
...
 
         Collection<Integer> pingPorts = this.ports;
-        if(ports == null) {
+        if (ports == null) {
             pingPorts = portBindings.getBoundPorts();
         }
         for (Integer port : pingPorts) {
-            switch(this.type) {
+            switch (this.type) {
                 case "ping": {
```
bartoszmajsak: Instead of `string` can we think about using strategy pattern here? Would that be beneficial? Maybe also will open it up  for SPI?

docker/docker/src/main/resources/org/arquillian/cube/docker/impl/await/wait-for-it.sh
```diff
@@ -0,0 +1,161 @@
+#!/usr/bin/env bash
```
bartoszmajsak: I guess we should add license header here?

docker/docker/src/main/resources/org/arquillian/cube/docker/impl/await/wait-for-it-sh.sh
```diff
@@ -0,0 +1,165 @@
+#!/usr/bin/env sh
```
bartoszmajsak: What is the difference between those two scripts?

docker/docker/src/main/java/org/arquillian/cube/docker/impl/await/PollingAwaitStrategy.java
```diff
@@ -116,40 +131,66 @@ public boolean await() {
...
-        final String[] commands = {"sh", "-c", command};
-        String result = dockerClientExecutor.execStart(cube.getId(), commands);
-        return result != null && result.trim().contains(MESSAGE);
+
+    }
+
+    private Path copyWaitForItScriptToTempDir() throws IOException {
+        final Path arquilliancube = Files.createTempDirectory("arquilliancube");
+        final Path waitForItLocation = arquilliancube.resolve(Paths.get(WAIT_FOR_IT_SCRIPT));
+        Files.copy(PollingAwaitStrategy.class.getResourceAsStream("/org/arquillian/cube/docker/impl/await/wait-for-it.sh"), waitForItLocation);
```
bartoszmajsak: You mention above we are copying our `-sh.sh` but here I see it's the old one.

docker/docker/src/main/java/org/arquillian/cube/docker/impl/await/PollingAwaitStrategy.java
```diff
@@ -116,40 +131,66 @@ public boolean await() {
     }
 
     private boolean executeWaitForIt(String containerIp, int port) {
-        if (! new File(WAIT_FOR_IT_SCRIPT_DIRECTORY).exists()) {
-            if (! new File(WAIT_FOR_IT_SCRIPT_DIRECTORY).mkdirs()) {
-                throw new IllegalArgumentException("Couldn't create the Arquillian directory at ~/.arquillian");
+
+        // We copy our wait-for-it-sh.sh file form classpath
+        // Our wait-for-it-sh.sh script also works with busybox/alpine which is not true with the official one.
```
bartoszmajsak: What's the difference? Can we contribute back?

docker/docker/src/main/java/org/arquillian/cube/docker/impl/await/PollingAwaitStrategy.java
lordofthejars: Currently it is an "SPI" I mean you can write your custom await strategy following an interface. This started as a simple await method that has been grown from time to time.

docker/docker/src/main/java/org/arquillian/cube/docker/impl/await/PollingAwaitStrategy.java
lordofthejars: In theory not, since the project is not maintained anymore according to comments and PR in queue.

docker/docker/src/main/java/org/arquillian/cube/docker/impl/await/PollingAwaitStrategy.java
lordofthejars: One uses bash and the other one sh POSIX. The problem is that POSIX version does not work yet. So it is there as my initial work but I need to spend some more time on this.

docker/docker/src/main/resources/org/arquillian/cube/docker/impl/await/wait-for-it-sh.sh
lordofthejars: one is bash other POSIX

lordofthejars:
My plan is to merge this, since it fixes the problem mentioned, and then open a new issue to fix the POSIX script.

docker/docker/src/main/java/org/arquillian/cube/docker/impl/await/PollingAwaitStrategy.java
lordofthejars: Currently the old one is the one that works. IntelliJ has renamed the previous comment automatically :)

bartoszmajsak: Damn you IDEs :) Or damn you comments :D

docker/docker/src/main/java/org/arquillian/cube/docker/impl/await/PollingAwaitStrategy.java
bartoszmajsak: The guy is looking for help, but it seems to be a bit maintained.

docker/docker/src/main/java/org/arquillian/cube/docker/impl/await/PollingAwaitStrategy.java
bartoszmajsak: Yeah, but somehow we limit ourselves here with a `switch` over limited set of "types"

docker/docker/src/main/java/org/arquillian/cube/docker/impl/await/PollingAwaitStrategy.java
lordofthejars: well see the PR, people ask why it is not merged, and yes he is asking for help, but this does not means it is maintained.

docker/docker/src/main/java/org/arquillian/cube/docker/impl/await/PollingAwaitStrategy.java
lordofthejars: we can open an issue to improve this

Issue #502, opened 2016-10-24
Unable to use log await strategy


jamesnetherton:
I've been trying to get the log await strategy working, so far with no success. Tested against WildFly and Tomcat images.

When debugging, I never see [onNext()](https://github.com/arquillian/arquillian-cube/blob/308287ec56a091d473309f21dcd3dff9f6a4fdf5/docker/docker/src/main/java/org/arquillian/cube/docker/impl/await/LogScanningAwaitStrategy.java#L127-L135) in `LogScanningAwaitStrategy` get called. So no match logic ever seems to get done against the container logs, even though the container is successfully started.

Some details about my environment - I run Fedora 24 with the Fedora Docker package installed (not the official Docker RPM):

```
$ docker version
Client:
 Version:         1.10.3
 API version:     1.22
 Package version: docker-1.10.3-53.gite03ddb8.fc24.x86_64
 Go version:      go1.6.3
 Git commit:      e03ddb8/1.10.3
 Built:           
 OS/Arch:         linux/amd64

Server:
 Version:         1.10.3
 API version:     1.22
 Package version: docker-1.10.3-53.gite03ddb8.fc24.x86_64
 Go version:      go1.6.3
 Git commit:      e03ddb8/1.10.3
 Built:           
 OS/Arch:         linux/amd64
```

Here's an example arquillian.xml config for Tomcat:

```
<extension qualifier="docker">
    <property name="definitionFormat">CUBE</property>
    <property name="dockerContainers">
        tomcat:
            image: tutum/tomcat:7.0
            exposedPorts: [8089/tcp]
            await:
                strategy: log
                match: Server startup
                stdOut: true
                stdErr: true
                sleepPollingTime: 10 s
                iterations: 6
            env: [TOMCAT_PASS=mypass, JAVA_OPTS=-Dcom.sun.management.jmxremote.port=8089 -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.authenticate=false]
            portBindings: [8089/tcp, 8080/tcp]
    </property>
</extension>

<container qualifier="tomcat" default="true">
    <configuration>
        <property name="host">localhost</property>
        <property name="httpPort">8080</property>
        <property name="user">admin</property>
        <property name="pass">mypass</property>
    </configuration>
</container>
```

lordofthejars:
Thank you very much for reporting this, tomorrow morning I will take a look.

PR #504, opened 2016-10-26
[issue #502] Log Strategy works with countdownlatch instead of polling.

lordofthejars:

Fixes log await strategy issue #502 

- Change `LogScanningAwaitStrategy` to use an asynchronous way instead of polling. Basically when the container log contains the expected value, it unblocks the thread. With this approach we can add inspect log line by line instead as a batch, which causes some problems in some containers.

**Fixes**: #502

### DIFF
--- a/docker/docker/src/main/java/org/arquillian/cube/docker/impl/await/LogScanningAwaitStrategy.java
+++ b/docker/docker/src/main/java/org/arquillian/cube/docker/impl/await/LogScanningAwaitStrategy.java
@@ -10,16 +10,18 @@ import org.arquillian.cube.spi.Cube;
 import java.io.IOException;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import java.util.regex.Pattern;
 
 public class LogScanningAwaitStrategy extends SleepingAwaitStrategyBase {
 
     public static final String TAG = "log";
 
+    private static final Logger logger = Logger.getLogger(LogScanningAwaitStrategy.class.getName());
     private static final String REGEXP_PREFIX = "regexp:";
-    private static final int DEFAULT_TIME_OUT = 15;
 
-    private int timeout = DEFAULT_TIME_OUT;
+    private int timeout = 15;
 
     private boolean stdOut;
     private boolean stdErr;
@@ -69,7 +71,6 @@ public class LogScanningAwaitStrategy extends SleepingAwaitStrategyBase {
         final CountDownLatch containerUp = new CountDownLatch(1);
 
         try (final LogContainerResultCallback callback = new LogContainerResultCallback(containerUp)) {
-            try {
 
                 client.logContainerCmd(cube.getId())
                         .withStdErr(true)
@@ -77,13 +78,9 @@ public class LogScanningAwaitStrategy extends SleepingAwaitStrategyBase {
                         .withFollowStream(true)
                         .exec(callback);
 
-                boolean result = containerUp.await(this.timeout, TimeUnit.SECONDS);
-                return result;
-            } catch (InterruptedException e) {
-                return false;
-            }
-        } catch (IOException e) {
-
+                return containerUp.await(this.timeout, TimeUnit.SECONDS);
+        } catch (IOException | InterruptedException e) {
+            logger.log(Level.SEVERE, String.format("Log Await Strategy failed with %s", e.getMessage()));
             return false;
         }
 
--- a/docker/docker/src/main/java/org/arquillian/cube/docker/impl/await/LogScanningAwaitStrategy.java
+++ b/docker/docker/src/main/java/org/arquillian/cube/docker/impl/await/LogScanningAwaitStrategy.java
@@ -1,52 +1,41 @@
 package org.arquillian.cube.docker.impl.await;
 
-import java.text.DateFormat;
-import java.text.ParseException;
-import java.text.SimpleDateFormat;
-import java.util.regex.Pattern;
-
-import org.arquillian.cube.docker.impl.client.config.Await;
-import org.arquillian.cube.docker.impl.docker.DockerClientExecutor;
-import org.arquillian.cube.docker.impl.util.Ping;
-import org.arquillian.cube.docker.impl.util.PingCommand;
-import org.arquillian.cube.spi.Cube;
-
 import com.github.dockerjava.api.DockerClient;
 import com.github.dockerjava.api.model.Frame;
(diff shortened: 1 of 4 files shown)


docker/docker/src/main/java/org/arquillian/cube/docker/impl/await/LogScanningAwaitStrategy.java
```diff
@@ -1,52 +1,41 @@
...
+
     private static final String REGEXP_PREFIX = "regexp:";
-    
-    private static final int DEFAULT_POLL_ITERATIONS = 10;
-    
-    private int pollIterations = DEFAULT_POLL_ITERATIONS;
-    
+    private static final int DEFAULT_TIME_OUT = 15;
+
+    private int timeout = DEFAULT_TIME_OUT;
```
bartoszmajsak: If we use it only once, there is no need for constant `DEFAULT_TIME_OUT`.

docker/docker/src/main/java/org/arquillian/cube/docker/impl/await/LogScanningAwaitStrategy.java
```diff
@@ -72,78 +66,57 @@ public boolean isStdErr() {
...
+                client.logContainerCmd(cube.getId())
+                        .withStdErr(true)
+                        .withStdOut(true)
+                        .withFollowStream(true)
+                        .exec(callback);
+
+                boolean result = containerUp.await(this.timeout, TimeUnit.SECONDS);
+                return result;
+            } catch (InterruptedException e) {
+                return false;
```
bartoszmajsak: Some logging here?

docker/docker/src/main/java/org/arquillian/cube/docker/impl/await/LogScanningAwaitStrategy.java
```diff
@@ -72,78 +66,57 @@ public boolean isStdErr() {
...
+            try {
+
+                client.logContainerCmd(cube.getId())
+                        .withStdErr(true)
+                        .withStdOut(true)
+                        .withFollowStream(true)
+                        .exec(callback);
+
+                boolean result = containerUp.await(this.timeout, TimeUnit.SECONDS);
+                return result;
```
bartoszmajsak: I think we can inline this return, can't we?

docker/docker/src/main/java/org/arquillian/cube/docker/impl/await/LogScanningAwaitStrategy.java
```diff
@@ -72,78 +66,57 @@ public boolean isStdErr() {
...
-    }
-    
-    private int parseTimestamp(String s) {
-        try {
-            return (int) dateFormat.parse(s.substring(0, s.indexOf('.'))).getTime() / 1000;
-        } catch (ParseException e) {
-            throw new IllegalArgumentException("Timestamp parse failure: " + s, e);
+        } catch (IOException e) {
+
+            return false;
```
bartoszmajsak: Multicatch?

docker/docker/src/test/java/org/arquillian/cube/docker/impl/await/AwaitStrategyTest.java
```diff
@@ -286,9 +286,7 @@ public void should_create_log_scanning_await_strategy() {
         AwaitStrategy strategy = AwaitStrategyFactory.create(null, cube, cubeContainer);
 
         assertThat(strategy, instanceOf(LogScanningAwaitStrategy.class));
-        assertThat(((LogScanningAwaitStrategy)strategy).getSleepTime(), is(500));
-        assertThat(((LogScanningAwaitStrategy)strategy).getTimeUnit(), is(TimeUnit.MILLISECONDS));
-        assertThat(((LogScanningAwaitStrategy)strategy).getPollIterations(), is(10));
+        assertThat(((LogScanningAwaitStrategy)strategy).getTimeout(), is(15));
         assertThat(((LogScanningAwaitStrategy)strategy).isStdOut(), is(true));
         assertThat(((LogScanningAwaitStrategy)strategy).isStdErr(), is(false));
```
bartoszmajsak: This looks really ugly. Can you rework those assertions?

docker/docker/src/test/java/org/arquillian/cube/docker/impl/await/AwaitStrategyTest.java
```diff
@@ -301,18 +299,15 @@ public void should_create_log_scanning_await_strategy_without_defaults() {
...
 
         AwaitStrategy strategy = AwaitStrategyFactory.create(null, cube, cubeContainer);
 
         assertThat(strategy, instanceOf(LogScanningAwaitStrategy.class));
-        assertThat(((LogScanningAwaitStrategy)strategy).getSleepTime(), is(2));
-        assertThat(((LogScanningAwaitStrategy)strategy).getTimeUnit(), is(TimeUnit.SECONDS));
-        assertThat(((LogScanningAwaitStrategy)strategy).getPollIterations(), is(30));
+        assertThat(((LogScanningAwaitStrategy)strategy).getTimeout(), is(20));
         assertThat(((LogScanningAwaitStrategy)strategy).isStdOut(), is(false));
         assertThat(((LogScanningAwaitStrategy)strategy).isStdErr(), is(true));
```
bartoszmajsak: This looks really ugly. Can you rework those assertions?

lordofthejars:
The problem is that a test for this would require an ftest, we can wrote one. Notice that a mocking strategy in this case would be really hard to implement because of the callback.

docker/docker/src/main/java/org/arquillian/cube/docker/impl/await/LogScanningAwaitStrategy.java
lordofthejars: I put in different statements because I wanted to add something between but now I cannot remember what :) I fox for now

docker/docker/src/main/java/org/arquillian/cube/docker/impl/await/LogScanningAwaitStrategy.java
lordofthejars: I totally forget although I though myself of doing it when writing the line 💃 .

bartoszmajsak:
Maybe some exercise for you with mocking? :) I'm curious if we could end up with some meaningful test.

lordofthejars:
Changes done, I cannot see how to write a test using mocking since the countdownlatch is inside a private class, and this private class should remain private and it is created inside the method. If you see something at `LogScanningAwaitStrategy.await()` to write a test, it is welcomed but I found hard to see how. What we can do is rewrite an ftest to use it.

Issue #506, opened 2016-10-27
Display reporting parameters like IO, network, disk usage into a table form


dipak-pawar:
##### Expected Behaviour
Display reporting parameters like IO, network, disk usage into a table form

##### Current Behaviour
Reporting parameters like IO, network, disk usage are displayed into key value entry.

PR #507, opened 2016-10-27
[Issue #506] Put reporting parameters like IO, network, disk usage into a table form

dipak-pawar:
Putting reporting parameters like IO, network, disk usage into a table form.
e.g.
 | rx_bytes  |
 | before | After | use |
 | 10 | 12 | 2 |
-  Reporting statistics only after method with format shown above.
- Some cleanup of remoteAPI, serverURI from ftest module.

**Fixes**: #506

### DIFF
--- a/docker/ftest/src/test/java/org/arquillian/cube/servlet/HelloWorldServletTest.java
+++ b/docker/ftest/src/test/java/org/arquillian/cube/servlet/HelloWorldServletTest.java
@@ -87,15 +87,6 @@ public class HelloWorldServletTest {
     }
 
     @Test
-    public void should_copy_files_from_container(@ArquillianResource CubeController cubeController, @ArquillianResource CubeID cubeID) throws IOException {
-        File newFolder = folder.newFolder();
-        cubeController.copyFileDirectoryFromContainer(cubeID, "/tomcat/logs", newFolder.getAbsolutePath());
-        File logFolder = newFolder.listFiles()[0];
-        assertThat(logFolder, notNullValue());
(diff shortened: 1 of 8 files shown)


docker/ftest/src/test/java/org/arquillian/cube/servlet/HelloWorldServletTest.java
```diff
@@ -87,15 +87,6 @@ public void should_get_changes_on_container(@ArquillianResource CubeController c
     }
 
     @Test
-    public void should_copy_files_from_container(@ArquillianResource CubeController cubeController, @ArquillianResource CubeID cubeID) throws IOException {
```
lordofthejars: Why this tests are removed?

dipak-pawar: Currently ftest is using 1.23 docker remote api. If look at [copyFileFromContainerCmd method](https://github.com/docker-java/docker-java/blob/master/src/main/java/com/github/dockerjava/api/DockerClient.java#L176) which is deprecated from version 1.20. So there is no use of this test for this new version.

this test was using remote docker api 1.12 previously.

lordofthejars:
But test passes right? Or it fails?

dipak-pawar:
Nopes. It's failing

lordofthejars:
Ok, no worries for now, but next time if you can I recommend to put different things in different PR so it is easier to for example revert in case of an error.

Issue #508, opened 2016-10-27
Make the port forwarder bind address configurable


spolti:
##### Issue Overview
Actually arquillian-cube does not provide a way to configure the port forwarder bind address, it is set to localhost by default.

By adding this functionality we can run more than one port forwarder at the same time.

PR #511, opened 2016-10-27
[Issue-#508] - Make possible to configure the port forwarder server bind address

spolti:
**Fixes**: #508

Make possible to configure the port forwarder server bind address.
In that way you can run more than once instance of the port forwarder
on the same machine.

### DIFF
--- a/openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/PortForwarder.java
+++ b/openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/PortForwarder.java
@@ -16,25 +16,8 @@ import io.undertow.server.XnioByteBufferPool;
 import io.undertow.util.Headers;
 import io.undertow.util.Methods;
 import io.undertow.util.StringReadChannelListener;
-
-import java.io.Closeable;
-import java.io.IOException;
-import java.net.Inet4Address;
-import java.net.InetSocketAddress;
-import java.net.URI;
-import java.nio.ByteBuffer;
-import java.security.cert.X509Certificate;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.atomic.AtomicInteger;
-
-import javax.net.ssl.TrustManager;
-import javax.net.ssl.X509TrustManager;
-
 import org.xnio.BufferAllocator;
 import org.xnio.ByteBufferSlicePool;
-import org.xnio.ChannelExceptionHandler;
 import org.xnio.ChannelListener;
 import org.xnio.ChannelListeners;
 import org.xnio.IoFuture;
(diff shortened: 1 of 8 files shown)


openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/PortForwarder.java
```diff
@@ -46,13 +29,29 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.atomic.AtomicInteger;
+
 public final class PortForwarder implements Closeable {
 
     private static final String PORT_FWD = "%sapi/v1/namespaces/%s/pods/%s/portforward";
 
+    private static InetAddress portForwardBindAddress;
```
jwendell: This shouldn't not be static. Different objects should be able to listen on different addresses.

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/PortForwarder.java
```diff
@@ -245,10 +255,19 @@ public void close() {
...
-            System.out.println("Usage: portforward <namespace> <pod> <source-port> <target-port>");
-            System.out.println("Example: portforward mynamespace somepod 8080 8080");
+            System.out.println("Usage: portforward <namespace> <pod> <source-port> <target-port> -b [optional] <bindAddress>");
+            System.out.println("Example: portforward mynamespace somepod 8080 8080 -b 10.1.1.1");
+
         }
+
+        for (int i = 0; i < args.length; i++) {
+            if (args[i].contains("-b")){
+                portForwardBindAddress = InetAddress.getByName(args[i+1]);
```
jwendell: Here you should call the set() method instead of assigning the variable directly.

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/GitServer.java
```diff
@@ -86,6 +86,8 @@ private void init() throws Exception {
             forwarder.close();
         }
         forwarder = new PortForwarder(config, server.getMetadata().getName());
+        // TODO: integrate it with the ability to change the port forwarder bind address
+        //forwarder.setPortForwardBindAddress(InetAddress);
         forwarder.forwardPort(Integer.valueOf(GIT_LOCALPORT), Integer.valueOf(GIT_REMOTEPORT));
```
jwendell: I don't think this is necessary. "localhost" if fine for this purpose.

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/PortForwarder.java
spolti: I think I didn't understand your point.

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/PortForwarder.java
spolti: I don't think a good idea make the method static, for this I think only keeping the variable static is enough.

@lordofthejars wdyt in this case?

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/PortForwarder.java
jwendell: Why is this member static?

spolti:
@jwendell ok, I see your point, pr updated.

jwendell:
LGTM

lordofthejars:
Cool, thanks. Merged :)

Issue #512, opened 2016-10-28
Set relative path for file entry for use in Reports.


hemanik:
##### Issue Overview

Reports generated have the location of logs files set as standard text in format  `target/reports/logs` which prevents it from rendering proper links due to duplication in the file path.  
Relates to Recorder Issue [#55](https://github.com/arquillian/arquillian-recorder/issues/created_by/hemanik)
##### Expected Behaviour

Proper path to the location of the logs should be rendered as links in the report so that the user can directly access the logs from the report by clicking them, without having to rely on the file explorer.
##### Current Behaviour

Currently log files are stored in the `target/reports/logs` directory and in the generated report (html format) of those logs are set as standard text.

PR #513, opened 2016-10-28
Updates file path to relative path and add file entry to group entry for improved readability in reports.

hemanik:

Update the file path to relative path to avoid duplication in the file path entry while rendering the file location as link in the report.
- Relativize log file path for use by Recorder for generating reports. 
- Grouping file entry for better readability in generated reports.  

**Fixes**: #512


lordofthejars:
@hemanik @dipak-pawar it seems there is a conflict probably because of the order on how I merge the PRs. Can you inspect why it is conflicting and update the PR with the correct version? Thanks :)

Issue #484, opened 2016-10-17
Add support for network ipam


lordofthejars:
Currently networking is support without IPAM options. An example of docker compose with ipam might be:

```
version: '2'

services:
  app:
    image: busybox
    command: ifconfig
    networks:
      app_net:
        ipv4_address: 172.16.238.10
        ipv6_address: 2001:3984:3989::10

networks:
  app_net:
    driver: bridge
    driver_opts:
      com.docker.network.enable_ipv6: "true"
    ipam:
      driver: default
      config:
      - subnet: 172.16.238.0/24
        gateway: 172.16.238.1
      - subnet: 2001:3984:3989::/64
        gateway: 2001:3984:3989::1
```

This issue implies adding support for `driver_ops` (if supported by docker-java), and the ipam elements in networks section.
1. This implies modifying the custom cube format for adding support for IPAM, it is partially supported https://github.com/arquillian/arquillian-cube/blob/master/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/config/Network.java#L8 but `driver_opts` not added yet
2. Adding the logic in DockerClientExecutor https://github.com/arquillian/arquillian-cube/blob/master/docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/DockerClientExecutor.java#L889 
3. Add in DockerComposeConverter the logic to convert from DockerCompose to Cube format https://github.com/arquillian/arquillian-cube/blob/master/docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/compose/ComposeBuilder.java#L66

Finally `ipv4_address` and `ipv6_address` fields of container network section must be set (and this is something to explore how to do it since it is not clear how to do it with docker-java)

dipak-pawar:
Opened #514

PR #514, opened 2016-11-04
[Issue #484]Implements network ipam support. 

dipak-pawar:
- Adds ipam support for network

- Added ipam support
- Added driver_opts support
- You can assign static ip Address using docker-compose v2.

**Fixes**: #484

### DIFF
--- a/docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/compose/ComposeBuilder.java
+++ b/docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/compose/ComposeBuilder.java
@@ -69,7 +69,8 @@ public class ComposeBuilder {
                             String networkName = serviceNetworks.iterator().next();
                             cubeContainer.setNetworkMode(networkName);
                             if (serviceNetworksConfig != null) {
-                                setNetworkOptions(serviceNetworksConfig.get(networkName), cubeContainer);
+                                Map<String, Object> configOption = (LinkedHashMap) serviceNetworksConfig.get(networkName);
+                                setNetworkOptions(configOption, cubeContainer);
                             }
                         } else {
                             throw new IllegalArgumentException("Networks not mentioned under services networks section.");
@@ -113,14 +114,13 @@ public class ComposeBuilder {
                 .getName() + NETWORK_NAME_SUFFIX;
     }
 
-    private void setNetworkOptions(Object networkOption, CubeContainer cubeContainer){
+    private void setNetworkOptions(Map<String, Object> networkOption, CubeContainer cubeContainer){
         if (networkOption != null) {
-            Map<String, Object> options = (LinkedHashMap) networkOption;
-            if (options.containsKey(IP_V4_Address)) {
-                cubeContainer.setIpv4Address(asString(options, IP_V4_Address));
+            if (networkOption.containsKey(IP_V4_Address)) {
+                cubeContainer.setIpv4Address(asString(networkOption, IP_V4_Address));
             }
-            if (options.containsKey(IP_V6_Address)) {
-                cubeContainer.setIpv6Address(asString(options, IP_V6_Address));
+            if (networkOption.containsKey(IP_V6_Address)) {
+                cubeContainer.setIpv6Address(asString(networkOption, IP_V6_Address));
             }
         }
     }
--- a/docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/compose/ComposeBuilder.java
+++ b/docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/compose/ComposeBuilder.java
@@ -58,22 +58,22 @@ public class ComposeBuilder {
                     if (serviceDefinition.containsKey(NETWORKS)) {
                         Collection<String> serviceNetworks;
                         Map<String, Object> serviceNetworksConfig = null;
-                            if (serviceDefinition.get(NETWORKS) instanceof ArrayList) {
-                                serviceNetworks = asListOfString(serviceDefinition, NETWORKS);
-                            } else {
-                                serviceNetworksConfig = asMap(serviceDefinition, NETWORKS);
-                                serviceNetworks = serviceNetworksConfig.keySet();
+                        if (serviceDefinition.get(NETWORKS) instanceof ArrayList) {
+                            serviceNetworks = asListOfString(serviceDefinition, NETWORKS);
+                        } else {
+                            serviceNetworksConfig = asMap(serviceDefinition, NETWORKS);
+                            serviceNetworks = serviceNetworksConfig.keySet();
+                        }
+                        if (!serviceNetworks.isEmpty()) {
+                            cubeContainer.setNetworks(new HashSet<>(serviceNetworks));
+                            String networkName = serviceNetworks.iterator().next();
+                            cubeContainer.setNetworkMode(networkName);
+                            if (serviceNetworksConfig != null) {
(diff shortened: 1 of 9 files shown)


docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/compose/ComposeBuilder.java
```diff
@@ -51,13 +56,24 @@ public DockerCompositions build(Map<String, Object> dockerComposeContainerDefini
...
-                        final Collection<String> serviceNetworks = asListOfString(serviceDefinition, NETWORKS);
-                        if (!serviceNetworks.isEmpty()) {
-                            cubeContainer.setNetworks(new HashSet<>(serviceNetworks));
-                            cubeContainer.setNetworkMode(serviceNetworks.iterator().next());
-                        } else {
-                            throw new IllegalArgumentException("Networks not mentioned under services networks section.");
-                        }
+                        Collection<String> serviceNetworks;
+                        Map<String, Object> serviceNetworksConfig = null;
+                            if (serviceDefinition.get(NETWORKS) instanceof ArrayList) {
```
bartoszmajsak: Indentation is wrong.

bartoszmajsak: Other types of lists/collections are not supported? Then we treat any other type as `Map`? WDYT @lordofthejars?

docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/compose/ComposeBuilder.java
```diff
@@ -97,4 +113,15 @@ private String getDefaultNetworkName() {
                 .getName() + NETWORK_NAME_SUFFIX;
     }
 
+    private void setNetworkOptions(Object networkOption, CubeContainer cubeContainer){
+        if (networkOption != null) {
+            LinkedHashMap<String, Object> options = (LinkedHashMap) networkOption;
```
bartoszmajsak: Doesn't have to be exact type - `Map` should work here I guess.

docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/compose/NetworkBuilder.java
```diff
@@ -42,6 +64,43 @@ public NetworkBuilder withDefaultDriver() {
...
+        }
+        this.configuration.setIpam(ipam);
+
+        return this;
+    }
+
+    private List<IPAMConfig> createIpamConfig(Collection<Map<String, Object>> configs) {
+        List<IPAMConfig> ipamConfigs = new ArrayList<>();
+        for (Object config : configs) {
+            if (config instanceof LinkedHashMap) {
```
bartoszmajsak: Any other concrete type of `Map` interface is not good here? Why?

docker/ftest-docker-machine-network/src/test/java/org/arquillian/cube/servlet/PingPongTest.java
```diff
@@ -73,4 +78,35 @@ public void container_should_be_attached_to_back_network() {
...
+        ExecCreateCmdResponse execCreateCmdResponse = dockerClient.execCreateCmd(pingpong.getId())
+                .withAttachStdout(true).withAttachStdin(true).withAttachStderr(true).withTty(false).withCmd("ifconfig")
+                .exec();
+
+        OutputStream outputStream = new ByteArrayOutputStream();
+        OutputStream errorStream = new ByteArrayOutputStream();
+        dockerClient.execStartCmd(execCreateCmdResponse.getId()).withDetach(false)
+                .exec(new ExecStartResultCallback(outputStream, errorStream)).awaitCompletion();
+
+        assertThat(outputStream.toString()).contains("inet addr:172.16.238.10", "inet6 addr: fe80::42:acff:fe10:ee0a/64");
```
bartoszmajsak: Shouldn't we close `outputStream` and `errorStream` after the test?

docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/compose/ComposeBuilder.java
dipak-pawar: Yes map will work.

docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/compose/NetworkBuilder.java
dipak-pawar: not really, you should use any other concrete type, not only LinkedHashMap

docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/compose/ComposeBuilder.java
dipak-pawar: Currently after parsing networks from serviceDefinition we are getting Only ArrayList or Map. Other types of lists/collections are not supported. Previously it was supported only for List.

docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/compose/NetworkBuilder.java
```diff
@@ -82,9 +82,9 @@ public NetworkBuilder addIpam(Map<String, Object> ipamConfig) {
     private List<IPAMConfig> createIpamConfig(Collection<Map<String, Object>> configs) {
         List<IPAMConfig> ipamConfigs = new ArrayList<>();
         for (Object config : configs) {
-            if (config instanceof LinkedHashMap) {
+            if (config instanceof Map) {
```
bartoszmajsak: Each element in the collection is a `Map` as per method signature. There is no need for using `instanceof` and casting if you tweak for-each loop properly.

docker/ftest-docker-machine-network/src/test/java/org/arquillian/cube/servlet/PingPongTest.java
```diff
@@ -107,6 +107,7 @@ public void container_should_have_static_ip_for_app_net_network() throws Interru
                 .exec(new ExecStartResultCallback(outputStream, errorStream)).awaitCompletion();
 
         assertThat(outputStream.toString()).contains("inet addr:172.16.238.10", "inet6 addr: fe80::42:acff:fe10:ee0a/64");
-
+        outputStream.close();
+        errorStream.close();
     }
```
bartoszmajsak: I believe we should use `try-with-resources` instead. Or at least try-catch-finally idiom

docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/compose/ComposeBuilder.java
```diff
@@ -97,4 +113,15 @@ private String getDefaultNetworkName() {
                 .getName() + NETWORK_NAME_SUFFIX;
     }
 
+    private void setNetworkOptions(Object networkOption, CubeContainer cubeContainer){
+        if (networkOption != null) {
+            Map<String, Object> options = (LinkedHashMap) networkOption;
```
bartoszmajsak: Why `Object networkOption` as parameter? Don't we immediately assume that this object is a map? We should be more specific about what are the expected inbound and outbound types, otherwise code becomes tricky to follow and reason about

lordofthejars:
LGTM

Issue #515, opened 2016-11-07
Statistics are not showing in human readable format for negative number.


dipak-pawar:
Bug

##### Issue Overview
Docker statistics are not showing converted human readable number for negative value.

##### Expected Behaviour
It should show negative as sign in front of human readable value.

##### Current Behaviour
Cube is not converting to negative number to human readable format.
 
##### Steps To Reproduce
1. Run any sample test with recorder dependency added in class path.
2. Check the report.
![screenshot from 2016-11-07 12-55-41](https://cloud.githubusercontent.com/assets/7941617/20053251/b4fb8986-a4fd-11e6-858a-69c85c4b0c6d.png)

PR #516, opened 2016-11-07
[Issue #515] Modifies logic to show negative numbers  in human readable format.

dipak-pawar:
If there is negative number in container statistics then it should show proper human readable number with sign.

- Changed logic to show human readable format.

**Fixes**: #515

### DIFF
--- a/docker/recorder/src/main/java/org/arquillian/cube/docker/impl/client/utils/NumberConversion.java
+++ b/docker/recorder/src/main/java/org/arquillian/cube/docker/impl/client/utils/NumberConversion.java
@@ -6,13 +6,15 @@ import org.arquillian.cube.docker.impl.util.NumberType;
 public class NumberConversion {
 
     public static String humanReadableByteCount(Long bytes, boolean decimal) {
-        int unit = decimal ? 1000 : 1024;
         if (bytes == null) bytes = 0L;
-        if (bytes < unit) return bytes + " B";
-        int exp = (int) (Math.log(bytes) / Math.log(unit));
+        String sign = bytes < 0 ? "-" : "";
+        Long absBytes = Math.abs(bytes);
+        int unit = decimal ? 1000 : 1024;
+        if (absBytes < unit) return sign + absBytes + " B";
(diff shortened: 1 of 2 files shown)


docker/recorder/src/main/java/org/arquillian/cube/docker/impl/client/utils/NumberConversion.java
```diff
@@ -6,13 +6,15 @@
...
-        int exp = (int) (Math.log(bytes) / Math.log(unit));
+        String sign = bytes < 0 ? "-" : "";
+        Long absBytes = Math.abs(bytes);
+        int unit = decimal ? 1000 : 1024;
+        if (absBytes < unit) return sign + absBytes + " B";
+        int exp = (int) (Math.log(absBytes) / Math.log(unit));
         String pre = (decimal ? "kMGTPE" : "KMGTPE").charAt(exp-1) + (decimal ? "" : "i");
 
-        return String.format("%.2f %sB", bytes / Math.pow(unit, exp), pre);
+        return sign + String.format("%.2f %sB", absBytes / Math.pow(unit, exp), pre);
```
lordofthejars: You can add sign into `format` something like `String.format("%s %.2f......", sign, .....)`

docker/recorder/src/main/java/org/arquillian/cube/docker/impl/client/utils/NumberConversion.java
```diff
@@ -6,13 +6,15 @@
...
 
     public static String humanReadableByteCount(Long bytes, boolean decimal) {
-        int unit = decimal ? 1000 : 1024;
         if (bytes == null) bytes = 0L;
-        if (bytes < unit) return bytes + " B";
-        int exp = (int) (Math.log(bytes) / Math.log(unit));
+        String sign = bytes < 0 ? "-" : "";
+        Long absBytes = Math.abs(bytes);
+        int unit = decimal ? 1000 : 1024;
+        if (absBytes < unit) return sign + absBytes + " B";
```
lordofthejars: can you do an expanded if? `if(){}` just to not skip that this if have a `return`

PR #518, opened 2016-11-08
[Issue #517] - NPE at ContainerConfigurationController.remapContainer()

### LINKED ISSUE #517: NPE at ContainerConfigurationController.remapContainer()
```
java.lang.NullPointerException	at org.arquillian.cube.impl.client.container.ContainerConfigurationController.remapContainer(ContainerConfigurationController.java:86)
```

This was introduced by commit 7b890fe535c625390bb8f716455c618c39c5e490.

spolti:
If the user specifies the managementPort the if condition on line 71 will not be satisfied 
and the mappingForPort object will not be populated, so when the management address is getting
overridden it will throw an NPE exception when trying to do mappingForPort.getIP().

### DIFF
--- a/core/src/main/java/org/arquillian/cube/impl/client/container/ContainerConfigurationController.java
+++ b/core/src/main/java/org/arquillian/cube/impl/client/container/ContainerConfigurationController.java
@@ -64,15 +64,15 @@ public class ContainerConfigurationController {
         Object newConfigurationInstance = configurationClass.newInstance();
 
         for (PropertyDescriptor configurationClassPortField : configurationClassPortFields) {
+            int containerPort = getDefaultPortFromConfigurationInstance(newConfigurationInstance,
+                    configurationClass, configurationClassPortField);
+            mappingForPort = bindings.getMappedAddress(containerPort);
+
             if (!portPropertiesFromArquillianConfigurationFile.contains(configurationClassPortField.getName()) && (configurationClassPortField.getPropertyType().equals(Integer.class) || configurationClassPortField.getPropertyType().equals(int.class))) {
                 // This means that port has not configured in arquillian.xml and it will use default value.
                 // In this case is when remapping should be activated to adequate the situation according to
                 // Arquillian Cube exposed ports.
-
-                int containerPort = getDefaultPortFromConfigurationInstance(newConfigurationInstance,
-                        configurationClass, configurationClassPortField);
-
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/arquillian/cube/impl/client/container/ContainerConfigurationController.java
```diff
@@ -64,15 +64,16 @@ public void remapContainer(@Observes BeforeSetup event, CubeRegistry cubeRegistr
         Object newConfigurationInstance = configurationClass.newInstance();
 
         for (PropertyDescriptor configurationClassPortField : configurationClassPortFields) {
+            // Move it outside if condition below, if there is a group of containers.
```
jwendell: What's the point of this comment here? Does it make sense to anyone reading the source code (and not this commit diff)?

core/src/main/java/org/arquillian/cube/impl/client/container/ContainerConfigurationController.java
```diff
@@ -64,15 +64,16 @@ public void remapContainer(@Observes BeforeSetup event, CubeRegistry cubeRegistr
...
+            int containerPort = getDefaultPortFromConfigurationInstance(newConfigurationInstance,
```
jwendell: The error will persist if the it doesn't enter on this for loop.
Does it make sense to move it above the for? If not you'll have to repeat the assignment and the conditional inside the second for.

spolti: I guess it will always enter in the first for loop.
I am not sure if we need to move it to the another for, because if there is no port to override also there is no address too.

Probably by adding this same if condition ( if (mappingForPort != null)  ) on the second for loop might be enough.

Does it make sense?

core/src/main/java/org/arquillian/cube/impl/client/container/ContainerConfigurationController.java
jwendell: Just remove it.

core/src/main/java/org/arquillian/cube/impl/client/container/ContainerConfigurationController.java
jwendell: I don't have enough knowledge on this code base to comment the correct approach.
However, generally speaking, you need the conditional to avoid a NPE.

Issue #519, opened 2016-11-09
StatsLogsResultCallback is not releasing thread only for Tomcat Container.


dipak-pawar:
Bug 

##### Issue Overview
When Using tomcat container with Recorder dependency added in your class path. Test execution is going in infinite loop due to StatsLogsResultcallback onComplete event.

##### Expected Behaviour
Test execution should be finished within few minutes.

##### Current Behaviour
Test execution is not finishing & going in infinite state.

##### Steps To Reproduce
1. run ftest-reporter module using `mvn test` or in editor.

PR #520, opened 2016-11-09
Added CountLatch to use it for StatsCmd.

dipak-pawar:
Only tomcat container is stopping other thread execution even if call `onComplete()` event in StatsLogsResultCallback when you have added recorder dependency in your class path.

- Added CountDownLatch for StatsLogsResultCallback So even if after onComplete event is stopping execution of other thread, it'll timeout after timeout period.

**Fixes**: #519

### DIFF
--- a/docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/DockerClientExecutor.java
+++ b/docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/DockerClientExecutor.java
@@ -19,6 +19,8 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.logging.Logger;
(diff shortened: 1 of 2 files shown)


docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/DockerClientExecutor.java
```diff
@@ -500,9 +502,12 @@ public Statistics statsContainer(String id) throws IOException {
 
         try {
             StatsCmd statsCmd = this.dockerClient.statsCmd(id);
-            StatsLogsResultCallback statslogs = new StatsLogsResultCallback();
+            CountDownLatch countDownLatch = new CountDownLatch(1);
+            StatsLogsResultCallback statslogs = new StatsLogsResultCallback(countDownLatch);
             try {
-                statsCmd.exec(statslogs).awaitCompletion();
+                StatsLogsResultCallback statscallback = statsCmd.exec(statslogs);
+                countDownLatch.await(3, TimeUnit.SECONDS);
```
lordofthejars: Can you change this to 5 seconds, I am thinking on remote docker host.

bartoszmajsak: Wouldn't it be possible to use `CompletableFuture` instead?

PR #528, opened 2016-11-23
Fix broken tests.

iocanel:
CubeConfiguratorTest is broken by 47dbf9d3f74c43c79100b018355b43e06216ae47.

Now that we add `tlsVerify` when `tcp` scheme is, the `certPath` will get removed. So we need to align tests accordingly.


lordofthejars:
Thank you so much

PR #529, opened 2016-11-24
Align

iocanel:
Introduce `@RequiresOpenshift` annotation.
Removes hardcoded server urls from arquillian.xml.
Fix some CNFE in openshift tests.


iocanel:
CircleCI fails at the moment, because the master is broken. Fix for master is at #528, so we should probablly apply that first.

lordofthejars:
Thank you so much. I merge this since errors does not come from this PR.

Issue #530, opened 2016-11-28
Change from configuring RestAssured using static code to RequestSpecBuilder


lordofthejars:
##### Issue Overview

Change from configuring RestAssured using static code to RequestSpecBuilder

```
@BeforeClass
public static void initSpec(){
    spec = new RequestSpecBuilder()
            .setContentType(ContentType.JSON)
            .setBaseUri("http://localhost:8080/")
            .addFilter(new ResponseLoggingFilter())//log request and response for better debugging. You can also only log if a requests fails.
            .addFilter(new RequestLoggingFilter())
            .build();
}
@Test
public void useSpec(){
    given()
            .spec(spec)
            .param("limit", 20)
            .when()
            .get("blogs")
            .then()
            .statusCode(200);
}
```

PR #531, opened 2016-11-28
changes configuring restassuerd from static code to RequestSpecBuilder

lordofthejars:

Instead of configuring RestAssured using static code, now users can inject `RequestSpecBuilder` in test with docker values initialized.

- RestAssured static code is not called anymore
- `RequestSpecBuilder` is injected with docker values
-

**Fixes**: #530

### DIFF
--- a/docker/ftest-restassured/src/test/java/org/arquillian/cube/restassured/PingPongTest.java
+++ b/docker/ftest-restassured/src/test/java/org/arquillian/cube/restassured/PingPongTest.java
@@ -1,8 +1,10 @@
 package org.arquillian.cube.restassured;
 
 import io.restassured.RestAssured;
+import io.restassured.builder.RequestSpecBuilder;
 import org.arquillian.cube.docker.impl.requirement.RequiresDockerMachine;
 import org.arquillian.cube.requirement.ArquillianConditionalRunner;
+import org.jboss.arquillian.test.api.ArquillianResource;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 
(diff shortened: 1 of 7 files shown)


docker/ftest-restassured/src/test/java/org/arquillian/cube/restassured/PingPongTest.java
```diff
@@ -14,9 +16,18 @@
...
 
+    @ArquillianResource
+    RequestSpecBuilder requestSpecBuilder;
+
     @Test
     public void should_receive_ok_message() throws MalformedURLException, InterruptedException {
-        RestAssured.get().then().assertThat().body("status", equalTo("OK"));
+        RestAssured
+                .given()
+                .spec(requestSpecBuilder.build())
```
bartoszmajsak: Formatting might use a bit more indentation to mark GWT blocks.

bartoszmajsak:
I have one general remark to this approach - it locks the user to the certain `rest-assured` version we use in Cube.

Wouldn't it be easier to just have all those configuration bits injectable to the test itself and let the user configure request spec as they wish?

lordofthejars:
All these information are also provided by Docker Cube Core, so if we let user create their own `RequestSpecBuilder` then this extension does not have sense at all. Notice that this extension is used to avoid user having to repeat the same over and over again in all tests (I mean injecting fields and creating the class). If we decide to not creating the class `RequestSpecBuilder` class in extension, then this extension is merely a reporter extension.

bartoszmajsak:
> All these information are also provided by Docker Cube Core

This is exactly my point - I'm not sure if there is a value in having it as an extension for the very reasons I mentioned in the previous comment (locking users with certain version of rest-assured)

lordofthejars:
Well the idea is to avoid the repeat yourself in every test doing the same, building the `RequestSpecBuilder`. We can remove this PR and just leave restassured module as reporting features.

lordofthejars:
What I would do is just squash this PR, and then open a new one with the removal. Just to make commits have the right change.

bartoszmajsak:
Ok, if you feel like removing it makes sense, go for it.

PR #533, opened 2016-12-13
CircleCI now uses a machine with docker.

iocanel:
We need to run our tests on a machine that has at least docker available, so that @RequiresDocker is satisfied.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -13,6 +13,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+machine:
+  java:
+    version: openjdk8
+  services:
+    - docker
 
 dependencies:
   override:


circle.yml
```diff
@@ -13,6 +13,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+machine:
+  java:
+    version: openjdk7
```
lordofthejars: can we change this to openjdk8? Notice that current parent jboss pom specifies java 8 as version.

iocanel: sure

bartoszmajsak:
Since we have introduced this our builds are failing with
```
com.github.dockerjava.api.exception.InternalServerErrorException: Unsupported: Exec is not supported by the lxc driver
```

Here's the suggested workaround https://circleci.com/docs/docker/#docker-exec (not sure if applicable in our context though).

Shall we open new issue or re-open this one?

lordofthejars:
@bartoszmajsak in our context this is not applicable since we are using the REST API inside Cube.

iocanel:
Damn!

bartoszmajsak:
No worries, you have broken two commits so far - it's only two beers you owe the project :P But we keep counting

evil laugh

PR #534, opened 2016-12-13
Add support for the f-m-p version 3.x.

iocanel:
The fabric8-maven-plugin in version 3.x generates and stores kubernetes/openshift resources in a different directory.

It would be awesome if we also looked there, when we are trying to discover kubernetes/openshift resources.

### DIFF
--- a/kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/locator/DefaultKubernetesResouceLocator.java
+++ b/kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/locator/DefaultKubernetesResouceLocator.java
@@ -7,24 +7,23 @@ import java.net.URL;
 public class DefaultKubernetesResouceLocator implements KubernetesResourceLocator {
 
     private static final String ROOT = "/";
-    private static final String RESOURCE_NAME = "kubernetes";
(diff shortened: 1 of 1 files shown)


kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/locator/DefaultKubernetesResouceLocator.java
```diff
@@ -7,24 +7,23 @@
...
 
-
     @Override
     public URL locate() {
         for (String suffix : ALLOWED_SUFFIXES) {
-            URL candidate = getResource(RESOURCE_NAME + suffix);
-            if (candidate != null) {
-                return candidate;
+            for (String resource : RESOURCE_NAMES) {
+                URL candidate = getResource(resource + suffix);
```
lordofthejars: We give priority to `kubernetes` location right? I think this is how is should be, but just to be sure.

iocanel: In practice I think that most user will use f-m-p version 3.x so using as a default `META-INF/fabric8/kubernetes` would possibly be best. On the other hand using something generic as a default felt better, so I went with `kubernetes`.

Issue #536, opened 2016-12-19
[Enhacement] Implement Reporter Integration


dipak-pawar:
Implementation of reporter integration.

User should be able to get report with configuration file & session status which include Pods, Services, Replication controller, namespace etc.

PR #537, opened 2016-12-19
Adds kubernetes reporter integration.

dipak-pawar:
Implements Reporter integration with cube kubernetes

Adds kubernetes reporter 
- To report config file.
- To report dynamic properties of session - namespace, pods, rc, services etc.

**Fixes**: #536

### DIFF
--- a/kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/SessionCreatedListener.java
+++ b/kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/SessionCreatedListener.java
@@ -15,9 +15,11 @@ import org.arquillian.cube.kubernetes.api.Logger;
 import org.arquillian.cube.kubernetes.api.NamespaceService;
 import org.arquillian.cube.kubernetes.api.Session;
 import org.arquillian.cube.kubernetes.impl.await.WaitStrategy;
+import org.arquillian.cube.kubernetes.impl.event.AfterStart;
 import org.arquillian.cube.kubernetes.impl.event.Start;
 import org.arquillian.cube.kubernetes.impl.event.Stop;
 import org.arquillian.cube.kubernetes.impl.visitor.CompositeVisitor;
+import org.jboss.arquillian.core.api.Event;
 import org.jboss.arquillian.core.api.Instance;
 import org.jboss.arquillian.core.api.annotation.Inject;
 import org.jboss.arquillian.core.api.annotation.Observes;
@@ -50,6 +52,9 @@ public class SessionCreatedListener {
 
     @Inject
     Instance<ServiceLoader> serviceLoader;
+
+    @Inject
+    Event<AfterStart> afterStartEvent;
 
     private ShutdownHook shutdownHook;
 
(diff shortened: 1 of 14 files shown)


kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/SessionCreatedListener.java
```diff
@@ -122,6 +127,7 @@ public void run() {
 
             if (!configuration.isEnvironmentInitEnabled() || waitStrategy.await()) {
                 displaySessionStatus(session);
+                afterStartEvent.fire(new AfterStart(session));
```
bartoszmajsak: Just curious - why do we need to fire this event instead of directly firing `SessionCreatedEvent`?

kubernetes/reporter/src/main/java/org/arquillian/cube/kubernetes/reporter/TakeKubernetesResourcesInformation.java
```diff
@@ -0,0 +1,175 @@
...
+
+    private void addEntry(GroupEntry groupEntry, PropertyEntry propertyEntry) {
+        groupEntry.getPropertyEntries().add(propertyEntry);
+    }
+
+    private String getFileType(Path path) throws IOException {
+        String extension = FilenameUtils.getExtension(path.toString());
+        String type = "";
+        if ("json".equals(extension)) {
+            type = Files.probeContentType(path);
```
bartoszmajsak: Do we really have to `probe`? We have total control and knowledge at this point what we are passing along, so I'm not sure if we really have to check if we are dealing with json?

dipak-pawar: yeah just wants to avoid hard coding values.

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/SessionCreatedListener.java
dipak-pawar: start & stop event classes are inheriting from SessionCreatedEvent. So if we use it directly then your observer is calling for firing of start & stop as well. I don't want to invoke it for all fires of start & stop.

bartoszmajsak: Ok it's now clear.

bartoszmajsak:
LGTM

kubernetes/reporter/src/main/java/org/arquillian/cube/kubernetes/reporter/KubernetesReporterExtension.java
```diff
@@ -0,0 +1,15 @@
+package org.arquillian.cube.kubernetes.reporter;
+
+import org.jboss.arquillian.core.spi.LoadableExtension;
+
+
+public class KubernetesReporterExtension implements LoadableExtension{
+
+    @Override
+    public void register(ExtensionBuilder builder) {
+        final boolean reportedInClasspath = Validate.classExists("org.arquillian.recorder.reporter.ReporterExtension");
```
lordofthejars: Do you mean `reporterInClasspath`? maybe there is a typo as you used `reported`.

MatousJobanek:
LGTM

Issue #539, opened 2016-12-22
[Enhancement] Add support to show multiple configuration files in reporter


dipak-pawar:
##### Issue Overview
All configuration files should be accessible from report.

PR #540, opened 2016-12-22
Adds support for multiple config files in kubernetes reporter

dipak-pawar:
If user is running services using multiple configuration file, then all files should be display in reports

- Support for multiple files from local resources & from http url

**Fixes**: #539 
screenshot from 2016-12-22 13-18-42

### DIFF
--- a/kubernetes/reporter/pom.xml
+++ b/kubernetes/reporter/pom.xml
@@ -61,18 +61,4 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
-
-  <build>
-    <plugins>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-compiler-plugin</artifactId>
-        <configuration>
-          <source>1.8</source>
-          <target>1.8</target>
-        </configuration>
-      </plugin>
-    </plugins>
-  </build>
-
 </project>
--- a/kubernetes/reporter/pom.xml
+++ b/kubernetes/reporter/pom.xml
@@ -61,4 +61,18 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <source>1.8</source>
+          <target>1.8</target>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+
 </project>
--- a/kubernetes/reporter/src/main/java/org/arquillian/cube/kubernetes/reporter/TakeKubernetesResourcesInformation.java
+++ b/kubernetes/reporter/src/main/java/org/arquillian/cube/kubernetes/reporter/TakeKubernetesResourcesInformation.java
@@ -139,8 +139,8 @@ public class TakeKubernetesResourcesInformation {
         return tableEntry;
     }
 
-    private ArrayList<FileEntry> getFilesForResourcesConfiguration(Session session, Configuration configuration, ReporterConfiguration reporterConfiguration) throws IOException {
-        ArrayList<FileEntry> fileEntries = new ArrayList<>();
+    private List<FileEntry> getFilesForResourcesConfiguration(Session session, Configuration configuration, ReporterConfiguration reporterConfiguration) throws IOException {
+        final List<FileEntry> fileEntries = new ArrayList<>();
         URL environmentConfigUrl = configuration.getEnvironmentConfigUrl();
 
         if (environmentConfigUrl != null) {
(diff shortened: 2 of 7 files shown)


kubernetes/reporter/pom.xml
```diff
@@ -61,4 +61,18 @@
...
     </dependency>
   </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <source>1.8</source>
```
bartoszmajsak: Is it necessary? If we inherit from `jboss-parent:21` it's by default JDK 8.

kubernetes/reporter/src/main/java/org/arquillian/cube/kubernetes/reporter/TakeKubernetesResourcesInformation.java
```diff
@@ -130,20 +139,53 @@ private TableEntry createTableEntryWithTHead(String name, String... heads) {
         return tableEntry;
     }
 
-    private FileEntry getFileForResourcesConfiguration(Configuration configuration, ReporterConfiguration reporterConfiguration) throws IOException {
-        String path = configuration.getEnvironmentConfigUrl().getPath();
+    private ArrayList<FileEntry> getFilesForResourcesConfiguration(Session session, Configuration configuration, ReporterConfiguration reporterConfiguration) throws IOException {
+        ArrayList<FileEntry> fileEntries = new ArrayList<>();
```
bartoszmajsak: Let's make it `final List<>`

kubernetes/reporter/src/main/java/org/arquillian/cube/kubernetes/reporter/TakeKubernetesResourcesInformation.java
```diff
@@ -130,20 +139,53 @@ private TableEntry createTableEntryWithTHead(String name, String... heads) {
...
+        URL environmentConfigUrl = configuration.getEnvironmentConfigUrl();
+
+        if (environmentConfigUrl != null) {
+            fileEntries.add(getFileForResourcesConfiguration(environmentConfigUrl, reporterConfiguration));
+        }
+
+        if (configuration.isEnvironmentInitEnabled()) {
+            List<URL> dependencyUrls = !configuration.getEnvironmentDependencies().isEmpty() ? configuration.getEnvironmentDependencies() : dependencyResolver.get().resolve(session);
```
bartoszmajsak: Can `configuration.getEnvironmentDependencies()` be `null`? If so we need to check that too.

kubernetes/reporter/pom.xml
bartoszmajsak: @lordofthejars are we fine with using `JDK 8` for Cube-Q?

kubernetes/reporter/pom.xml
```diff
@@ -61,4 +61,18 @@
...
+          <target>1.8</target>
```
lordofthejars: I think that jboss parent pom we are inheriting already defines this

kubernetes/reporter/src/main/java/org/arquillian/cube/kubernetes/reporter/TakeKubernetesResourcesInformation.java
```diff
@@ -130,20 +139,53 @@ private TableEntry createTableEntryWithTHead(String name, String... heads) {
...
-        fileEntry.setPath(relativize.toString());
+        fileEntry.setPath(filePath);
         fileEntry.setMessage(CONFIGURATION);
-        fileEntry.setType(getFileType(relativize));
+        fileEntry.setType(getFileType(filePath));
 
         return fileEntry;
     }
 
+    private String relativizePath(URL url, Path rootDir) {
```
lordofthejars: I would return `Path` here and be the caller who decides if to do a `toString` or whatever they need.

kubernetes/reporter/src/main/java/org/arquillian/cube/kubernetes/reporter/TakeKubernetesResourcesInformation.java
dipak-pawar: no it can't be null

dipak-pawar:
@Bartosz - 
1.Config URL -> you can give it local file path or HTTP URL(`env.config.url`) or Resource Name(`env.config.resource.name`). firstly it will check for property in arquillian.xml. If not defined then it'll look for default file by name kubernetes.json in test resources folder.

2.Environment dependies:
You can give multiple files as environment dependencies - using `env.dependencies` property.

So before applying kubernetes configuration, it has config URL & environment dependencies URLs.

kubernetes/reporter/pom.xml
dipak-pawar: missed it as Jboss has lower version(19) in parent which is using java 7. now updating it 21

kubernetes/reporter/src/main/java/org/arquillian/cube/kubernetes/reporter/TakeKubernetesResourcesInformation.java
dipak-pawar: Yeah to convert uri to path as it depends in filesystem, it will remove one backslash on linux
so `http://foo.com` will convert to `http:/foo.com`. So to avoid this. continuing with the same.

bartoszmajsak:
Can we document this behaviour as part of this PR?

lordofthejars:
I squash since documentation comes in PR https://github.com/arquillian/arquillian-cube/pull/542

Issue #527, opened 2016-11-18
Unable to use ArquillianConditionalRunner for in-container tests


jamesnetherton:
##### Issue Overview

I'd like to make use of `ArquillianConditionalRunner` and most of the other good stuff in the `org.arquillian.cube.requirement` package for in-container tests (E.g on WildFly).

If I try to do this I get a `ClassNotFoundException`. I guess a workaround would be to add this class / package to my test deployment, but this would need to be duplicated on each and every test case.

##### Expected Behaviour

`ArquillianConditionalRunner` and other supporting classes in `org.arquillian.cube.requirement` are available to in-container tests.

##### Current Behaviour

CNFE:

```
java.lang.TypeNotPresentException: Type org.arquillian.cube.requirement.ArquillianConditionalRunner not present
	at sun.reflect.annotation.TypeNotPresentExceptionProxy.generateException(TypeNotPresentExceptionProxy.java:46)
	at sun.reflect.annotation.AnnotationInvocationHandler.invoke(AnnotationInvocationHandler.java:84)
	at com.sun.proxy.$Proxy19.value(Unknown Source)
	at org.junit.internal.builders.AnnotatedBuilder.runnerForClass(AnnotatedBuilder.java:21)
	at org.junit.runners.model.RunnerBuilder.safeRunnerForClass(RunnerBuilder.java:59)
	at org.junit.internal.builders.AllDefaultPossibilitiesBuilder.runnerForClass(AllDefaultPossibilitiesBuilder.java:26)
	at org.junit.runners.model.RunnerBuilder.safeRunnerForClass(RunnerBuilder.java:59)
	at org.junit.internal.requests.ClassRequest.getRunner(ClassRequest.java:26)
	at org.junit.internal.requests.FilterRequest.getRunner(FilterRequest.java:31)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:138)
	at org.jboss.arquillian.junit.container.JUnitTestRunner.execute(JUnitTestRunner.java:66)
	at org.jboss.arquillian.protocol.jmx.JMXTestRunner.doRunTestMethod(JMXTestRunner.java:180)
	at org.jboss.as.arquillian.service.ArquillianService$ExtendedJMXTestRunner.doRunTestMethod(ArquillianService.java:247)
	at org.jboss.arquillian.protocol.jmx.JMXTestRunner.runTestMethodInternal(JMXTestRunner.java:162)
[...]

PR #543, opened 2017-01-09
Adds requirements classes if tests running as in-container.

dipak-pawar:
 Adds required classes to run ArquillianConditionalRunner if we run tests as incontainer tests.

**Fixes**: #527

### DIFF
--- a/core/src/main/java/org/arquillian/cube/impl/client/container/remote/CubeAuxiliaryArchiveAppender.java
+++ b/core/src/main/java/org/arquillian/cube/impl/client/container/remote/CubeAuxiliaryArchiveAppender.java
@@ -2,8 +2,11 @@ package org.arquillian.cube.impl.client.container.remote;
 
 import org.arquillian.cube.CubeController;
 import org.arquillian.cube.impl.client.enricher.CubeControllerProvider;
+import org.arquillian.cube.requirement.ArquillianConditionalRunner;
+import org.arquillian.cube.spi.requirement.Requires;
 import org.jboss.arquillian.container.test.spi.RemoteLoadableExtension;
 import org.jboss.arquillian.container.test.spi.client.deployment.AuxiliaryArchiveAppender;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/arquillian/cube/impl/client/container/remote/CubeAuxiliaryArchiveAppender.java
```diff
@@ -12,10 +15,17 @@
...
     public Archive<?> createAuxiliaryArchive() {
-        return ShrinkWrap.create(JavaArchive.class, "arquillian-cube.jar")
+        JavaArchive javaArchive = ShrinkWrap.create(JavaArchive.class, "arquillian-cube.jar")
                 .addPackage(CubeController.class.getPackage())
                 .addPackages(true, CubeRemoteExtension.class.getPackage())
                 .addClass(CubeControllerProvider.class)
                 .addAsServiceProvider(RemoteLoadableExtension.class, CubeRemoteExtension.class);
+
+        if (LoadableExtension.Validate.classExists("org.arquillian.cube.requirement.ArquillianConditionalRunner") &&
+                LoadableExtension.Validate.classExists("org.arquillian.cube.spi.requirement.Requires")) {
```
lordofthejars: Requires will be always there since it is in `spi` module.

dipak-pawar: yes but I don't want to assume things. so it's always good to check it even if it's there always

lordofthejars: but this is not an assumption, if you are using `CubeAuxiliaryArchiveAppender` it means that `spi` is also there because it is a direct dependency. Also keep in mind that checking if a class is on classpath is not a free operation it takes some time to do it so it is something to take into consideration. But of course if needed it should be done, but this is not the case because it is direct.

Issue #544, opened 2017-01-09
Alternative media format for UI videos.


hemanik:
##### Issue Overview

Currently recorded VNC session is in flash (.flv) format which is no longer supported by HTML hence rendering the video as non playable in reports.
Conversion from flv to another format like mp4 needed for video support in HTML reports. 
 
##### Expected Behaviour

VNC session recording to have .mp4 media format that is supported by HTML.

##### Current Behaviour

VNC session recording is in flash media (.flv) format.

PR #545, opened 2017-01-09
#544: Adds supports for .mp4 media format for UI videos.

hemanik:

Adds supports for .mp4 media format for UI video, necessary to render videos playable in HTML reports. 

- Conversion from flash media format to mp4 format with introduction of new docker image.
- Adds implementation for creation of the new image instance. 
- Updates video linkage in reports to converted image. 

**Fixes**: #544


docker/drone/src/main/java/org/arquillian/cube/docker/drone/SeleniumContainers.java
```diff
@@ -75,13 +84,47 @@ public String getVncContainerName() {
...
+
+
+        CubeContainer cubeContainer = new CubeContainer();
+        cubeContainer.setImage(Image.valueOf(CONVERSION_IMAGE));
+
+        cubeContainer.setBinds(
+                Arrays.asList(dockerVolume.toAbsolutePath().toString() + ":/" + VOLUME_DIR + ":rw")
+        );
+
+        //final Link link = Link.valueOf(VNC_HOSTNAME + ":" + CONVERSION_CONTAINER_NAME);
```
lordofthejars: Can you remove all the comments? They are not adding any value to the code.

docker/drone/src/main/java/org/arquillian/cube/docker/drone/event/AfterConversion.java
```diff
@@ -0,0 +1,11 @@
+package org.arquillian.cube.docker.drone.event;
+
+import org.jboss.arquillian.core.api.Event;
+
+public class AfterConversion implements Event {
```
lordofthejars: You need to implement `org.jboss.arquillian.core.spi.event.Event` not `org.jboss.arquillian.core.api.Event`

docker/drone/src/main/java/org/arquillian/cube/docker/drone/event/AfterConversion.java
```diff
@@ -0,0 +1,11 @@
...
+
+    @Override
+    public void fire(Object o) {
```
lordofthejars: Then you'll be able to remove this empty method.

lordofthejars:
From my side, really good work, some minimal changes to be done, but overall looks pretty good.

docker/flv2mp4-docker-image/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM centos:7
+
+MAINTAINER Hemani Katyal <hkatyal@redhat.com>
```
lordofthejars: Really happy to see this line (MAINTAINER ones), you've been abducted by Arquillian 😆

docker/flv2mp4-docker-image/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+
+RUN yum install epel-release -y && yum update -y && \ 
+    rpm --import http://li.nux.ro/download/nux/RPM-GPG-KEY-nux.ro && \
+    rpm -Uvh http://li.nux.ro/download/nux/dextop/el7/x86_64/nux-dextop-release-0-5.el7.nux.noarch.rpm && \
+    yum install ffmpeg ffmpeg-devel -y
+
+ADD flv2mp4.sh /usr/local/bin/
```
bartoszmajsak: I think more reproducible way (or easier to rebuild) would be to `wget` or `curl` this file from a location using the same tag as the image itself - `http://...../../.../1.0.0.Alpha16/flv2mp4.sh` or sth similar. WDYT @lordofthejars?

About the docker image itself, there are few questions we should address before releasing anything:

  * Should we keep docker file in the same repo as cube or create another one, similar to [jboss-dockerfiles](https://github.com/jboss-dockerfiles)?
  * How should we version/tag images? Should we follow versioning alongside Cube? So the first published image would be `1.0.0.Alpha16`

hemanik: So what should be done with the Dockerfile ?

docker/drone/src/main/java/org/arquillian/cube/docker/drone/SeleniumContainers.java
```diff
@@ -25,6 +28,7 @@
     private static final int VNC_EXPOSED_PORT = 5900;
     public static final String SELENIUM_CONTAINER_NAME = "browser";
     public static final String VNC_CONTAINER_NAME = "vnc";
+    public static final String CONVERSION_CONTAINER_NAME = "flv2mp4";
```
lordofthejars: True, probably releasing the docker image should be a priority in before merging.

docker/drone/src/main/java/org/arquillian/cube/docker/drone/VideoConversionLifecycleManager.java
```diff
@@ -0,0 +1,46 @@
...
+import org.arquillian.cube.spi.Cube;
+import org.arquillian.cube.spi.CubeRegistry;
+import org.jboss.arquillian.core.api.Event;
+import org.jboss.arquillian.core.api.annotation.Inject;
+import org.jboss.arquillian.core.api.annotation.Observes;
+import org.jboss.arquillian.test.spi.event.suite.AfterSuite;
+
+public class VideoConversionLifecycleManager {
+
+    Cube flv2mp4;
```
lordofthejars: Probably because in case of testing it, you can inject the mocked instance.

docker/flv2mp4-docker-image/Dockerfile
lordofthejars: I would leave Dockerfile inside this project, so anyone can see how it is created instead of having to navigate through different projects. 

Also I think that lifecycle of the docker image should be different from project, since we could use in other projects in the future, but not strong opinion.

And yes doing the wget of a concrete file might be better to have a reproducible way. The only thing is that then first we need to tag the version then modify the Dockerfile and create the image.

bartoszmajsak: > And yes doing the wget of a concrete file might be better to have a reproducible way. The only thing is that then first we need to tag the version then modify the Dockerfile and create the image.

or we refer to has.

lordofthejars: ok let's do it in that way. @hemanik can you work on this? And then release the docker image? @bartoszmajsak do we have arquillian organization there?

bartoszmajsak: Yes, we do. Please create an account and I will add you.

Issue #548, opened 2017-01-12
docker-compose Functional tests are failing due to `Unsupported major.minor version 52.0`


dipak-pawar:
##### Issue Overview
docker-compose v1 /  docker-compose v2 Functional tests are failing due to `Unsupported major.minor version 52.0` for tomcat.

##### Expected Behaviour
All functional tests should pass.

##### Current Behaviour
functional tests are failing.

##### Steps To Reproduce
1. From the root of project run `mvn test -pl docker/ftest-docker-compose-v2`

PR #549, opened 2017-01-12
Fixed failing functional tests for compose.

dipak-pawar:
Fixed failing tests for docker-compose.

- tomcat docker image is using jre8

**Fixes**: #548

### DIFF
--- a/docker/ftest-docker-compose-v2/docker-compose.yml
+++ b/docker/ftest-docker-compose-v2/docker-compose.yml
@@ -2,7 +2,7 @@ version: '2'
 services:
   tomcat:
     env_file: envs
-    image: tutum/tomcat:7.0
+    build: src/test/resources/tomcat
     ports:
       - "8089:8089"
       - "8088:8088"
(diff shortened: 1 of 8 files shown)


docker/ftest-docker-compose-v2/docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: '2'
 services:
   tomcat:
     env_file: envs
-    image: tutum/tomcat:7.0
```
lordofthejars: why not just use tomcat:jre8 directly here?

dipak-pawar: We have to add users explicitly by adding `ADD tomcat-users.xml conf/` in dockerfile, which is not provided in base image. ref - https://hub.docker.com/_/tomcat/

lordofthejars: ah right, now I understand the problem.

lordofthejars:
Good work, thank you so much.

PR #550, opened 2017-01-16
removed unwanted env property.

dipak-pawar:
Removed unwanted env property.

- Removed unwanted env property after changing docker image to dockerhub.


docker/ftest-docker-compose-v2/envs
```diff
@@ -1,3 +1 @@
-TOMCAT_PASS=mypass
-CATALINA_OPTS=-Djava.security.egd=file:/dev/./urandom
```
lordofthejars: Don't remove this line of CATALINA_OPTS, this is important to control entropy of securerandom classes generated during start up of Tomcat. Removing it causes that your tests might run really slowly randomly. The cause would be that entropy takes 2 minutes to be calculated. 

With this attribute you make Tomcat to calculate entropy using different algorithm (not based on hw attributes) that makes everything works better.

Issue #553, opened 2017-01-18
Provide URL enrichment for standalone mode


lordofthejars:
##### Issue Overview

Currently when running Arquillian Cube in standalone mode, you have no way to inject a URL of kind `dockerhost:portbinding/context`, you need to construct yourself by enriching test with required data and then construct the URL.

##### Expected Behaviour

```java
@DockerUrl(protocol="http", containerName = "helloworld", exposedPort = "8080", context = "/hello")
@ArquillianResource
URL url;

// Then URL would be the docker host IP + binding port of exposed port 8080 of container helloworld + context

```

##### Current Behaviour

```java
@HostIP
String ip;

@HostPort(containerName = "tomcat", value = 8080)
int tomcatPort;

@Test
public void test() {
  new URL("http", ip, tomcatPort, "hello");
}
```

PR #556, opened 2017-01-19
adds @DockerUrl annotation enrichment in URLs

lordofthejars:
##### Issue Overview

Currently when running Arquillian Cube in standalone mode, you have no way to inject a URL of kind `dockerhost:portbinding/context`, you need to construct yourself by enriching test with required data and then construct the URL.

##### Expected Behaviour

```java
@DockerUrl(protocol="http", containerName = "helloworld", exposedPort = "8080", context = "/hello")
@ArquillianResource
URL url;

// Then URL would be the docker host IP + binding port of exposed port 8080 of container helloworld + context

```

##### Current Behaviour

```java
@HostIP
String ip;

@HostPort(containerName = "tomcat", value = 8080)
int tomcatPort;

@Test
public void test() {
  new URL("http", ip, tomcatPort, "hello");
}
```
**Fixes**: #553

### DIFF
--- a/docker/ftest-standalone-autostart/src/test/java/org/arquillian/cube/StandaloneTestCase.java
+++ b/docker/ftest-standalone-autostart/src/test/java/org/arquillian/cube/StandaloneTestCase.java
@@ -4,6 +4,7 @@ import org.arquillian.cube.docker.impl.requirement.RequiresDockerMachine;
 import org.arquillian.cube.requirement.ArquillianConditionalRunner;
 import org.hamcrest.CoreMatchers;
 import org.jboss.arquillian.junit.InSequence;
+import org.jboss.arquillian.test.api.ArquillianResource;
 import org.junit.Assert;
 import org.junit.Test;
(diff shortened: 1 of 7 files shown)


docker/ftest-standalone-autostart/src/test/java/org/arquillian/cube/StandaloneTestCase.java
```diff
@@ -36,6 +43,15 @@ public void shouldBeAbleToCreateAndStart() throws IOException {
         assertThat(pong, containsString("OK"));
     }
 
+    @Test
+    @InSequence(2)
+    public void should_be_able_to_inject_url_in_standalone() {
+        assertThat(url, is(notNullValue()));
+        assertThat(url.getProtocol(), is("http"));
+        assertThat(url.getHost(), is(ip));
+        assertThat(url.getPort(), is(8080));
```
bartoszmajsak: Would also good to cover port binding where binding port is different.

Issue #555, opened 2017-01-19
Create Environment variables for Container objects


lordofthejars:
##### Issue Overview

If you want to set environment variables in containers defined in a container object you can only do it by creating a custom Dockerfile.

Create an `@Environment` annotation that can be repeated to set them.

##### Expected Behaviour

```java
@Cube(“tomme”)
@Image("tomee:8-jre-1.7.2-webprofile")
@Environment("A=a")
@Environment("B=b")
public static class ImageContainerObject {
}
```

PR #557, opened 2017-01-19
adds environment variable support in container objects

lordofthejars:

Adds `@Environment` annotation support for Container objects

- Creates an environment annotation to set environment variables on running container
-
-

**Fixes**: #555

### DIFF
--- a/docker/docker/src/test/java/org/arquillian/cube/docker/impl/client/containerobject/CubeContainerObjectTestEnricherTest.java
+++ b/docker/docker/src/test/java/org/arquillian/cube/docker/impl/client/containerobject/CubeContainerObjectTestEnricherTest.java
@@ -2,6 +2,7 @@ package org.arquillian.cube.docker.impl.client.containerobject;
 
 
 import static org.hamcrest.CoreMatchers.hasItem;
+import static org.hamcrest.CoreMatchers.hasItems;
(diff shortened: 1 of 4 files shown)


docker/docker/src/test/java/org/arquillian/cube/docker/impl/client/containerobject/CubeContainerObjectTestEnricherTest.java
```diff
@@ -392,6 +450,12 @@ public TestContainerObject() {
     public static class ImageContainerObject {
     }
 
+    @Image("tomee:8-jre-1.7.2-webprofile")
+    @Environment("a=b")
+    @Environment("c=d")
```
bartoszmajsak: How can this possibly work? Shouldn't it be wrapped with `@Environments`?

lordofthejars: It is Java ( new feature :)

bartoszmajsak: I'm so outdated ;)

bartoszmajsak:
LGTM

Issue #572, opened 2017-02-06
Add AssertJ matcher for current running processes


lordofthejars:
##### Issue Overview

Add AssertJ matcher for current running processes
Code of assertions is placed athttps://github.com/arquillian/arquillian-cube/tree/master/docker/assertj

##### Expected Behaviour

You can assert on current running processes doing `assertThat(docker).container("mycontainer").hasProcessRunning("name of process")`

dipak-pawar:
@lordofthejars Just wondering what will be the use case of this assertion scenario?

Does this name imply only name of process OR it also includes PID?

lordofthejars:
The idea is that user can validate its Dockerfile

dipak-pawar:
could you please give me some example where you can give process name in dockerfile which start the process by same name inside container?

lordofthejars:
well in Dockerfile you can do `CMD /usr/local/tomcat/catalina.sh start` then I want to validate that tomcat is really started. In this case it is pretty obvious but imagine that a container are running some daemon processes and you want to check that they are really working.

After that we should implement another assertion which is `canExecute` but this is another thing.

dipak-pawar:
I think instead of checking process running by it's name, we should check port is exposed by running telnet or some other networking tools. 

I am not sure about process name & it's naming convention strategy. I have seen but I didn't find process name for any running process.

lordofthejars:
exposed port is already checked in assertj. The process in case of Tomcat is for example `java` right?

dipak-pawar:
sure. but it might be possible at the same time you will have two java process inside container. what about adding n with `withSize()` methos to it? Something like

`assertThat(docker).container("mycontainer").hasProcessRunning("name of process").withSize(2)`

lordofthejars:
yes in theory this should not happen since one container should be one process. But yes maybe you are right about doing in that way.

dipak-pawar:
nopes. theory is correct. I realize it later.

lordofthejars:
:)

dipak-pawar:
@lordofthejars above situation can be possible in the case of systemd. I mean if we are starting systemd  which is starting some other processes.

PR #573, opened 2017-02-07
Adds AssertJ Matcher hasProcessRunning.

dipak-pawar:
Adds assertJ matcher for container processes.

**Fixes**: #572

### DIFF
--- a/docker/assertj/src/main/java/com/github/dockerjava/assertions/ContainerAssert.java
+++ b/docker/assertj/src/main/java/com/github/dockerjava/assertions/ContainerAssert.java
@@ -7,12 +7,9 @@ import com.github.dockerjava.api.model.HostConfig;
 import org.assertj.core.api.AbstractAssert;
 import org.assertj.core.util.Objects;
 
-import java.io.BufferedReader;
-import java.io.InputStreamReader;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
-import java.util.stream.Collectors;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
@@ -39,24 +36,6 @@ public class ContainerAssert extends AbstractAssert<ContainerAssert, InspectCont
       assertThat(getContainerConfig().getExposedPorts())
               .overridingErrorMessage("%nExpecting:%n <%s>%nto contain:%n <%s>", getContainerConfig().getExposedPorts(), Arrays.asList(ports))
               .contains(myports);
-
-      return this;
-   }
-
-   public ContainerAssert hasProcessRunning(String process) {
-      isNotNull();
-
-      InspectContainerResponse.ContainerState state = getContainerState();
-      ContainerStateAssert stateAssert = new ContainerStateAssert(state);
-      stateAssert.isNotNull();
-
-      this.isRunning();
-
-      final String actualProcessName = getProcessName(state.getPid());
-
-      assertThat(actualProcessName).isNotNull();
-      assertThat(actualProcessName).isNotEmpty();
-      assertThat(actualProcessName).isEqualToIgnoringCase(process);
 
       return this;
    }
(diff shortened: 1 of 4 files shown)


docker/assertj/src/main/java/com/github/dockerjava/assertions/ContainerAssert.java
```diff
@@ -175,4 +196,23 @@ private ContainerConfig getContainerConfig() {
       return this.actual.getConfig();
    }
 
+    private static String getProcessName(int processId) {
```
lordofthejars: I was thinking in inspect the output of https://docs.docker.com/engine/api/v1.23/#/containers section List processes running inside a container which I have already implemented in DockerClient. So using the REST API rather than doing it using `Runtime`. Also notice that this would allows you to write a test.

docker/assertj/src/main/java/com/github/dockerjava/assertions/DockerJavaAssert.java
```diff
@@ -31,6 +35,12 @@ public ContainerAssert container(String name) {
       return new ContainerAssert(container);
    }
 
+   public CubeOutputAssert withContainer(String name) {
+         String[] cmd = {"bash", "-c", "for i in $(ps -axo pid --no-headers); do echo `ps -p $i -o comm=`; done"};
```
lordofthejars: How about using this method https://github.com/arquillian/arquillian-cube/blob/master/docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/DockerClientExecutor.java#L813 it is a pure API solution and we avoid two things, the first is having to do 2 REST commands and the second one and more important, not all containers has installed `bash`. So we avoid being incompatible with these ones.

dipak-pawar: It is giving only process id with command executed

bartoszmajsak: On top of that.... Windows :P

dipak-pawar: we are executing it inside container. not on windows @bartoszmajsak

lordofthejars: Well I see that actually it gets `"USER","PID","%CPU","%MEM","VSZ","RSS","TTY","STAT","START","TIME","COMMAND"`

Could we parse the `COMMAND` to see the exact name?

I think that if this is not enough for us, maybe we should drop the support because of `bash`thing. Or we just release with this limitation.

bartoszmajsak: https://hub.docker.com/r/microsoft/iis/

giphy 4

lordofthejars: So what we do then? Is it possible to parse the command just something simple like `containerString` or it is not possible?

dipak-pawar: @lordofthejars I think `sh` is better than `bash`. It'll have only problem with windows.
 WDYT?

dipak-pawar: > Could we parse the COMMAND to see the exact name?

No we can't get exact name from COMMAND

dipak-pawar: > Is it possible to parse the command just something simple like containerString?

I think it's not possible

lordofthejars: ok then let's do it with sh and just merge it :) thanks

docker/assertj/src/main/java/com/github/dockerjava/assertions/CubeOutputAssert.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    public CubeOutputAssert hasProcessesRunning(String... processes) {
+        isNotNull();
+
+        assertThat(this.actual.getError()).isEmpty();
+        assertThat(this.actual.getOutput()).isNotNull();
+        assertThat(this.actual.getOutput()).isNotEmpty();
+
+        assertThat(this.actual.getOutput()).contains(processes);
```
bartoszmajsak: I see I'm a bit too late to the party, but... if we use multiple assertions within one I would suggest wrapping it in `SoftAssertion` from `assertj` instead. Will give better information of the overall state. Could we still improve it?

lordofthejars: I can revert

bartoszmajsak: Does not make sense. Incremental improvements :)

Issue #564, opened 2017-01-29
Allow creating cubes from container objects programatically


rivasdiaz:
##### Issue Overview

With CubeController we can create/destroy cubes that are pre-configured in arquillian.xml. There is no API to create/destroy cubes defined as container objects.

##### Expected Behaviour

I would like to be able to write code like this:

```
@ArquillianResource
CubeController cubeController;

public testBlah() {
    MyContainerObject myContainer = cubeController.start(MyContainerObject.class);
   try { /* ... */ } finally { cubeController.destroy(myContainer); }
}
```
And have a docker container created and destroyed from the container object definition.

##### Current Behaviour

There is no API for creating container objects.

rivasdiaz:
I think I can work on implementing this feature, but I would like input on how to expose the API. Extending CubeController seems like the logical option for me, but this interface is also implemented by other Cube implementations (other than Docker), while container objects seems to be Docker only. And I haven't seen any public API specific to Container Objects.

lordofthejars:
Currently `CubeController` is only implemented in Docker, so the interface can be updated with methods:

```java
public void register(Class<?> containerObject);
```

There are two implementations because one is used when using as client test mode and the other one when it is used in a test running in container. But since container objects are only supported in client mode, in case of `ContainerCubeController` implementation you can just throw an `UnsupportedOperationException`

The important thing here that I am not sure how to resolve is that you need to move this logic https://github.com/arquillian/arquillian-cube/blob/master/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/containerobject/CubeContainerObjectTestEnricher.java which is docker project into `CubeController` which is in core project, and you don't have access there for these classes.

Extending it it is an option the problem is that then you have two different CubeControllers on Arquillian registry which is a problem to inject it on test.

So in this sense I am not sure how to maintain the abstraction logic.

rivasdiaz:
I was working on possible solution by defining a new interface like this:

```java
public interface ContainerObjectFactory {
    <T> T createContainerObject(
            String cubeId, Class<T> containerObjectClass, Object containerObjectContainer);
}
```
Where:
```cubeId``` would allow (if not null) to specify/override the name of the container.
```containerObjectClass``` is the Container Object class; and
```containerObjectContainer``` (if not null) would allow registering the container to that class so if it is a test class, it can be stopped automatically at the end of the test.

[...]

PR #574, opened 2017-02-07
Allows creating cubes from container objects programmatically

rivasdiaz:

Allows creating cubes from container objects programmatically

- A new interface injectable inside Arquillian test classes: ```ContainerObjectFactory```
- support for programmatically creating container objects with ```factory.createContainerObject(...)```. Created container objects will work exactly as if they were created by injecting them as fields. Associated Cube instances can also be stopped/destroyed using an injected ```CubeController```

**Fixes**: #564

### DIFF
--- a/api/src/main/java/org/arquillian/cube/ContainerObjectFactory.java
+++ b/api/src/main/java/org/arquillian/cube/ContainerObjectFactory.java
@@ -1,0 +1,32 @@
+package org.arquillian.cube;
+
+/**
+ * Factory to instantiate container objects
+ *
+ * @author <a href="mailto:rivasdiaz@gmail.com">Ramon Rivas</a>
+ */
+public interface ContainerObjectFactory {
+
+    /**
+     * Creates an instance of the container object. It also creates and starts a cube defined by the container object
+     * class. Some configuration can be overridden by passing an additional ContainerObjectConfiguration instance.
+     *
+     * @param containerObjectClass type of the container object to instantiate. Required
+     * @param configuration if not null, allows specifying some configuration parameters. Optional (can be null)
+     * @param <T> type of the container object
+     * @return the newly created container object
+     */
+    <T> T createContainerObject(Class<T> containerObjectClass, ContainerObjectConfiguration configuration);
+
+    /**
+     * Creates an instance of the container object. It also creates and starts a cube defined by the container object
+     * class. Some configuration can be overridden by passing an additional ContainerObjectConfiguration instance.
+     *
+     * @param containerObjectClass type of the container object to instantiate. Required
+     * @param configuration if not null, allows specifying some configuration parameters. Optional (can be null)
+     * @param containerObjectContainer if not null, marks this object as the container of the created container object. Optional (can be null)
+     * @param <T> type of the container object
+     * @return the newly created container object
+     */
+    <T> T createContainerObject(Class<T> containerObjectClass, ContainerObjectConfiguration configuration, Object containerObjectContainer);
+}
--- a/api/src/main/java/org/arquillian/cube/ContainerObjectFactory.java
+++ b/api/src/main/java/org/arquillian/cube/ContainerObjectFactory.java
@@ -9,10 +9,20 @@ public interface ContainerObjectFactory {
 
     /**
      * Creates an instance of the container object. It also creates and starts a cube defined by the container object
+     * class.
+     *
+     * @param containerObjectClass type of the container object to instantiate
+     * @param <T> type of the container object
+     * @return the newly created container object
+     */
+    <T> T createContainerObject(Class<T> containerObjectClass);
+
+    /**
+     * Creates an instance of the container object. It also creates and starts a cube defined by the container object
      * class. Some configuration can be overridden by passing an additional ContainerObjectConfiguration instance.
      *
-     * @param containerObjectClass type of the container object to instantiate. Required
-     * @param configuration if not null, allows specifying some configuration parameters. Optional (can be null)
+     * @param containerObjectClass type of the container object to instantiate
+     * @param configuration allows specifying some configuration parameters
      * @param <T> type of the container object
      * @return the newly created container object
      */
@@ -22,9 +32,9 @@ public interface ContainerObjectFactory {
      * Creates an instance of the container object. It also creates and starts a cube defined by the container object
      * class. Some configuration can be overridden by passing an additional ContainerObjectConfiguration instance.
      *
-     * @param containerObjectClass type of the container object to instantiate. Required
-     * @param configuration if not null, allows specifying some configuration parameters. Optional (can be null)
-     * @param containerObjectContainer if not null, marks this object as the container of the created container object. Optional (can be null)
+     * @param containerObjectClass type of the container object to instantiate
+     * @param configuration if not null, allows specifying some configuration parameters
+     * @param containerObjectContainer marks this object as the container of the created container object
      * @param <T> type of the container object
      * @return the newly created container object
      */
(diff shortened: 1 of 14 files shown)


rivasdiaz:
Some implementation details of this pull request:

It contains a big refactoring of ```CubeContainerObjectTestEnricher``` class. That class was originally in charge of:
- extracting information from the container object class
- extracting information from the field that would be initialized with the container object
- instantiating the container object
- creating/starting the associated ```Cube```.
- calling other ```TestEnricher``` instances on the created object and some built-in support for other annotations (```@CubeIp``` and ```@HostPort```)

After refactoring, everything related to the container object is handled by ```DockerContainerObjectBuilder``` class:
- extracting information from the container object class
- instantiating the container object
- creating/starting the associated ```Cube```.
- calling any registered ```TestEnricher``` instances (including ```CubeContainerObjectTestEnricher```) on the created object and some built-in support for other annotations (```@CubeIp``` and ```@HostPort```)

The class ```CubeContainerObjectTestEnricher``` now uses ```DockerContainerObjectFactory``` to create container objects. It still handles metadata extraction from fields annotated with ```@Cube``` and passes that information to the container object factory.

Creating container objects with the introduced ```DockerContainerObjectFactory``` is more powerful than ```CubeContainerObjectTestEnricher``` as it can receive any configuration. It enables, for example, the use of any ```Await``` strategy, something not currently supported with annotations.

All test cases that existed for ```CubeContainerObjectTestEnricher``` now depend on an injected instance of ```DockerContainerObjectFactory``` that does half of the work.

There is a new unit test class for ```DockerContainerObjectBuilder``` that contains an equivalent test case for each of the existing test cases in ```CubeContainerObjectTestEnricher```. I think that eventually the existing tests in ```CubeContainerObjectTestEnricher``` should be removed and simpler tests added, as the difference among test cases is now handled outside of the class being tested.

There are 2 more test cases in ```DockerContainerObjectBuilder``` for the built-in support for ```@CubeIp``` and ```@HostPort```. Those didn't exist in ```CubeContainerObjectTestEnricher```.

rivasdiaz:
cc @lordofthejars

bartoszmajsak:
Awesome stuff! Thanks a lot @rivasdiaz

docker/docker/src/test/java/org/arquillian/cube/docker/impl/client/containerobject/CubeContainerObjectTestEnricherTest.java
```diff
@@ -13,10 +12,14 @@
...
 
 import java.io.File;
-import java.io.FilenameFilter;
-import java.util.ArrayList;
+import java.io.IOException;
+import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
+
+import com.google.common.base.StandardSystemProperty;
```
lordofthejars: I would not suggest to use Guava here. The main idea is that this dependency is transitive, this means that maybe in the future is not bundled anymore because a third library just removes its usage.

Also I would not add it as "real" dependency on project because it is a 3MB library for just using one class.

docker/docker/src/test/java/org/arquillian/cube/docker/impl/client/containerobject/DockerContainerObjectBuilderTest.java
```diff
@@ -0,0 +1,499 @@
+package org.arquillian.cube.docker.impl.client.containerobject;
+
+import static org.hamcrest.CoreMatchers.*;
+import static org.hamcrest.core.Is.is;
+import static org.junit.Assert.*;
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.*;
```
lordofthejars: We tend to avoid start operator, I think that some IDEs do this automatically. No strong opinion here, so if you change ok if not is ok too.

docker/docker/src/test/java/org/arquillian/cube/docker/impl/client/containerobject/DockerContainerObjectBuilderTest.java
```diff
@@ -0,0 +1,499 @@
...
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.security.AccessController;
+import java.security.PrivilegedExceptionAction;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicReference;
+
+import com.google.common.base.StandardSystemProperty;
```
lordofthejars: Same problem related before CubeContainerObjectTestEnricherTest.java about Guava.

lordofthejars:
Also if you can update docs with this new enrichment (in Container Object section) would be awesome

rivasdiaz:
Thanks for the review @lordofthejars. I will be really busy the next three days. I'll try to work on your comments as soon as possible and update the pull request.

rivasdiaz:
cc @lordofthejars 
I finally completed all requested changes.

While reviewing the current documentation for Container Objects, I decided to change some of the wording on other sections to clarify its meaning. Hopefully as this PR touches a lot of areas related to container objects, those changes are OK too. Please let me know if you prefer me to only make a small change to the doc related to this PR, and I will rollback my other changes.

Also let me know if you have any other comments with the PR and I will do my best to try to fix them.

api/src/main/java/org/arquillian/cube/ContainerObjectFactory.java
```diff
@@ -0,0 +1,32 @@
...
+ * @author <a href="mailto:rivasdiaz@gmail.com">Ramon Rivas</a>
+ */
+public interface ContainerObjectFactory {
+
+    /**
+     * Creates an instance of the container object. It also creates and starts a cube defined by the container object
+     * class. Some configuration can be overridden by passing an additional ContainerObjectConfiguration instance.
+     *
+     * @param containerObjectClass type of the container object to instantiate. Required
+     * @param configuration if not null, allows specifying some configuration parameters. Optional (can be null)
```
lordofthejars: I have noticed this now, I don't like that users can add `null` value in public API. Of course there are some exceptions, but I usually prefer in public APIs to create a `Nullified` object or just create a new method. So for example:

`<T> T createContainerObject(Class<T> containerObjectClass)` and then if you want internally you call `this.createContainerObject(containerObjectClass, null)`

Another approach might be the nullified object which `ContainerObjectConfiguration` class has an static method called `defaults()` which returns an empty object and then inside the method again you can do something like:

```
if (configuration == ContainerObjectConfiguration.defaults()) {
  this.createContainerObject(containerObjectClass, null)
}
```

What I am trying to do is that the API does not expose the null in public API.

lordofthejars:
Final thing that I always miss as well, update docs after a change in the API :) but apart from this we are ready :)

rivasdiaz:
cc @lordofthejars 
Added a new method on the interface for clients to call if they don't want to pass any configuration.
Internally it passes an empty configuration to the overloaded method that expects a configuration. I also added a few validations that expect non null parameters.

lordofthejars:
Yes I have seen my point (and sorry to be a bit PITA) is to update this line https://github.com/arquillian/arquillian-cube/pull/574/files#diff-d2bcbb506a1f4067aa20b24e8c78f97cR254 to new method.

rivasdiaz:
I'm sorry for that. If that sample code were in a real test, it would have thrown an exception. The doc is fixed in my last commit. Unfortunately docs are not compiled :smile: 

Let me know if you see anything else. I will be happy to try to fix it

lordofthejars:
Nothing to forgive, this happens to best families. About compiling docs it is possible to do it with Asciidoctor and using `include` keyword, so you point directly to real test and you render docs in your compilation pipeline, so if there is a fail, then docs are not rendered. Sadly we cannot use this approach because of restrictions in Github.

Thank you very much for your help and your great work.

Alex.

PR #578, opened 2017-02-10
Adds minor improvement to Custom AssertJ Matcher.

dipak-pawar:
Adds SoftAssertions & custom messages to assertJ


lordofthejars:
Maybe @bartoszmajsak has a preference on soft assertions using JUnit `Rule`, but this is the only thing I see.

bartoszmajsak:
> using JUnit Rule

This would lock us with Junit, so no, I don't prefer this option :)

bartoszmajsak:
Thanks Dipak!

PR #579, opened 2017-02-10
Add resource providers for Deployments, ReplicaSet and DeploymentConf…

iocanel:
This pr adds additional resource providers for core k8s and openshift resources.


lordofthejars:
Thank you so much.

PR #583, opened 2017-02-15
Changed tomcat 7 image to use jre8.

dipak-pawar:
Functional tests are failing when we try to run it  tutum/tomcat:7.0 from docker/ftest.
It is skipping on travis as it needs docker-machine to run. So we couldn't analyse this before.

- Removed all unnecessary tomcat configuration.
- Adds tomcat:jre8 image

@lordofthejars Can you please check docker-machine ftests to execute on your machine?

### DIFF
--- a/docker/ftest-docker-compose-v2/src/test/resources/tomcat/Dockerfile
+++ b/docker/ftest-docker-compose-v2/src/test/resources/tomcat/Dockerfile
@@ -1,4 +1,4 @@
-FROM tomcat:jre8
+FROM tomcat:7-jre8
 
 ADD tomcat-users.xml conf/
 
--- a/docker/ftest-docker-compose-v2/src/test/resources/tomcat/Dockerfile
+++ b/docker/ftest-docker-compose-v2/src/test/resources/tomcat/Dockerfile
@@ -1,4 +1,4 @@
-FROM tomcat:7-jre8
+FROM tomcat:7.0.75-jre8
 
 ADD tomcat-users.xml conf/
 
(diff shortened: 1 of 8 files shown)


docker/ftest/src/test/resources/tomcat/Dockerfile
```diff
@@ -1,7 +1,6 @@
-FROM tutum/tomcat:7.0
-
+FROM tomcat:jre8
```
bartoszmajsak: I would suggest to lock it on the latest tag which we know works. Otherwise when `:jre8` tag gets updated we might run into issues of not working containers

dipak-pawar: @bartoszmajsak you mean tag this image & push to arquillian dokcerhub account & use it from there.

bartoszmajsak: This might be another option. With just `tomcat:jre8` I don't even know which tomcat it is - I would assume latest. So how about locking it on a version which we know our stuff works with? e.g. `tomcat:8.5.11-jre8` (just an example - not sure if that's the right version).

lordofthejars: @dipak-pawar it is not urgent, but when you have 1 minute, can you update with the comments that Bartosz did so we can merge the PR?

bartoszmajsak:
Let's be even more concrete here - how about `7.0.75-jre8`?

docker/ftest-docker-compose-v2/src/test/resources/tomcat/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM tomcat:jre8
+FROM tomcat:7-jre8
```
lordofthejars: I think @bartoszmajsak meant to use this tag `7.0.75-jre8` with full version.

PR #584, opened 2017-02-17
Allow creating immutable variations of the core kubernetes services, …

iocanel:
The kubernetes extension, is creating a shutdown hook, that it performs cleanup if the test is aborted.

This does not play nicely, mostly because all `Instance.get()` return null, during the shutdown hook _(like the DI framework is unavailable)_.

The only possible solution is to create and use immutable variations of the services that are DI free.

This pull request does exactly this.


lordofthejars:
Thank you so much, in Docker part we did something like https://github.com/arquillian/arquillian-cube/blob/master/core/src/main/java/org/arquillian/cube/impl/client/ForceStopDockerContainersShutdownHook.java

iocanel:
Yeah, the concept is that same really. 

In our case, we don't have cube mapping to namespaces and also we have deep nesting of services using DI.

lordofthejars:
Ok cool, thank you very much for this improvement. 
Some day, maybe in JBcnConf we could try to join both models so K8s part is also a part of CubeRegistry. But this is something we need to see in calm.

Issue #392, opened 2016-07-07
Allow parallel execution of tests


lordofthejars:
It is perfectly fine to execute in same docker host same containers to run different tests against them.

For this reason we need to indicate to container name that there is a part of the name that should be dynamically generated.

For example given:

```
tomcat:
  image: tutum/tomcat
```

implies that container instance is going to be called `tomcat` and this means that test cannot run in parallel in the same docker host since it will throw a conflict.

So it will be interesting do something like:

```
tomcat*:
  image: tutum/tomcat
```

The `*` indicates that before starting everything, the definition will be redefined to something like:

```
tomcat1:
  image: ....
```

Of course things get worst when you have links since you need to duplicate everything and set as environment variable the new DNS but it can be done without much problem.

Then if you do this and you run from IDE, nothing is going to happen, you will still have only one container, but if you run in `Gradle` or `Maven` and you run tests in parallel, you will see different instances of same container in Dockerhost.

Of course you could even replicate the business container and share the persistence container acorss all the instances.

bartoszmajsak:
One thing to keep in mind is to deal with port offsets somehow in a smart way, as external ports have to be unique.

pwoelfle:
In which way it is possible to inject the host port or url to the Arquillian test?
I have tried following, but `hostPort` and `containerUrl` return `0` and `null`:
```
@HostPort(containerName = "tomcat*", value = 8080)
int hostPort

@DockerUrl(containerName = "tomcat*", exposedPort = 8080)
@ArquillianResource
URL containerUrl
```
Did I missed something?

I think the problem is the equals-check in `org.arquillian.cube.impl.model.LocalCubeRegistry#getCube(java.lang.String)`. A simple solution could be something like:

```
    @Override
    public Cube<?> getCube(String cubeId) {
        for(Cube<?> cube : this.cubes) {
            if(isMatchingCubeId(cube.getId(), cubeId))) {
                return cube;
            }
        }
        return null;
    }

    private boolean isMatchingCubeId(String cubeId, String requestedCubeId){
        if(requestedCubeId.endsWith("*")){
            String starlessRequestedCubeId = requestedCubeId.substring(0, requestedCubeId.length() - 1);
            return cubeId.startsWith(starlessRequestedCubeId);
        }

        return cubeId.equals(requestedCubeId);
    }
```

[...]

PR #586, opened 2017-02-17
[ISSUE #392] Fixed: Using star operator with @HostPort and @DockerUrl

pwoelfle:
As reported in [ISSUE 392](https://github.com/arquillian/arquillian-cube/issues/392), it is not possible to use `@HostPort` and `@DockerUrl` when using the star operator.

**Fixes**: #392

### DIFF
--- a/core/src/main/java/org/arquillian/cube/impl/model/CubeId.java
+++ b/core/src/main/java/org/arquillian/cube/impl/model/CubeId.java
@@ -1,0 +1,9 @@
+package org.arquillian.cube.impl.model;
+
+public interface CubeId {
+
+    String getId();
+
+    boolean isMatching(CubeId other);
+
+}
--- a/core/src/main/java/org/arquillian/cube/impl/model/CubeIdFactory.java
+++ b/core/src/main/java/org/arquillian/cube/impl/model/CubeIdFactory.java
@@ -1,0 +1,26 @@
+package org.arquillian.cube.impl.model;
+
+import java.util.Objects;
+
+public class CubeIdFactory {
+    private static final CubeIdFactory INSTANCE = new CubeIdFactory();
+
+    public static CubeIdFactory get() {
+        return INSTANCE;
+    }
+
+    public CubeId create(String id) {
+        Objects.requireNonNull(id, "Id must not be null.");
+        final CubeId cubeId;
+
+        if (id.matches(StarredCubeId.PATTERN)) {
+            cubeId = new StarredCubeId(id);
+        } else if (id.matches(ParallelizedCubeId.PATTERN)) {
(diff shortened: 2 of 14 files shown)


lordofthejars:
Overall the PR is really awesome and fixes in a clean way the problem you mention, but I have got a deep question. Suppose we have a Docker Host with same container started in parallel with id (`tomcat_1234` and `tomcat_5678`). Then you do `@HostPort(containerName = "tomcat*", value = 8080)` and here it is injected the first one found. Is it right to have it in that way? I suppose that from the point of view of test it has not much sense which one is taken, but if it is always returned the same (first) then you are not using all the resources. But of course maybe I am missing something.

pwoelfle:
Using Gradle, the parallel test execution is realized by splitting the tests in separate JVM processes. But within each process only one test is execute per time [1]. So only one `tomcat_[UUID]` should be present on execution of a test, because `LocalCubeRegistry` instance is per JVM process.

I'm not very familiar with Maven, but it seems that Maven allow parallel execution on method level. 
I will verify the behaviour and will post my results here.

[1] https://docs.gradle.org/current/userguide/java_plugin.html#sec:test_execution
[2] http://maven.apache.org/surefire/maven-surefire-plugin/examples/fork-options-and-parallel-execution.html

lordofthejars:
@pwoelfle yes you are right in case of more than one JVM, but I know that in Maven you can use parallel in same jvm, but well you have also the option to fork the execution.

pwoelfle:
@lordofthejars 
If I understand the existing code right, then a starred container defined in an `arquillian.xml` is auto-started. After all tests finished the container is destroyed.
Only using container objects will destroy the container after class, won't it? (See `org.arquillian.cube.docker.impl.client.containerobject.AfterClassContainerObjectObserver`)

So there is only one container `tomcat_[UUID]` which is used by all tests. Even when forking the tests, then there is one container per JVM process, which is used for all tests of the forked JVM process. I tried this with simple tests. 

This means that the current solution of parallism only provides the possibility to run multiple builds of the same project. This is the case on CI servers, where multiple builds with different branches or commits are running the same time on the same node.

Parallism on test level would be a new feature in my opinion. There should be a pool of containers, which belongs to a starred container. The parallel executed tests could then exclusivly request a container and return the instance back to pool after finish. This feature would only make sense for Maven, because there multiple tests can run in parallel within the same JVM process.

Btw is there a way to force container restart for each test or test class?

lordofthejars:
@pwoelfle I agree with you, I think for now the solution is clear and it works. Maybe it has no sense at all for now implementing more complex things if no one requests because of an specific reason. For this reason I choose to merge the PR :)

As you said there is no way to fore a restart of a container per test class, but we could add this property as container property which by default is false.

PR #587, opened 2017-02-18
Openshift configuration now extends the kubernetes configuration. Ope…

iocanel:
Noways, the Openshift extension works on top of the Kubernetes extension.

However, their lifecycles are not aligned, so there is no synergy between them and also can be a source of issues.

This pull request further aligns the two extesions:

- Aligns configuration and clients
- Creates and starts `auto containers` within the environment created by the kubernetes extension.

So with this pull request both extension share the same mechanism of workspace management for both `containerless` and `in container` tests.


lordofthejars:
Thank you so much.

PR #600, opened 2017-02-24
Bump kubernetes-client to 2.1.0.

iocanel:
2.1.0 has some fixes related to readiness checks and more.


iocanel:
@lordofthejars: could we have an alpha release with this change?

lordofthejars:
Tomorrow morning I will do, so on Tue will be at maven central, I'll ping
you when done.

PR #611, opened 2017-02-28
Misc

iocanel:
Improvements and bug fixes for Kubernetes/Openshift.

### DIFF
--- a/openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/CubeOpenShiftConfiguration.java
+++ b/openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/CubeOpenShiftConfiguration.java
@@ -1,6 +1,5 @@
 package org.arquillian.cube.openshift.impl.client;
 
-import org.arquillian.cube.impl.util.Strings;
 import org.arquillian.cube.kubernetes.api.Configuration;
 import org.arquillian.cube.kubernetes.impl.DefaultConfiguration;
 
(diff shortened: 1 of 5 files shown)


dipak-pawar:
@iocanel Can we add some test/ftest to verify everything works fine?

iocanel:
I'll add a test for the resource installer. 
The rest should be covered by our existing tests.

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/CubeOpenShiftConfiguration.java
```diff
@@ -140,47 +135,4 @@ public static CubeOpenShiftConfiguration fromMap(Configuration c, Map<String, St
             }
         }
     }
-
-    public static CubeOpenShiftConfiguration fromMap(Map<String, String> map) {
```
dipak-pawar: If we remove this, can we override properties from openshift extension?

iocanel: Yes, the CubeOpenShiftConfigurator is using a piece of code like: 

     CubeOpenShiftConfiguration cubeConfiguration = CubeOpenShiftConfiguration.fromMap(configuration, properties);

dipak-pawar: but you are removing this method (fromMap from CubeOpenShiftConfiguration) completely which is used in `CubeOpenShiftConfigurator` and I don't see any new addition.

iocanel: There were two fromMap() methods.

- one with map arguments
- one with map and configuration arguments.

Since the first was no longer used (for a while) I removed that.

dipak-pawar: ohh now I got it.

Issue #597, opened 2017-02-23
Automatic Resolution of Docker elements


lordofthejars:
##### Issue Overview

Automatic Resolution of Docker files like `docker-compose` or `Dockerfile`

##### Expected Behaviour

If user *does not* set `dockerContainers` nor `dockerContainersFile/s` then it would be great that Arquillian Cube does next things:

1) search in `/docker-compose.y(a)ml` if found then it is used
2) if not then search in `/src/test/docker/docker-compose.y(a)ml`
3) if not `/src/test/resources/docker-compose.y(a)ml`
4) if not `/src/test/resources/docker/docker-compose.y(a)ml`

if it is not present in any of these locations search the same but in `src/main` instead of `src/test`

If again there is no `docker-compose` there, just search the same but for `Dockerfile`, and in case of finding it, get the EXPOSE port/s and create dynamically a `dockerContainers` expression building that container

If none of this things happens then an exception is thrown.

In this way we avoid having to create an `arquillian.xml` file all the time.

dipak-pawar:
@lordofthejars Idea looks good. But I would look for following locations.
1. project's current dir.
2. test resources dir -> recursively look for docker-compose.yml OR Dockerfile in all directories inside test resources.
3. main resources -> recursively look for docker-compose.yml OR Dockerfile in all directories inside test resources.

In this way we are not forcing user to put docker file inside test/resources/docker/Dockerfile.

User can put Dockerfile in any folder under resources.

lordofthejars:
@dipak-pawar I agree with you, the problem is that I have seen some people putting these files in `src/main/docker` which it is not a resources dir. SO we can do what you mention + `src/main(test)/docker`

dipak-pawar:
@lordofthejars I would assume priority for `src/main(test)/docker` as last.

lordofthejars:
Easy: https://twitter.com/alexsotob/status/834649977599193088

bartoszmajsak:
From my point of view we should look it up in the following way:

1. Project root folder
2. `src/{main,test}/docker`
3. `src/{main,test}/resources`

I doubt we will have files in more than one place listed above, but you never know, so in such a case I would report and suggest configuration (still taking precedence as a default).

MatousJobanek:
[...]

PR #612, opened 2017-03-01
Add default docker-compose locations

lordofthejars:

Add default docker-compose locations, so it is not necessary to specify in `arquillian.xml`

**Fixes**: #597


docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/DockerContainerDefinitionParser.java
```diff
@@ -98,36 +101,113 @@ public static DockerCompositions convert(String content, DefinitionFormat defini
...
+
+        return docker;
+
+    }
+
+    private static URI checkRoot(String filename) {
+        final Path rootPath = Paths.get(filename);
+        Path finalPath;
+
+        if ((finalPath = exists(rootPath)) != null) {
```
bartoszmajsak: I dont really like this approach of assignment within an `if` statement. Makes it somehow blurry and reminds me of my ANSI C times ;)

bartoszmajsak: Also method name is misleading - it suggests to return `bool` but returns the `Path` - can we rework this flow slightly?

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/DockerContainerDefinitionParser.java
```diff
@@ -98,36 +101,113 @@ public static DockerCompositions convert(String content, DefinitionFormat defini
...
+
+    private static URI checkDistributionDirectory(String filename) {
+        final Path rootPath = Paths.get("src", "distribution", filename);
+        Path finalPath;
+
+        if ((finalPath = exists(rootPath)) != null) {
+            return finalPath.toUri();
+        }
+
+        return null;
```
bartoszmajsak: Wouldn't `Optional` make it nicer?

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/DockerContainerDefinitionParser.java
```diff
@@ -98,36 +101,113 @@ public static DockerCompositions convert(String content, DefinitionFormat defini
...
+        if ((finalPath = exists(testPath)) != null) {
+            return finalPath.toUri();
+        } else {
+            final Path mainPath = Paths.get("src", "main", outterDirectory, filename);
+            if ((finalPath = exists(mainPath)) != null) {
+                return finalPath.toUri();
+            }
+        }
+
+        return null;
```
bartoszmajsak: Wouldn't `Optional` make it nicer?

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/DockerContainerDefinitionParser.java
```diff
@@ -98,36 +101,113 @@ public static DockerCompositions convert(String content, DefinitionFormat defini
...
+        if (Files.exists(ymlPath)) {
+            return ymlPath;
+        } else {
+            final Path yamlPath = fullpath.resolveSibling(fullpath.getFileName() + ".yaml");
+            if (Files.exists(yamlPath)) {
+                return yamlPath;
+            }
+        }
+        
+        return null;
```
bartoszmajsak: Wouldn't `Optional` make it nicer?

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/DockerContainerDefinitionParser.java
lordofthejars: I completely agree with you but it is really simple logic and I was not really sure to do something more complex like chain pattern. Any other approach for this simple case?

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/DockerContainerDefinitionParser.java
lordofthejars: Originally I had with optional but at the end since are internal methods I was not really sure about it. I think that Optional should be used for public methods, maybe for private ones has no sense at all.

lordofthejars:
@bartoszmajsak I think it is ready to merge

PR #616, opened 2017-03-02
Ns cleanup destroy

iocanel:
Originally, there was the concept of `namespace cleanup` and `namespace destroy`. During the move to `arquillian-cube` both concepts collapsed into `namespace destroy`.

However, there is a valid use case "Use a single namespace/project, don't delete but clean it up". In fact in restricted environment this is likely to be the most common use case, so we need to be able to handle it.

This pull request, fixes cleanup vs destroy so that we can selectively do both.


lordofthejars:
@iocanel ups https://travis-ci.org/arquillian/arquillian-cube/builds/207038301#L5965

iocanel:
Hmmm, I'll have a look later tonight.

lordofthejars:
Let me retest

iocanel:
@lordofthejars: it should be fine now.

lordofthejars:
Thank you very much.

PR #625, opened 2017-03-10
Changed definitionsFile to env.config.url

dipak-pawar:
Change property `definitionsFile` to `env.config.url` in openshift.

- We are not using definitionsFile after this change. User has to use `env.config.
- If no config file found, it should throw exception instead of failing to some other place. So it's easy for user to debug it.

### DIFF
--- a/kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/SessionManager.java
+++ b/kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/SessionManager.java
@@ -109,6 +109,8 @@ public class SessionManager implements SessionCreatedListener {
                         try (InputStream is = definitionsFileURL.openStream()) {
                             resources.addAll(resourceInstaller.install(definitionsFileURL));
                         }
+                    } else {
+                        throw new IllegalStateException("Did not find any k8s or openshift resources configuration file.");
                     }
                 }
 
--- a/openshift/ftest/src/test/resources/arquillian.xml
+++ b/openshift/ftest/src/test/resources/arquillian.xml
@@ -4,7 +4,7 @@
     xsi:schemaLocation="http://jboss.org/schema/arquillian http://jboss.org/schema/arquillian/arquillian_1_0.xsd">
 
 	<extension qualifier="openshift">
-		<property name="definitionsFile">src/test/resources/hello_pod.json</property>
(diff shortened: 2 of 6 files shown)


iocanel:
I am looking at it, please don't merge yet.

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/SessionManager.java
```diff
@@ -109,6 +109,8 @@ public void run() {
                         try (InputStream is = definitionsFileURL.openStream()) {
                             resources.addAll(resourceInstaller.install(definitionsFileURL));
                         }
+                    } else {
+                        throw new IllegalStateException("Did not find any k8s or openshift resources configuration file.");
```
iocanel: I can see cases, where the user will not use an `env.config.url` at all (e.g. run tests against a fixed namespace, or just create a namespace to run plain `in container` tests).

openshift/ftest/src/test/resources/arquillian.xml
```diff
@@ -4,7 +4,7 @@
     xsi:schemaLocation="http://jboss.org/schema/arquillian http://jboss.org/schema/arquillian/arquillian_1_0.xsd">
 
 	<extension qualifier="openshift">
-		<property name="definitionsFile">src/test/resources/hello_pod.json</property>
+		<property name="env.config.url">src/test/resources/hello_pod.json</property>
```
iocanel: This is an example of a case, where we want to use `definitionsFile` and `env.config.url` separate. If we collapse them into a signle property, then we loose all the cubecontroller stuff. 

So I think that my idea about collapsing them, was not so good. Sorry for that.

lordofthejars: So we can close without merge this PR?

iocanel: @lordofthejars: I'd like @dipak-pawar to confirm first, in case I miss something. 

I tried to address #626 with #627. If that makes sense, then we could close this one.

lordofthejars: Ok then I wait for @dipak-pawar to decide if we close it or not. For now the other PR is already merged.

dipak-pawar: @lordofthejars @iocanel We can close this. #627 Looks good to me.

Issue #632, opened 2017-03-15
Problems with Private Registry


mgparada:
##### Issue Overview

I have an Arquillian Cube working building my own images, but some days ago I just config a private Registry (under ssl and with user&pass) and trying to pull the images from it, fails always.

PS: I can pull images from Registry manually ;-) So it's not a Registry misconfiguration or something like this.

##### Expected Behaviour

I expect that Arquillian downloads (pull) images from my own Registry, in order to use always a pre-build image.

##### Current Behaviour

When I execute the gradle task and the Arquillian Runner starts, configured to find the image from the Registry, it fails with some NullPointerExceptions. 

##### Additional Information

The arquillian cube config: 
```
<extension qualifier="docker">
        <property name="serverVersion">1.12</property>
        <property name="dockerInsideDockerResolution">false</property>
        <property name="serverUri">https://127.0.0.1:2376</property>
        <property name="tlsVerify">DOCKER_CERT_PATH</property>
        <property name="definitionFormat">CUBE</property>
        <property name="dockerRegistry">xx.yy.com:1234</property>
        <property name="username">xx</property>
        <property name="password">xxxxxx</property>
        <property name="email">mauro@yy.com</property>
        <property name="dockerContainers">
            my-arquillian-service:
                image: aaa/bbb/ccc/my-image-name
        </property>
    </extension>
```
I tried changing some parameters.. 
```<property name="dockerRegistry">https://xx.yy.com:1234/v2</property>```,
```
<property name="dockerContainers">
            my-arquillian-service:
                image: xx.yy.com:1234/aaa/bbb/ccc/my-image-name
        </property>
```
So, i've tried with some differents options but i cannot make it works.. 
The error is as follows: 
```
Arquillian has previously been attempted initialized, but failed. See cause for previous exception
java.lang.RuntimeException: Arquillian has previously been attempted initialized, but failed. See cause for previous exception
	at org.jboss.arquillian.junit.Arquillian.run(Arquillian.java:107)
	at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecuter.runTestClass(JUnitTestClassExecuter.java:114)
	at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecuter.execute(JUnitTestClassExecuter.java:57)
	at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassProcessor.processTestClass(JUnitTestClassProcessor.java:66)
[...]

PR #640, opened 2017-03-20
Sets Remote URL at connection level instead of pull

lordofthejars:

Sets Remote URL at connection level instead of pull

-
-
-

**Fixes**: #632

### DIFF
--- a/docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/DockerClientExecutor.java
+++ b/docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/DockerClientExecutor.java
@@ -161,7 +161,9 @@ public class DockerClientExecutor {
         dockerUri = URI.create(dockerServerUri);
         dockerServerIp = cubeConfiguration.getDockerServerIp();
 
-        configBuilder.withApiVersion(cubeConfiguration.getDockerServerVersion()).withDockerHost(dockerUri.toString());
+        configBuilder.withApiVersion(cubeConfiguration.getDockerServerVersion())
+                .withDockerHost(dockerUri.toString());
+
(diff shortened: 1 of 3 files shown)


bartoszmajsak:
Just a suggestion - if you don't have details to fill-in don't leave the template - it makes unnecessary noise in the PR description.

docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/DockerClientExecutor.java
```diff
@@ -712,10 +718,6 @@ public void pullImage(String imageName) {
 
             PullImageCmd pullImageCmd = this.dockerClient.pullImageCmd(image.getName());
 
-            if (this.cubeConfiguration.getDockerRegistry() != null) {
-                pullImageCmd.withRegistry(this.cubeConfiguration.getDockerRegistry());
```
bartoszmajsak: So before we were pulling with registry and now we set registry globally? Would that work for all images?

lordofthejars: No, before we were just attaching the repo url to the image name. It is a way to append the url of registry to image name programmatically. But this should be already provided by user on docker-compose.

bartoszmajsak: Ok, so docker client executor and cubeConfiguration is per image? What if I have 1 image available in my private repo and other from the docker hub? Are we able to handle it?

lordofthejars:
Yes if you define correctly the docker-compose with correct urls.

El dl., 20 de març 2017 a les 10:41, Bartosz Majsak (<
notifications@github.com>) va escriure:

PR #644, opened 2017-03-22
636: Add openshift client resource provider and ensure that it doesn'…

iocanel:


lordofthejars:
@iocanel There is a really easy conflict in PR, do you want to fix or I do it? It is about an import that I think that both of them should be present.

iocanel:
I can do it, but a little bit later. If you are in a hurry for the release, please feel free to pick it up.

PR #656, opened 2017-03-29
Added condition to cover also a possible specification of the mount permission (i.e. /source:/target:ro)

mahnkong:
Hi, 

thanks for adding the volume mount fix so quickly. This small change covers also a possible specification of the mount permission (i.e. /source:/target:ro) and ensures that in that case also the volume is taken from element 1 of the split result.

Best regards.


lordofthejars:
Thank you very much for your help.

PR #659, opened 2017-04-02
Use the shaded/uberjar version of the kubernetes-client.

iocanel:
The pull request switches to a SNAPSHOT version of the [kubernetes-client](https://github.com/fabric8io/kubernetes-client) that provides a **shaded** uberjar. The uberjar is relocating client packages and adds the major and minor version as part package name, for example:

    io.fabric8.kubernetes.clnt.v2_2.KubernetesClient.

This pretty much ensures that we avoid conflicts, with the user dependencies.

This pull request also adds  enriches that internally use serialization -> deserialization + reflection so that they return objects, based on the kubernetes-client version that user is using. Of course, the user can use the internal types too (if doesn't need a different version).


iocanel:
For anyone that wants to try the client uberjar. The pull request can be found at: https://github.com/fabric8io/kubernetes-client/pull/723

iocanel:
@cescoffier: Trying to make it happen *this* week :-)

iocanel:
@cescoffier: It should be complete now! Any feedback is more than welcome :-)

lordofthejars:
@iocanel two questions does this change something from the user perspective? I mean from existing tests, I mean not having to change any type in terms of enrichers. For example if they were enriching a test with `@KubernetesClient` this class is still the same right?

iocanel:
There are two cases:

- suites that already have the kubernetes-client in their classpath
- suites that only have the kubernetes-client as a transitive dependency from arquillian-cube-kubernetes

For the first kind of users it should be completely transparent. For the later case, users will have to either add `kubernetes-client` as a dependency or use the internal version of the client.

lordofthejars:
So in case of transitive you need to adapt to the new packages right? If so can you update docs with this? We will need to update minor version then for this release.

lordofthejars:
LGTM

@cescoffier waiting your review :D

cescoffier:
Looks good to me.
So 👍

PR #665, opened 2017-04-04
Fixes #664. Uncaught ExecutionException in CubeDockerConfigurator

### LINKED ISSUE #664: Uncaught ExecutionException in CubeDockerConfigurator
##### Issue Overview

Similar to #660. When cube initializes, it runs through [`CubeDockerConfigurator.configure`](https://github.com/arquillian/arquillian-cube/blob/8bcf1960df5be7b2536884e4d2d3513ae21e53cb/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfigurator.java#L63). The potential exists for executables like docker-machine or Boot2Docker to not be available. When this happens, tests fail whether they need to make use of Arquillian Cube or not.

This prevents a test suite being composed from a mix of tests where some use Arquillian Cube and some that do not. 

##### Expected Behaviour

ExecutionException is trapped and logged, enabling the test to proceed. 

##### Current Behaviour

Uncaught ExecutionException is thrown which prevents tests that do not use Arquillian Cube from running.

jamesnetherton:

Relates to #664. Catch any occurrences of  `ExecutionException` prior to test execution.

### DIFF
--- a/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfigurator.java
+++ b/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfigurator.java
@@ -6,7 +6,9 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Random;
 import java.util.UUID;
+import java.util.logging.Level;
 import java.util.logging.Logger;
+
 import org.arquillian.cube.HostIpContext;
 import org.arquillian.cube.docker.impl.client.config.CubeContainer;
 import org.arquillian.cube.docker.impl.client.config.DockerCompositions;
@@ -18,6 +20,7 @@ import org.arquillian.cube.docker.impl.util.OperatingSystemFamily;
 import org.arquillian.cube.docker.impl.util.OperatingSystemResolver;
 import org.arquillian.cube.docker.impl.util.Top;
 import org.arquillian.cube.spi.CubeConfiguration;
+import org.arquillian.spacelift.execution.ExecutionException;
 import org.jboss.arquillian.config.descriptor.api.ArquillianDescriptor;
 import org.jboss.arquillian.core.api.Injector;
 import org.jboss.arquillian.core.api.Instance;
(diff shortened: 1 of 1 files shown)


docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfigurator.java
```diff
@@ -61,18 +64,22 @@ public void configure(@Observes CubeConfiguration event, ArquillianDescriptor ar
...
+                boot2DockerInstance.get(),
+                operatingSystemFamilyInstanceProducer.get());
+            resolver.resolve(config);
+            CubeDockerConfiguration cubeConfiguration = CubeDockerConfiguration.fromMap(config, injectorInstance.get());
+            cubeConfiguration = resolveDynamicNames(cubeConfiguration);
+            System.out.println(cubeConfiguration);
+            hostUriContextInstanceProducer.set(new HostIpContext(cubeConfiguration.getDockerServerIp()));
+            configurationProducer.set(cubeConfiguration);
+        } catch (ExecutionException e) {
+            log.log(Level.WARNING, "Error executing docker command", e);
```
lordofthejars: Maybe this should be `SEVERE` since you cannot run anything isn't it? Also at then end should throw a Runtime exception to not try to execute this test. Without stopping it, the test is not aborted so a test failure about cannot find docker host would be thrown.

jamesnetherton: My reason for not throwing RuntimeException here is for the scenario I outlined in #664.

For example, in my case I have a test suite where some tests that use Cube and are annotated with `@RequiresDocker`. Also other tests that do not make use of Cube.

If Docker is not available, I'd expect the Cube tests to be skipped and the non-cube tests to proceed to run as normal. This is not possible at the moment because the ExecutionException breaks _every_ test.

The alternative to this change would be for me to split the Cube tests into their own module, but I wanted to avoid doing that if possible.

Hope that clarifies things a little.

lordofthejars: Yes I understand your use case, but an exception is going to be thrown anyway of your code right? At the end not throwing an exception at this point does not imply that test is skipped just that Docker Client is going to be executed and would not find the docker host.

In case of `RequiresDocker` it might have sense because it will avoid running that test, but in this case another runtime exception will be thrown. For example instead of throwing an `ExecutionException`  exception, a NPE or ConnectionRefused exception are going to be thrown.

jamesnetherton: Ok. I'll close this PR (and associated issue) and rethink things.

lordofthejars: Cool, I think that best approach would be have each level of tests in its own module. Also it is a good practice to get classpath isolation

PR #666, opened 2017-04-04
Fix visibility for immutable versions of the annotation/label providers.

iocanel:


lordofthejars:
Thank you very much for your help :)

PR #670, opened 2017-04-07
Adds HealtCheck annotation

lordofthejars:


docker/docker/src/main/java/org/arquillian/cube/docker/impl/await/HealthCheckBeforeClassObserver.java
```diff
@@ -0,0 +1,118 @@
...
+        final int pollIterations = healthCheck.iterations();
+        final long sleepTime = Timespan.toMilliseconds(healthCheck.interval());
+        final int responseCode = healthCheck.responseCode();
+        final String method = healthCheck.method();
+        final Long connectTimeout = Timespan.toMilliseconds(healthCheck.timeout());
+
+        return Ping.ping(pollIterations, sleepTime, TimeUnit.MILLISECONDS, () -> {
+            try {
+
+                HttpURLConnection urlConnection = (HttpURLConnection) url.openConnection();
```
bartoszmajsak: Shouldn't we close this connection afterwards? 
Something to investigate maybe https://scotte.org/2015/01/httpurlconnection-socket-leak/

lordofthejars: True, I 'll check it out.

Issue #676, opened 2017-04-13
docker-compose command results in quoted command


adhesivee:
I have a docker-compose.yml configuration and the problem that I have is that the action in `command` gets quoted. 
I have the following files:
docker-compose.yml
```
version: '2'
services:
    wildfly-test:
        build: .
        ports:
            - 8080:8080
            - 9990:9990
        command: /opt/jboss/wildfly/bin/standalone.sh -b 0.0.0.0 -bmanagement 0.0.0.0
```
Dockerfile
```
FROM jboss/wildfly:10.1.0.Final

RUN /opt/jboss/wildfly/bin/add-user.sh test test --silent
```
arquillian.xml
```<?xml version="1.0" encoding="UTF-8"?>
<arquillian xmlns="http://jboss.org/schema/arquillian"
            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
            xsi:schemaLocation="http://jboss.org/schema/arquillian http://jboss.org/schema/arquillian/arquillian_1_0.xsd">

    <container qualifier="wildfly" default="true">
            <property name="chameleonTarget">wildfly:10.1.0.Final:remote</property>
            <property name="username">test</property>
            <property name="password">test</property>
        </configuration>
    </container>
    <extension qualifier="docker">
        <property name="serverUri">tcp://127.0.0.1:2375</property>
        <property name="definitionFormat">COMPOSE</property>
        <property name="dockerContainersFile">docker-compose.yml</property>
        <property name="clean">true</property>
        <property name="cubeSpecificProperties">
          wildfly-test:
            await:
              strategy: http
              responseCode: 200
              url: http://localhost:8080/
              sleepPollingTime: 1s
              iterations: 15
        </property>
    </extension>
</arquillian>
```

When I run this with Cube, then I get error:
`Caused by: com.github.dockerjava.api.exception.NotFoundException: {"message":"oci runtime error: container_linux.go:247: starting container process caused \"exec: \\\"\\\": executable file not found in $PATH\"\n"}`

I can reproduce this error when I do the following command in docker: `docker exec <wildfly-container-name>"/opt/jboss/wildfly/bin/standalone.sh -b 0.0.0.0 -bmanagement 0.0.0.0"`. (exec **with** quotes)

It works when I move the command from docker-compose.yml to Dockerfile.
So it seems that Cube adds quotes around the command which is defined in the docker-compose.yml.

lordofthejars:
[...]

PR #677, opened 2017-04-13
Resolve qouted docker-compose command

adhesivee:
Resolving issue #676. 
I have taken steps described by @lordofthejars and verified that the change works.


lordofthejars:
Thank you very much for the PR could you provide two tests like written in https://github.com/arquillian/arquillian-cube/blob/master/docker/docker/src/test/java/org/arquillian/cube/docker/impl/client/CubeConfigurationTest.java#L90 where you verify that it works with string approach and array approach (https://docs.docker.com/compose/compose-file/#command)? I think that in case of Array approach it will throw a `ClassCastException`, thank you very much.

adhesivee:
@lordofthejars You were right, that part broke. I have updated the tests and the array part.

lordofthejars:
Thank you very much for moving this from a bug to a PR. This fix will be a part of 1.2.0 release, and in that release there will be anew feature that I think you are interested of: https://github.com/arquillian/arquillian-cube/pull/670

Issue #681, opened 2017-04-20
Adds annotation for sleeping before each test


lordofthejars:
Adds annotation for sleeping before each test. As we are giving support to `@HealthCheck` annotation, we can also provide one for static sleep. This is useful for containers where there is no http healtheck.

PR #682, opened 2017-04-20
Adds sleep annotation

lordofthejars:

Adds `@sleep` annotation to pause a test the specified timespan before executing all the test methods.

-
-
-

**Fixes**: #681


bartoszmajsak:
Maybe if you don't use the whole template then don't leave bullet list empty :) It might be a bit confusing.

PR #697, opened 2017-05-15
Fixed infinite loop

### LINKED ISSUE #696: Arquillian initialization ends in infinite loop when a test class has a parent
As a reproducer please try this project - https://github.com/jpechane/arquillian-cube-standalone-example

The problem is located in Reflectionutil.isClassWithAnnotation. When this method is called there is a line in loop - `nextSource = source.getSuperclass()`. When the test class has a parent then the first iteration returns Parent2 class and subsequent iterations return the very same class! thus ending in infinite loop.

IMHO the should code would be
```
                Class<?> nextSource = source;
                while (nextSource != Object.class) {
                    if (nextSource.isAnnotationPresent(annotationClass)) {
                        return true;
                    }
                    nextSource = nextSource.getSuperclass();  // Move up in hierarchy
                }
```

jpechane:
Fixed #696

### DIFF
--- a/core/src/main/java/org/arquillian/cube/impl/util/ReflectionUtil.java
+++ b/core/src/main/java/org/arquillian/cube/impl/util/ReflectionUtil.java
@@ -184,7 +184,7 @@ public final class ReflectionUtil {
                     if (nextSource.isAnnotationPresent(annotationClass)) {
                         return true;
                     }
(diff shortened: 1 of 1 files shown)


lordofthejars:
Thank you very much for your help on this. We really appreciate any kind of help on improving Arquillian.

PR #703, opened 2017-05-22
Use current namespace and fix Openshift resource locator

iocanel:
This pull request addresses two issues:

- adds the ability to use the current namespace (without requiring us to be explicit)
- fixes a minor glitch that prevented the Openshift resource locator to be properly registered.

both issues have been reported as part of: https://github.com/spring-cloud-incubator/spring-cloud-kubernetes/issues/52


lordofthejars:
https://github.com/arquillian/arquillian-cube/issues/710

Issue #708, opened 2017-05-29
Set System Properties with Docker container information


lordofthejars:
##### Issue Overview

Set System Properties with Cube info

##### Expected Behaviour

When running a test next System Properties should be available:

`arq.cube.docker.host` setting the docker host ip.
And for each docker container:

* `arq.cube.docker.<cubeid>.ip` with container ip.
* `arq.cube.docker.<cubeid>.internal.ip` with container internal ip.
* `arq.cube.docker.<cubeid>.port.<exposedport>` with binding port of giving exposed port.

##### Current Behaviour

No system properties.

PR #709, opened 2017-05-29
feat(#708): Adds cube properties as system properties for test

lordofthejars:


docker/docker/src/test/java/org/arquillian/cube/docker/impl/client/SystemPropertiesCubeSetterTest.java
```diff
@@ -0,0 +1,129 @@
...
+import org.mockito.runners.MockitoJUnitRunner;
+
+import java.util.HashSet;
+import java.util.Set;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.when;
+
+@RunWith(MockitoJUnitRunner.class)
+public class SystemPropertiesCubeSetterTest {
```
bartoszmajsak: Would it make sense to have some integration test for that too?

lordofthejars: you mean using Arquillian capabilities? I don't think so since it is just a fire of the events. In general I usually this approach for testing observers. I mean in this concrete simple case firing the event or calling the method directly does not offer really a plus.

bartoszmajsak: I mean against real docker and arq event machine instead of just unit test.

docs/enrichers.adoc
```diff
@@ -336,3 +336,16 @@ So previous example could be modified to:
 ----
 
 And in case you are running on _Windows_ or _MacOS_, `host`property will be automatically set to the +boot2docker +_ip_.
+
+=== System Properties Injection
+
+Arquillian Cube sets some system properties to be used in tests in case you cannot use enrichers (i.e in case of JUnit rules).
```
MatousJobanek: Correct me If I'm wrong, but in JUnit rules it is possible to use enrichers...

lordofthejars: I meant Arquillian Enrichers. If you are using Arquillian Cube or Arquillian APE JUnit rules then you are not using  Arquillian runner.

docker/docker/src/test/java/org/arquillian/cube/docker/impl/client/SystemPropertiesCubeSetterTest.java
lordofthejars: Ok, more like an ftest. Yes I can do it for sure, I think that it has sense.

docs/enrichers.adoc
MatousJobanek: Ah, OK then... maybe it would be good to clarify it there...

PR #719, opened 2017-06-06
Port remaining fixes from fabric8 to cube (#704)

nicolaferraro:
Porting:
- OSFUSE-488
- OSFUSE-534
- OSFUSE 324

The implementation is a bit different (since many classes changed) from the fabric8io version (especially for 324), but it should be equivalent.

I've verified that the following issues are already fixed in arquillian cube:
- OSFUSE-487
- OSFUSE-486
- OSFUSE-317

If @astefanutti confirms the other 2 remaining issues, #704 can be closed.

Issue #704, opened 2017-05-22
Port remaining fixes from fabric8 to cube.


iocanel:
I think that its time to deprecate the `Fabric8 Arquillian` in favor of `Arquillian Cube Kubernetes/Openshift`.

However there are still a couple of fixes that went in `Fabric8 Arquillian` the last couple of months that need to be ported. As these fixes were not implemented by me, it would be fair to call the original authors for the migration.

The fixes are:
- OSFUSE-534
- OSFUSE-499
- OSFUSE-488
- OSFUSE-487
- OSFUSE-486
- OSFUSE-324
- OSFUSE-317
- OSFUSE-175

And the authors are: @chirino, @nicolaferraro and @astefanutti.
Guys, can you port your fixes over here? I'd be happy to provide any help if needed or even port them for you, but I have to ask first :-D

lordofthejars:
+1 Thank you very much @iocanel for taking this.

astefanutti:
It may be that OSFUSE-175 and OSFUSE-499 (OpenShift basic user role permission issues) are actually already _contained_ in the current codeline (by having a quick look at it). I'll do an actual test ASAP to confirm.

iocanel:
@astefanutti: Thanks a lot!

lordofthejars:
https://github.com/arquillian/arquillian-cube/pull/719

bartoszmajsak:
Thanks for the joined effort guys! Awesome stuff.

Once we merge all the changes we should discuss next steps. I can think of

  * communication about what happened (both on arquillian and fabric8 channels: blogs, twitter etc)
  * update old repository to link to cube
  * review and update documentation

Anything else I missed?

astefanutti:
I've just created #731 that should cover OSFUSE-175. I've just tested OSFUSE-499 is already covered.

PR #731, opened 2017-06-13
OSFUSE-175: Better existing project check for basic user

astefanutti:
Follows up #719 and complete #704.


lordofthejars:
Ok, now when the build finishes I'll squash both PRs. Then tomorrow morning I'll release a new Cube version.

Thank you so much.

PR #745, opened 2017-06-27
Openshift: Consider the cleanup flag when destroying pods/services

jwendell:
To keep the consistency.


lordofthejars:
Thank you so much for your help.

Issue #725, opened 2017-06-07
Provide a way of passing environment variables to setup scripts


iocanel:
At the moment, we pass the following environment variables to setup scripts (k8s & openshift):

- existing env variables.
- master url
- namespace
- docker registry
- domain

But we don't provide means of explicitly specifying passing env variables.

PR #759, opened 2017-07-18
fix: #725 Allow arquillian descriptor properties prefixed with env. a…

iocanel:
…s env variables to setup/teardown scripts in k8s/openshift.

-
-
-

**Fixes**: #

### DIFF
--- a/kubernetes/kubernetes/src/test/java/org/arquillian/cube/kubernetes/impl/DefaultConfigurationTest.java
+++ b/kubernetes/kubernetes/src/test/java/org/arquillian/cube/kubernetes/impl/DefaultConfigurationTest.java
@@ -1,0 +1,46 @@
+package org.arquillian.cube.kubernetes.impl;
+
+import org.junit.Test;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.assertj.core.api.Assertions.*;
+
+
+public class DefaultConfigurationTest {
+
+    @Test
+    public void shouldParseNullMap() throws Exception {
+        Map<String, String> result = DefaultConfiguration.parseMap(null);
+        assertThat(result).hasSize(0);
+    }
+
+    @Test
+    public void shouldParseEmptyMap() throws Exception {
+        Map<String, String> result = DefaultConfiguration.parseMap("");
+        assertThat(result).hasSize(0);
+    }
+
+    @Test //We just want to know it won't barf
+    public void shouldParseJustAKey() throws Exception {
+        String s = "KEY1";
+
+        Map<String, String> result = DefaultConfiguration.parseMap(s);
+        assertThat(result).hasSize(1);
+    }
+
+    @Test
+    public void shouldParseNormalMap() throws Exception {
+       String s = "KEY1=VALUE1\n" +
+        "KEY_TWO=VALUE2\n" +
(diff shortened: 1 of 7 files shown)


kubernetes/kubernetes/src/test/java/org/arquillian/cube/kubernetes/impl/DefaultConfigurationTest.java
```diff
@@ -0,0 +1,33 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.Assert.*;
+
+/**
+ * Created by iocanel on 6/7/17.
```
bartoszmajsak: Do we need this header? :)

kubernetes/kubernetes/src/test/java/org/arquillian/cube/kubernetes/impl/DefaultConfigurationTest.java
```diff
@@ -0,0 +1,33 @@
...
+        map.put("KEY_TWO", "VALUE2");
+        map.put("env.KEY3", "VALUE3");
+        map.put("env.key4", "VALUE4");
+        map.put("env.KEY_5", "VALUE5");
+
+        Map<String, String> result = DefaultConfiguration.filterEnvironmentVariables(map);
+        Assert.assertEquals(3, result.size());
+        Assert.assertTrue(result.containsKey("KEY3"));
+        Assert.assertTrue(result.containsKey("key4"));
+        Assert.assertTrue(result.containsKey("KEY_5"));
```
bartoszmajsak: I think we are using `assertj`, so maybe worth a small change?

iocanel:
@bartoszmajsak: addressed both comments

lordofthejars:
Thank you so much, ready for squash then?

bartoszmajsak:
Thanks @iocanel. Just wanted to ask you one thing. In `arquillian-core` we support expressions in `arquillian.xml` which refer to environment variables, e.g. `${env.JAVA_HOME}` will be evaluated to `JAVA_HOME` value. Do we need this extra filtering?

lordofthejars:
`arquillian.xml` supports out of the box setting environemnt variables on it. For example:

```
<property name="home">${env.JAVA_HOME}</property>
```

Is this what you need?

iocanel:
No, I don't need something like that. 

I am starting to this that this is a strong indication that this is going to wrong direction (since its too confusing)...

So let's start fresh:

 "How can I use a map property via arquillian.xml"?

Can I just use something like:

    <property key="scriptEnv">
        key1=value1
        key2=value2
        key3=value3
    </property>


and If so how can I read it?

lordofthejars:
You will read this as String. So what you'll need to do is `new ByteInputStream(content)` and pass it to `Properties`.

lordofthejars:
Notice that you can still use `${}` syntax anywhere in the string since Arquillian will just replace any occurrences on given configured string.

iocanel:
Thanks I will update the pull request accordingly.

lordofthejars:
Thank you so much.

lordofthejars:
https://github.com/arquillian/arquillian-cube/issues/783

Issue #736, opened 2017-06-15
Docker drone is expecting the wrong responseCode for Selenium 3.4.0 


jollytoad:
Docker-Drone fails to detect that selenium-chrome-debug has started with Selenium 3.4.0.
In https://github.com/arquillian/arquillian-cube/blob/8bcf1960df5be7b2536884e4d2d3513ae21e53cb/docker/drone/src/main/java/org/arquillian/cube/docker/drone/SeleniumContainers.java#L163 the await is set to expect a response code of 403, but it seems that may no longer be the case in 3.4.0, instead you get a 200.

I also can't seem to override the await in the arquillian.xml docker configuration 

After setting a breakpoint on that line and manually overriding the responseCode via the debugger, it seems to work.

lordofthejars:
Thank you so much, yes in case of drone integration, you cannot configure it in `arquillian.xml`. PRs are more than welcome would you like to send a PR just fixing this? The thing is that should be something like if seleniumVersiion is < 3.5.0 then use 403 else 200.

jollytoad:
I'm not sure exactly which version of Selenium this changed in, and unfortunately don't have time to investigate atm.

I managed to override with an extension for the time being...

```java
public class CubeDroneHackExtension implements LoadableExtension {
    @Override
    public void register(LoadableExtension.ExtensionBuilder extensionBuilder) {
        extensionBuilder.observer(CubeDroneHack.class);
    }
}
```

```java
public class CubeDroneHack {
    @Inject
    @ApplicationScoped
    InstanceProducer<SeleniumContainers> seleniumContainersInstanceProducer;

    public void install(@Observes(precedence = 80) CubeDockerConfiguration configuration) {
        final SeleniumContainers seleniumContainers = seleniumContainersInstanceProducer.get();
        final CubeContainer seleniumContainer = seleniumContainers.getSeleniumContainer();
        seleniumContainer.getAwait().setResponseCode(200);
    }
}
```

georgwolf:
The behaviour has been changed with Selenium 3.x.
I created pull request #760 for fixing this.

PR #760, opened 2017-07-18
Fix for #736: Expect response code 200 for Selenium containers version >= 3

georgwolf:
Docker drone is expecting the wrong responseCode for Selenium from version 3 onwards

- Retrieve Selenium major version from image tag
- For version >= 3 and as default (e.g. to cover "latest" tag) expect 200 response code
- For versions < 3 expect 403 response code

**Fixes**: #736

### DIFF
--- a/docker/drone/src/main/java/org/arquillian/cube/docker/drone/SeleniumContainers.java
+++ b/docker/drone/src/main/java/org/arquillian/cube/docker/drone/SeleniumContainers.java
@@ -159,7 +159,6 @@ public class SeleniumContainers {
 
         Await await = new Await();
         await.setStrategy("http");
-
         await.setResponseCode(getSeleniumExpectedResponseCode());
         await.setUrl("http://dockerHost:" + SELENIUM_BOUNDED_PORT);
 
--- a/docker/drone/src/main/java/org/arquillian/cube/docker/drone/SeleniumContainers.java
+++ b/docker/drone/src/main/java/org/arquillian/cube/docker/drone/SeleniumContainers.java
@@ -159,8 +159,19 @@ public class SeleniumContainers {
 
         Await await = new Await();
         await.setStrategy("http");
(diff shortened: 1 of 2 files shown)


lordofthejars:
There are some build failures https://travis-ci.org/arquillian/arquillian-cube/builds/254850686?utm_source=github_status&utm_medium=notification

georgwolf:
Sorry, forgot to put the WIP in the title

lordofthejars:
No worries, and thanks for the contribution.

georgwolf:
Not really sure what to do in the case of custom browser images: 

- Either add a cube drone configuration property for setting the expected response code
- or add a cube drone configuration property setting the selenium version the image is based on
- or take Selenium version from classpath

Any preferences there?

docker/drone/src/main/java/org/arquillian/cube/docker/drone/SeleniumContainers.java
```diff
@@ -159,8 +159,19 @@ private static void setDefaultSeleniumCubeProperties(CubeContainer cubeContainer
 
         Await await = new Await();
         await.setStrategy("http");
-        // Doing an http request to selenium node returns a 403 if Jetty is up and running
-        await.setResponseCode(403);
+
+        // Selenium from 3.x onwards returns 200 if started
+        int expectedResponseCode = 200;
```
lordofthejars: Can you add all calculation of `expectedResponseCode ` calculation into  a method?

lordofthejars:
I think that this way _Selenium version from classpath_ would be the best one here, also because we already have this logic implemented.

georgwolf:
Always using the version from class path now to be consistent and extracted to separate method.
Tested for Selenium 3.x by explicitly picking 3.4.0 in pom.xml and adjusting SeleniumContainersTest appropriately.

lordofthejars:
That's great!!! It looks really awesome. Thank you very much for your contribution.

Issue #768, opened 2017-07-27
[Cube-Docker] Add support for building images that reference an image from a private registry


georgwolf:
##### Issue Overview

Building Docker images based on an image from a private registry fails with
`com.github.dockerjava.api.exception.DockerClientException: Could not build image: Get https://[...]/manifests/latest: no basic auth credentials`

##### Expected Behaviour
Registry settings from arquillian.xml docker extension are used for build too.

##### Steps To Reproduce
1. Build an image with a FROM that references an image from a private repository that hasn't been pulled yet

lordofthejars:
Have you tried to configure all credential things in `arquillian.xml`, in theory same mechanism should be used

georgwolf:
Yes, pulling that same image works but building from it does not.
I checked with the debugger and the java-docker lib does use the registry settings from the DockerClientConfig for pulling and pushing images by default but it doesn't use them for building.
This requires explicitly setting them via BuildImageCmd#withBuildAuthConfigs.

I'll create a PR for this.

lordofthejars:
Great!!! Thank you very much for your investigations.

PR #769, opened 2017-07-27
Add support for private Docker registry to build command

georgwolf:
Details: #768

**Fixes**: #768

### DIFF
--- a/docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/DockerClientExecutor.java
+++ b/docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/DockerClientExecutor.java
@@ -785,14 +785,15 @@ public class DockerClientExecutor {
             buildImageCmd.withDockerfile(new File((String) params.get(DOCKERFILE_NAME)));
         }
         
-        AuthConfig buildAuthConfig = new AuthConfig()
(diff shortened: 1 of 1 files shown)


docker/docker/src/main/java/org/arquillian/cube/docker/impl/docker/DockerClientExecutor.java
```diff
@@ -782,6 +784,15 @@ private void configureBuildCommand(Map<String, Object> params, BuildImageCmd bui
         if (params.containsKey(DOCKERFILE_NAME)) {
             buildImageCmd.withDockerfile(new File((String) params.get(DOCKERFILE_NAME)));
         }
+        
+        AuthConfig buildAuthConfig = new AuthConfig()
```
lordofthejars: Should we add here an if to make this only set in case of user set a username and password?

georgwolf: It'll use default docker hub and your logon name if values are null, but probably still a good idea to only set the auth config in case username and password are not null

lordofthejars: Basically to protect us in case some day defaults works differently in `docker-java`.

Thanks

PR #773, opened 2017-08-01
Move the PortFowarder tests to kubernets module

spolti:
Move the PortFowarder tests from openshift package to kubernetes as the port forwarder was code was moved to the kubernetes.


lordofthejars:
@iocanel Is this change affecting the bug you found?

iocanel:
No, this just moves a test to a place where it makes more sense.

lordofthejars:
https://github.com/arquillian/arquillian-cube/issues/779

PR #774, opened 2017-08-02
Openshift: Fix @Observer parameter

jwendell:
It's the Configuration class that gets injected, not
CubeOpenShiftConfiguration, so, this second parameter will
never be resolved and thus the method will never be invoked.

This suppresses the following warning:

WARNING: Argument 1 for OpenShiftSuiteLifecycleController.startAutoContainers is null. It won't be invoked.


lordofthejars:
https://github.com/arquillian/arquillian-cube/issues/780

Issue #744, opened 2017-06-27
[Arquillian Cube OpenShift] Implement resource provider for the OpenShift route


RadekKoubsky:
It would be nice to have a resource provider for the OpenShift routes to inject them into a test class. 

An example of such solution:
```
@RunWith(Arquillian.class)
public class ExampleTest {
    @ArquillianResource
     OpenShiftClient client;

     @ArquillianResource
     Route route;
}
```

instead of getting a route using the openshift client:
```
final Route route = this.client.adapt(OpenShiftClient.class)
                .routes()
                .inNamespace(this.session.getNamespace())
                .withName(this.applicationName)
                .get();
```

bartoszmajsak:
Hey @RadekKoubsky. Thanks for suggesting this improvement. Would you be interested in providing PR for that? 

If you need some help discussing details of the implementation just ping me or @lordofthejars.

RadekKoubsky:
@bartoszmajsak Hi, I would love to work on Arquillian Cube. However, I have lot of prioritized items for our target release which is not focused on using Arquillian Cube (that is supposed to be for the future release, 1.1.0.Next in Jira). I will come back to this later on.

spolti:
I think we have this option on ce-arq (an extension of arquillian-cube).
Instead to use arquillianResource we use this to obtain the route based on the application name:
`   
@RouteURL("eap-app")
private URL url;`

https://github.com/jboss-openshift/ce-arq/blob/master/cube/src/main/java/org/jboss/arquillian/ce/cube/RouteURL.java

lordofthejars:
Then maybe we can try to reuse this inside cube

spolti:
@lordofthejars I'll try to do that, I just have to finish some some priorities.

jwendell:
dupe of #465?

PR #776, opened 2017-08-03
Move RouteURL functionality from ce-arq

spolti:
This commit aggregate a useful functionality ported from CE-ARQ (https://github.com/jboss-openshift/ce-arq) that allows user to get the routes to their services just by using the @RouteURL annotation in the test class, example:

```
@RouteURL("eap-app")
private URL routeToMyApp
```

**Fixes**: #744


lordofthejars:
Thank you so much

lordofthejars:
Hi, can you update docs as well with this feature so every user is aware of that?

Thank you very much.

spolti:
@lordofthejars sure thing, I'll do that.

PR #788, opened 2017-08-22
Add the ability to capture pods logs

jwendell:


douglaspalmer:
Would it be better for the default output location to be target/surefire-reports? This is where users expect to find test logs.

douglaspalmer:
After running with this PR for a couple of days I have a further request to change the output location.
As well as writing the file to target/surefire-reports I think it would be better to avoid subdirectories and use pod names in the filename instead. i.e. target/surefire-reports/ClassName-PodName.log

jwendell:
I'd  love some feedback WRT property names, I'm not sure I have chosen the best ones (env.console.logs.copy and env.console.logs.path).

jwendell:
I updated the patch to save the files in a flat structure. Indeed it seems better.

douglaspalmer:
WRT property names. How about console.logging.enabled and console.logging.path?

douglaspalmer:
Also, I feel that this should be enabled by default.

dipak-pawar:
@jwendell Can we add some tests around this functionality?

jwendell:
@dipak-pawar done.

jwendell:
Notice: Until https://github.com/fabric8io/kubernetes-client/issues/843 is fixed, all log files are created but they will be empty.

docs/kubernetes.adoc
```diff
@@ -47,43 +47,45 @@ For example **foo.bar.baz** will be converted to **FOO_BAR_BAZ**.
...
+| env.init.enabled                    | Bool (true)    | Any | Flag to initialize the environment (apply kubernetes resources)
+| env.config.url                      | URL            | Any | URL to the Kubernetes JSON/YAML (defaults to classpath resource kubernetes.json)
+| env.config.resource.name            | String         | Any | Option to select a different classpath resource (other than kubernetes.json)
+| env.script.env                      | String         | Any | Key/Value pairs to pass to setup scripts as environment variables
+| env.setup.script.url                | URL            | Any | Option to select a shell script that will setup the environment
+| env.teardown.script.url             | URL            | Any | Option to select a shell script to tear down / cleanup the environment
+| env.dependencies                    | List           | Any | Whitespace separated list of URLs to more dependency kubernetes.json
+| env.console.logs.copy               | Bool (false)   | Any | Whether to capture the pods logs and save them into the filesystem - as individual files, one for each pod. Filenames will be "ClassName-[MethodName-]-PodName.log".
```
dipak-pawar: `env.console.logs.copy` is little ambiguos. How about `env.pod.logs.copy`?

docs/kubernetes.adoc
```diff
@@ -47,43 +47,45 @@ For example **foo.bar.baz** will be converted to **FOO_BAR_BAZ**.
...
+| env.console.logs.path               | String         | Any | Directory where to save the pods logs. Defaults to "target/surefire-reports".
+| wait.timeout                        | Long (5mins)   | Any | The total amount of time to wait until the env is ready
```
dipak-pawar: env.pod.logs.path

docs/kubernetes.adoc
dipak-pawar: As per my understanding, we are capturing pod logs for entire execution of test suite. If so then do we really need classname & methodname for pod logs `fileName`? Only pods name is not enough here?

correct me if I am wrong

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/SessionManager.java
```diff
@@ -185,21 +202,102 @@ public void createEnvironment(Session session) {
...
+        IoUtils.safeClose((Closeable[]) lw.toArray());
+    }
+
+    private void setupConsoleListener() {
+        if (!configuration.isEnvironmentConsoleLogCopyEnabled())
+            return;
+
+        consoleLogPath = configuration.getEnvironmentConsoleLogPath();
+        if (Strings.isNullOrEmpty(consoleLogPath))
+            consoleLogPath = String.format("%s/target/surefire-reports", System.getProperty("user.dir"));
```
dipak-pawar: What if user is running this for `IntelliJ IDEA` & they didn't set property to change this default path?
 
I believe directory `target/surefire-reports`  will be missing in case of `IDEA` as it is not using surefire to run tests.

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/SessionManager.java
```diff
@@ -185,21 +202,102 @@ public void createEnvironment(Session session) {
...
+            className = "NOCLASS";
+        fileName += String.format("/%s", className);
+
+        if (Strings.isNotNullOrEmpty(methodName))
+            fileName += String.format("-%s", methodName);
+        fileName += String.format("-%s.log", podName);
+
+        try {
+            final FileOutputStream stream = new FileOutputStream(fileName);
+            LogWatch lw = client.pods().inNamespace(session.getNamespace()).withName(podName).watchLog(stream);
```
dipak-pawar: How about if pod has multiple containers? I believe container name is optional if pod has only one container.

Again if we provide support for this we can get logs of container from pod.

docs/kubernetes.adoc
douglaspalmer: I container mode is set to class instead of suite then we need classname. I don't think we need method.

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/SessionManager.java
douglaspalmer: If the dir does not exist then it will be created. See line 257.

new File(consoleLogPath).mkdirs();

docs/kubernetes.adoc
douglaspalmer: Sounds good to me

docs/kubernetes.adoc
douglaspalmer: Sounds good to me

docs/kubernetes.adoc
jwendell: Yeah, agreed, although I think we can get rid of "env" prefix.

BTW I'm about to push a commit that expands this functionality to also capture kubernetes events from the cluster. This helps us to find errors on the server side, such as an image not found on registry, or some other kubernetes failure.

So far, I'm using this same flag to enable such feature, once *I* think it has the same semantics: "I want to debug what's going on the server, send me all the logs you have". Does it make sense? If so, we could think of a more generic flag here, something like "logs.copy". Otherwise I'll open a separate PR and will use other flags for it.

jwendell: Yeah, the idea is to easily identify which tests/methods created which pods. So, if you have lots (think of hundreds, thousands) of tests and just a few of them are failing, you want to check the logs of the pods created by that failing test, instead of digging into a huge log file.

douglaspalmer: I still think it should be enabled by default.

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/SessionManager.java
jwendell: Yeah, this is only the default value. It can be overridden by the user. If you have a better default value, please tell me and I'll update the code.

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/SessionManager.java
jwendell: Yeah, I didn't think of pods with multiple containers... The easiest way would be to capture the logs of the first container, though I'm not sure it's the best thing to do... I'll think a bit more about it.

docs/kubernetes.adoc
jwendell: @dipak-pawar can you opine here about the flag name? As well as whether to enable it by default or not?

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/SessionManager.java
jwendell: I've updated the patch so that we get the logs for all containers within a pod. So, if a pod has 3 containers, we'll get 3 log files, with the container name being part of the filename.

jwendell:
As stated in previous comment, I pushed one more commit, that adds the ability to also capture the kubernetes events, in addition of pods logs.

Please tell me if you prefer this to be a separated PR, with its own set of flags to enable/disable it. As I mentioned earlier, I think it makes sense to think of this feature (pods logs + k8s events) as a single debugging feature.

lordofthejars:
For me it is fine in this way. @dipak-pawar I think that changes you requested has been done. Can we merge?

jwendell:
My only "concern" is about the flag name.
It is `env.console.logs.copy`. Should we change it to `env.pod.logs.copy`? Or even `pod.logs.copy` (and its counterpart `pod.logs.path`?

Or maybe better and shorter `logs.copy` and `logs.path` once we are fetching both pods logs and kubernetes cluster logs?

lordofthejars:
@jwendell I am very bad giving name to things but if we have `ansi.logger.enabled`  maybe it should be called `pod.logs.copy` since it is the log of pods. WDYT?

jwendell:
@lordofthejars haha, I think this is one of the biggest problem in computer science, along with cache invalidation - https://martinfowler.com/bliki/TwoHardThings.html :-)

I suggested the shorter `logs.copy` (or `logs.enabled`) because we are not restricted to **pods** logs, we are also fetching kubernetes cluster events log (`kubectl get events`).

That said, I don't have a strong opinion here. I'm fine with any of the alternatives.

lordofthejars:
Ok then `logs.copy` is the best for me, because you are creating a file with logs, not just printing on console.

jwendell:
Ok, I'll update the PR accordingly.

jwendell:
pushed

Issue #789, opened 2017-08-31
NPE when getting default admin port


jensstn:
core/src/main/java/org/arquillian/cube/impl/client/container/ContainerConfigurationController.java:getDefaultPortFromConfigurationInstance can throw a NullPointerException when calling invoke on the Method returned by ```fieldName.getReadMethod()``` (L.105).
This has been observed running against a glassfish-remote-container.

bartoszmajsak:
Any idea why it throws `NPE`?

jensstn:
i didn't have time to dig a lot deeper and it's been a few days, but as far as i remember it happens because it finds a PropertyDescriptor which only has a write method.
do you want me to dig further?

bartoszmajsak:
If you could afford some time for figuring it out that would be awesome. Thanks a lot for your contribution!

lordofthejars:
I close the issue because I have merged the PR, but of course if you have time it is still interesting to know why this was null. I just merge the PR so we can release a new version with this fix.

PR #790, opened 2017-08-31
Fix potential NullPointerException

jensstn:
This fixes a potential null pointer exception when getting default container ports.
Observed running against a remote glassfish container. 

https://github.com/arquillian/arquillian-cube/issues/789


dipak-pawar:
lgtm. Thank you @jensstn

Issue #791, opened 2017-09-06
Support for parallel tests with Cube Drone extension


georgwolf:
##### Issue Overview
Cube drone uses a static name for browser/vnc/flv2mp4 containers thus blocking running tests in parallel.

##### Expected Behaviour
Option to generate unique container names.

I'll submit a PR for this.

PR #792, opened 2017-09-06
Drone parallel test support

georgwolf:
Enables running tests using Cube Drone in parallel.

- Two new cubedrone configuration options
  - containerNameStrategy
  - containerNamePrefix

- STATIC name strategy: same behaviour as before
- STATIC_PREFIX name strategy: Adds a configurable prefix to the static container name
- RANDOM name strategy: Adds a random UUID to the container name

STATIC_PREFIX and RANDOM also cause a random port being picked for the Selenium bound port.

**Fixes**: #791


lordofthejars:
Thank you very much for your help on this. Have you tried in real project to check that it works? The theory after reading the code is that it should work, but just to be sure.

Thank you very much.

lordofthejars:
Thank you so much for all your efforts on making this better and better.

Issue #802, opened 2017-09-12
@RouteURL injection doesn't work when routerHost isn't configured.


Ladicek:
##### Issue Overview

`@RouteURL` injection doesn't work, it complains that

```
java.lang.RuntimeException: Could not set RouteURL value on field private java.net.URL io.openshift.booster.OpenshiftIT.url
Caused by: java.lang.IllegalArgumentException: Must specify routerHost!
```

This is consistent with what documentation says (http://arquillian.org/arquillian-cube/#_enrichers), but it doesn't make any sense. The route is read directly from the OpenShift API, the configured `routerHost` isn't even used in `RouteURLEnricher`, it's just being checked.

If I remove the check, build Arquillian Cube myself and use the built snapshot, it works just fine.

##### Expected Behaviour

`routerHost` configuration isn't required for injecting `@RouteURL`.

##### Current Behaviour

`@RouteURL` injection doesn't work when `routerHost` isn't configured.

##### Steps To Reproduce
1. build current Arquillian Cube master locally, the following testing project depends on 1.8.1-SNAPSHOT
2. `git clone -b arquillian-cube https://github.com/Ladicek/wfswarm-rest-http.git`
3. `mvn clean verify -Popenshift,openshift-it -Dnamespace.use.current=true`

##### Additional Information

CC @rcernich as original author of `RouteURLEnricher`

jwendell:
+1. This is only necessary when the routes provided by Openshift cannot be resolved by default (think of addresses like `<service>.<random-namespace-name>.svc.cluster.local`).

If the route can be resolved somehow (valid dns, entry in /etc/hosts, etc), then specifying routerHost is not mandatory.

I'll put some comments on the PR.

Ladicek:
`service-name.namespace-name.svc.cluster.local` is a valid hostname of a _service_. I thought that routes always have publically resolvable names, otherwise there's no point in creating a route in the first place? Anyway, thanks for review!

rcernich:
The names have to be resolved by somebody, which means integration with some DNS service.  It's been a while since I've used any of the development packages of OpenShift and I don't know if they now integrate a DNS server for resolving routes directly.  Because of that, @RouteURL was implemented in such a way that a user could specify the host IP being used for the router so we could resolve the route names directly.  That said, perhaps replacing the error with a log message would work better.

PR #803, opened 2017-09-12
skip the routerHost check in RouteURLEnricher

Ladicek:

With this change, `@RouteURL` injection works even if `routerHost` isn't configured.

- Remove the check for `routerHost` in `RouteURLEnricher`.
- Remove unreadable text in `@RouteURL` javadoc.

**Fixes**: #802

### DIFF
--- a/openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/enricher/RouteURL.java
+++ b/openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/enricher/RouteURL.java
@@ -9,11 +9,8 @@ import java.lang.annotation.Target;
 import static java.lang.annotation.RetentionPolicy.RUNTIME;
 
 /**
- * OpenShiftRouter
- * <p/>
- * Used with @ArquillianResource URL objects. Provides a URL to the OpenShift
- * router configured on the ce-cube extension, u    e.g.
- * arq.extension.openshift.routerHost, or through arquillian.xml using routerHost parameter, setting the hostname appropriately.
+ * Provides a URL of given OpenShift route. If the route is not resolvable, you need to set the {@code routerHost} setting
+ * to the IP address of the OpenShift router.
  *
  * @author Rob Cernich
  */
--- a/openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/enricher/RouteURLEnricher.java
+++ b/openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/enricher/RouteURLEnricher.java
@@ -76,11 +76,6 @@ public class RouteURLEnricher implements TestEnricher {
             throw new NullPointerException("CubeOpenShiftConfiguration is null.");
         }
 
-        final String routerAddress = config.getRouterHost();
-        if (routerAddress == null || routerAddress.length() == 0) {
-            throw new IllegalArgumentException("Must specify routerHost!");
-        }
-
         final OpenShiftClient client = clientInstance.get();
         final Route route = client.getClient().routes().inNamespace(config.getNamespace()).withName(routeName).get();
         if (route == null) {


openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/enricher/RouteURL.java
```diff
@@ -9,11 +9,7 @@
 import static java.lang.annotation.RetentionPolicy.RUNTIME;
 
 /**
- * OpenShiftRouter
- * <p/>
- * Used with @ArquillianResource URL objects. Provides a URL to the OpenShift
- * router configured on the ce-cube extension, u    e.g.
- * arq.extension.openshift.routerHost, or through arquillian.xml using routerHost parameter, setting the hostname appropriately.
+ * Provides a URL of given OpenShift route.
  *
```
jwendell: I suggest you comment that, if the route is not resolvable, then the user will need to specify the routerHost containing the router ip of the Openshift cluster, otherwise addresses returned by this annotation won't be resolvable.

Ladicek:
I could put an `if` around `ArqCubeNameService.setRoutes` in `BuildablePodCube.start`, because that seems to be the only place from where `ArqCubeNameService.setRoutes` is called, but I don't feel very confident in doing that. So I just adjusted the `@RouteURL` javadoc.

jwendell:
ok, that's fine to me. we can tweak it later...

Ladicek:
Right. I don't really understand that part of the codebase, nor the motivations behind it, so I'd prefer if someone more knowledgeable than me did that part.

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/enricher/RouteURLEnricher.java
```diff
@@ -76,11 +76,6 @@ private URL lookup(RouteURL routeURL) {
             throw new NullPointerException("CubeOpenShiftConfiguration is null.");
         }
 
-        final String routerAddress = config.getRouterHost();
-        if (routerAddress == null || routerAddress.length() == 0) {
-            throw new IllegalArgumentException("Must specify routerHost!");
-        }
-
         final OpenShiftClient client = clientInstance.get();
         final Route route = client.getClient().routes().inNamespace(config.getNamespace()).withName(routeName).get();
```
lordofthejars: If routeName is null, then I guess that `OpenShiftClient` just detects a `null` and do nothing isn't it?

Ladicek: If `routeName` is `null`, then a NPE is thrown a couple of lines above.

lordofthejars: So we are getting a `NPE` right? So this is not fixing the error, or am I missing something?

Ladicek: So, there are two things: `routeName`, which is the value of the `@RouteURL` annotation, and that is of course mandatory. And then there's this `routerAddress`, which is like totally useless in this case, and it is the check for `routerAddress` that I'm removing here.

lordofthejars: ah right sorry I miss the names, true, sorry for my confusion. Maybe it would be better a log line? Or this would not add any interesting information for the user. I am just thinking to not missing something important that the developer might be interested with like this.

Ladicek: So, frankly, I really don't understand the purpose of configuring the address of the router, as routes are AFAIK always resolvable (that's the purpose of routes, right?). But apparently, there are cases that I'm not aware of, so I need to lean on your advice.

spolti: @Ladicek the purpose of RouteURL is resolve the routes outside OPenshif for the tests which are annotated with @RunAsClient, it read all routes from Openshift and add the the JVM custom DNS naming service.
If you remove this check when use the @RoutURL it will thrown a NPE later when the tests are running in the time arquillian will try to add the DNS records to the JVM.
If you are not running tests with RunAsClient so there is no reason to you to use this annotation.

jwendell: @Ladicek routes are not always resolvable. It depends on how the cluster administrator set them, or how users create them. Like we keep saying, addresses must be resolvable by somebody. Be it yourself, or your provider dns server, or anybody else.

So, like I said in the past, I'm +1 for this patch.

lordofthejars: But what @spolti suggests is that this patch might produce a NPE isn't it?

lordofthejars: Clear now :) Then we are ready to merge isn't it?

Ladicek: @spolti Sorry, I don't follow. I'm pretty sure that the change in this PR is actually OK, though, because this branch https://github.com/Ladicek/wfswarm-rest-http/tree/arquillian-cube builds and runs just fine for me. (`mvn clean verify -Popenshift,openshift-it -Dnamespace.use.current=true` is the command to use, in case anyone's interested.) I'm not using `@RunAsClient` there, but I tried it and it doesn't make a difference.

jwendell: @lordofthejars yep

lordofthejars: @Ladicek yes he tell that he was wrong, that it is currently working as expected

Ladicek: Thanks all!

Ladicek:
Sorry I didn't update docs in this PR, I sent https://github.com/arquillian/arquillian-cube/pull/807 as a follow-up.

Issue #804, opened 2017-09-12
Cube Drone with recordingMode NONE throws java.lang.IllegalArgumentException


georgwolf:
##### Issue Overview
When recordingMode is set to NONE
"java.lang.IllegalArgumentException: Video conversion cube is not present in the registry." is thrown.

I'll create a PR for this.

PR #805, opened 2017-09-12
Don't start flv2mp4 container when recordingMode is NONE

georgwolf:
Prevents "java.lang.IllegalArgumentException: Video conversion cube is not present in the registry." being thrown when recording is disabled.

**Fixes**: #804

### DIFF
--- a/docker/drone/src/main/java/org/arquillian/cube/docker/drone/VideoConversionLifecycleManager.java
+++ b/docker/drone/src/main/java/org/arquillian/cube/docker/drone/VideoConversionLifecycleManager.java
@@ -19,13 +19,16 @@ public class VideoConversionLifecycleManager {
     @Inject
     Instance<SeleniumContainers> seleniumContainersInstance;
 
-    public void startConversion(@Observes AfterSuite afterSuite, CubeRegistry cubeRegistry) {
+    public void startConversion(@Observes AfterSuite afterSuite, CubeDroneConfiguration cubeDroneConfiguration,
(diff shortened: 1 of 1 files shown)


docker/drone/src/main/java/org/arquillian/cube/docker/drone/VideoConversionLifecycleManager.java
```diff
@@ -19,13 +19,16 @@
...
-
-        initConversionCube(cubeRegistry);
-        flv2mp4.create();
-        flv2mp4.start();
-
-        afterConversionEvent.fire(new AfterConversion());
+    public void startConversion(@Observes AfterSuite afterSuite, CubeDroneConfiguration cubeDroneConfiguration,
+        CubeRegistry cubeRegistry) {
+
+        if (cubeDroneConfiguration.isRecording()) {
```
bartoszmajsak: Alternatively, you could reverse an if and `return;`  but that's more a matter of style.

georgwolf: I chose this way to be consistent with VncRecorderLifecycleManager where the check is already implemented :)

bartoszmajsak:
Thank you!

bartoszmajsak:
Hey @georgwolf can you contact me by email (`git show 44218b10 | grep Author
` against the latest `master` of Cube ;))? We would like to give you kudos and recognition you so much deserve on http://arquillian.org/

PR #808, opened 2017-09-18
Fix namespace visitor

iocanel:
The idea is that we only want to set namespace on metadata (ObjectMeta) and not on ObjectReferences.

### DIFF
--- a/kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/api/ResourceInstaller.java
+++ b/kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/api/ResourceInstaller.java
@@ -18,16 +18,6 @@ public interface ResourceInstaller extends WithToImmutable<ResourceInstaller> {
     List<HasMetadata> install(URL url);
 
     /**
-     * Installs the resources found in the specified URL.
-     *
-     * @param url
-     *     The URL to read resources from.
-     *
-     * @return The list with the created resources.
-     */
-    List<HasMetadata> install(URL url, List<Class<? extends io.fabric8.kubernetes.api.builder.v2_6.Visitor>> excludedVisitors);
-
-    /**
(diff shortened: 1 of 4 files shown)


lordofthejars:
The only problem I see is that can be sure that is enough? I mean that there is no ObjectReferences that requires to be visited?

iocanel:
I took a look at the kuberntes-model and the only cases that I see a user might specify an object reference are the following:

- as part of an image source (which can be used in many different place including DeploymentConfig)
- as part of the persistent volume claim spec (for pre-binding: https://docs.openshift.org/latest/dev_guide/persistent_volumes.html#persistent-volumes-volumes-and-claim-prebinding).

My understanding is that in both cases, we want to avoid messing with the namespace, since its for referring existing objects rather than anything else. 

We can take it one step further and create a visitor that will also validate that the object exists, and if not display a warning...

What I am not sure about, is if this is enough to address fix the RHOAR issue. I tried to run it locally but it doesn't even build for me (with or without this change).

lordofthejars:
I see then, let me do next, I'll clone your branch and I tried. If it works I squash the PR, if not I tell you what's happening.

iocanel:
Awesome, I am trying this again, but it seems to get stuck in the s2i build.

lordofthejars:
It works, and I think your solution is cleaner than mine. So let's merge :D Thank you so much.

iocanel:
@lordofthejars: Awesome! After that, where we stand in terms of fmp support? Do we need to do more in that front?

lordofthejars:
Probably. My idea is to take the bits of core f8 maven plugin and create an integration with cube. So instead of generating resources as a part of maven lifecycle be as a part of test. So in that case everything will occurs in the same namespace. But I suspect it is going to be a long task.

iocanel:
@lordofthejars: I'll comment on the issue.

PR #817, opened 2017-10-05
Fix Kubernetes-Openshift compatibility

### LINKED ISSUE #815: Make tests work on Openshift and Kubernetes without changes
We were using fabric8-arquillian in quickstarts and now we are switching to Arquillian Cube Kubernetes.
I've managed to migrate [one quickstart](https://github.com/fabric8-quickstarts/spring-boot-camel) by just changing the dependencies in the `pom.xml` file.

However, previously tests were working without changes on both Openshift and Kubernetes, while now:

**On Openshift**
The project should depend explicitly on `arquillian-cube-openshift`.
An additional flag must be used (`enableImageStreamDetection=true`) to ensure that imagestreams are correcly created in the testing project. I've put it in the [surefire configuration for simplicity](https://github.com/fabric8-quickstarts/spring-boot-camel/blob/8735321adb8b5253bc91a16d55ee80c7d702f2fe/pom.xml#L131).

**On Kubernetes**
*Need not* to depend on `arquillian-cube-openshift`, but on `arquillian-cube-kubernetes`. The sole dependency on `arquillian-cube-openshift` turns on a "openshift mode".
*Need not* to detect image streams.

It would be better if:
- `arquillian-cube-openshift` can detect if we are connected to a standard kubernetes cluster and delegate to `arquillian-cube-kubernetes` (that is imported transitively) all processing steps.
- The `enableImageStreamDetection` flag can be defaulted to `true` when running on Openshift.

If this is ok, I can work on it.

Also: it is not related to this PR, but since you had problems with failsafe in #743 , I can tell you that surefire works like a charm.

nicolaferraro:
This fixes #815.

Defaulted the imagestream detection to `true` when running on Openshift.

Deferred the choice of `kubernetes.json`/`openshift.json` file to a later stage (if not explicitly set by the user) when it's known if we are connected to a Kubernetes or Openshift cluster.

### DIFF
--- a/docs/kubernetes.adoc
+++ b/docs/kubernetes.adoc
@@ -86,7 +86,7 @@ For example **foo.bar.baz** will be converted to **FOO_BAR_BAZ**.
 | definitionsFile                     | String         | Any | Definitions file path
 | proxiedContainerPorts               | List           | Any | Comma Separated List following Pod:containerPort OR Pod:MappedPort:ContainerPort
 | routerHost                          | String         | Any | Define OpenShift router address thats is used to resolve the pod's address and to get its route
-| enableImageStreamDetection          | Bool (true)    | Any | Define if you want that Arquillian Cube apply Image Stream files (`*-is.yml`) before deploying
+| enableImageStreamDetection          | Bool (true)    | Any | Define if you want that Arquillian Cube apply Image Stream files (`*-is.yml`) to Openshift before deploying
 |===============================================================================================================================================
 
 
--- a/docs/kubernetes.adoc
+++ b/docs/kubernetes.adoc
@@ -86,7 +86,7 @@ For example **foo.bar.baz** will be converted to **FOO_BAR_BAZ**.
 | definitionsFile                     | String         | Any | Definitions file path
 | proxiedContainerPorts               | List           | Any | Comma Separated List following Pod:containerPort OR Pod:MappedPort:ContainerPort
 | routerHost                          | String         | Any | Define OpenShift router address thats is used to resolve the pod's address and to get its route
-| enableImageStreamDetection          | Bool (false)   | Any | Define if you want that Arquillian Cube apply Image Stream files (`*-is.yml`) before deploying
+| enableImageStreamDetection          | Bool (true)    | Any | Define if you want that Arquillian Cube apply Image Stream files (`*-is.yml`) before deploying
 |===============================================================================================================================================
 
 
--- a/kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/api/Configuration.java
+++ b/kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/api/Configuration.java
@@ -73,8 +73,6 @@ public interface Configuration {
 
     URL getEnvironmentConfigUrl();
 
-    List<URL> getEnvironmentConfigAdditionalUrls();
-
     List<URL> getEnvironmentDependencies();
 
     String getSessionId();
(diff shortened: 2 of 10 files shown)


docs/kubernetes.adoc
```diff
@@ -86,7 +86,7 @@ For example **foo.bar.baz** will be converted to **FOO_BAR_BAZ**.
 | definitionsFile                     | String         | Any | Definitions file path
 | proxiedContainerPorts               | List           | Any | Comma Separated List following Pod:containerPort OR Pod:MappedPort:ContainerPort
 | routerHost                          | String         | Any | Define OpenShift router address thats is used to resolve the pod's address and to get its route
-| enableImageStreamDetection          | Bool (false)   | Any | Define if you want that Arquillian Cube apply Image Stream files (`*-is.yml`) before deploying
+| enableImageStreamDetection          | Bool (true)    | Any | Define if you want that Arquillian Cube apply Image Stream files (`*-is.yml`) before deploying
```
lordofthejars: I think it should say here that is false in case of kubernetes and true in case of OpenShift

lordofthejars:
Thank you very much for your contribution which makes Cube even better, just some small modifications before merge :)

docs/kubernetes.adoc
Ladicek: Looking at the code, it seems it's actually `true` always...? (I don't mind. I think it probably _should_be `true` by default, because it's required when running tests in a new Cube-managed namespace, which is the default behavior.)

nicolaferraro: Yeah, I think it's an error to set it always to `true`, imagestream detection should not be enabled on Kubernetes. I was running tests with `mvn clean ...` so I didn't notice that if for some reason a `*-is.yml` file is on the build dir it's applied on Kubernetes too, even if Kubernetes does not support them.

Ladicek: I might be wrong here. The `*-is.yml` files are only searched for in `CubeOpenShiftConfiguration`. Which, before this PR, was never supposed to be in place when using Kubernetes. With this PR, when you want to bypass all the OpenShift stuff when running against Kubernetes, I'm not sure.

nicolaferraro: The problem is.. I looked at how to bypass completely the openshift stuff when I'm connected to kubernetes, but the Openshift `LoadableExtension` overrides some services before knowing if we are running on Kube or Openshift. Using the kubernetes client there to find the cluster type is not allowed, because the kubernetes client needs the configuration and the configuration depends on the cluster type (a typical chicken-egg problem).
There were already many calls to `kubernetesClient.isAdaptable(OpenshiftClient.class)` and they were meant to avoid doing openshift-related tasks on Kubernetes. I used the same reasoning on the resource locator.
So the solution towards I'm going is to make openshift stuff behave like Kubernetes stuff (many Openshift objects have a "is a" relationship with Kubernetes objects). So I'm trying to refactor a bit the configuration..

nicolaferraro:
I've moved the additional resources declaration to the locator, since that config was not exposed to the user.

This way the imagestream detection is enabled by default, but it has no effect on Kubernetes. I made it clearer in the description.

lordofthejars:
@nicolaferraro I think we are almost done, if you agree I resolve the conflicts and I merge this PR.

nicolaferraro:
Ok, no more changes from my side. Thanks!

lordofthejars:
@nicolaferraro can you resolve merge conflicts, I wanted to do but since there were multiple changes, I don't want to break anything :)

nicolaferraro:
Sure, will do!

PR #818, opened 2017-10-06
fix config value lookup ordering

### LINKED ISSUE #816: properties in arquillian.xml take precedence over system properties
##### Issue Overview

I have `arquillian.xml` looking like this:

```xml
<arquillian>
    <property name="namespace.use.current">true</property>
    <property name="env.init.enabled">true</property>
  </extension>
</arquillian>
```

(Ignore the `env.init.enabled` bit, I know it defaults to `true`, I include it just to make it clear that I depend on that behavior.)

I intentionally set `namespace.use.current` to `true` here so that the tests _default_ to running in current namespace. But sometimes, I want Cube to create the temporary `itest-xxxx` namespace and use it, and I'm invoking Maven with `-Dnamespace.use.current=false` to do that.

##### Expected Behaviour

A temporary namespace is created and used.

##### Current Behaviour

Current namespace is used even though I specify `-Dnamespace.use.current=false`.

##### Steps To Reproduce
1. git clone -b arquillian-cube http://github.com/Ladicek/wfswarm-rest-http/
2. cd wfswarm-rest-http
3. mvn clean verify -Popenshift,openshift-it -Dnamespace.use.current=false -DenableImageStreamDetection=true

##### Additional Information

This is caused by the `org.arquillian.cube.impl.util.ConfigUtil` implementation. The first method's javadoc actually says

```
The lookup order is system > env > map > defaultValue.
```

but the implementation doesn't conform to that javadoc.

I can send a PR, it's a pretty simple change, but I want to confirm first that it's really a bug. (I think it is.)

Ladicek:
Fixes #816.

### DIFF
--- a/core/src/main/java/org/arquillian/cube/impl/util/ConfigUtil.java
+++ b/core/src/main/java/org/arquillian/cube/impl/util/ConfigUtil.java
@@ -23,11 +23,10 @@ public class ConfigUtil {
      *     The value that should be used if property is not found.
      */
     public static String getStringProperty(String name, Map<String, String> map, String defaultValue) {
-        if (map.containsKey(name)) {
-            return map.get(name);
-        } else {
-            return getPropertyOrEnvironmentVariable(name, defaultValue);
+        if (map.containsKey(name) && Strings.isNotNullOrEmpty(map.get(name))) {
+            defaultValue = map.get(name);
         }
+        return getPropertyOrEnvironmentVariable(name, defaultValue);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Ladicek:
Fixed `getKubernetesConfigurationUrl`, the test failure revealed an actual issue, yay tests!

lordofthejars:
@Ladicek So I need to wait before merging this PR isn't it?

Ladicek:
Do you? The tests are green now.

lordofthejars:
Ok, I didn't understand if you did more modifications or not. 

Cool.

Ladicek:
Ah, sorry. I broke one test originally, but I fixed the code and all tests pass now. So this is ready to go from my POV.

lordofthejars:
Thank you so much.

PR #819, opened 2017-10-16
add @AwaitRoute

Ladicek:
The process that injects `@RouteURL`-s now awaits availability
of those routes. This can be configured using the `await`
parameter of the `@RouteURL` annotation.

I'm sending this PR just to start a discussion. I know Cube already awaits pod readiness, but this awaits the _route_. This is because there might be slight delay between pod becoming ready and being exposed on the router, and actually there's also a bug (I believe) on OpenShift Online where the delay between pod readiness and route availability can be several minutes!

### DIFF
--- a/openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/enricher/AwaitRoute.java
+++ b/openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/enricher/AwaitRoute.java
@@ -1,0 +1,39 @@
+package org.arquillian.cube.openshift.impl.enricher;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+import java.util.concurrent.TimeUnit;
+
+import javax.inject.Qualifier;
+
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * Used inside {@link RouteURL} to configure awating the route URL to be available.
+ * Availability is defined as returning a known good HTTP status code.
+ */
+@Qualifier
+@Documented
+@Retention(RUNTIME)
+@Target({})
+public @interface AwaitRoute {
+    /**
+     * Path that should be appended to the route URL for checking route availability.
+     * Useful when there's nothing exposed directly at the root route URL, only on some paths below.
+     * Defaults to {@code /}.
+     */
+    String path() default "/";
+
+    /**
+     * Set of HTTP status codes that are considered good. Defaults to a single value, {@code 200}.
+     */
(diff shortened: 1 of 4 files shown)


Ladicek:
CC @cescoffier @alesj @spisiakm

Ladicek:
Also adding the await code directly into the class that performs `@RouteURL` injection is probably a ... not-very-good idea, I'd be happy to refactor this.

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/enricher/AwaitRoute.java
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Used inside {@link RouteURL} to configure awating the route URL to be available.
+ * Availability is defined as returning a known good HTTP status code.
+ */
+@Qualifier
+@Documented
+@Retention(RUNTIME)
+@Target({ElementType.FIELD, ElementType.PARAMETER})
+public @interface AwaitRoute {
```
lordofthejars: What's happening if there is nothing in default root location? I think an attribute like `context` should be added, which by default is set to `/`

Ladicek: Ah totally makes sense!

lordofthejars:
Yeah after that, I think that just need some sentence on documentation about this annotation and we are fine.

Ladicek:
Added a `path` parameter and some docs.

But I'm now thinking this is a breaking change. You're totally right that the root location doesn't have to respond with 200 by default, in which case `@RouteURL` injection will now fail when it used to work just fine.

I'm thinking maybe the `@AwaitRoute` annotation shouldn't be part of `@RouteURL` but standalone -- and if it isn't present, no route awaiting happens. WDYT?

lordofthejars:
I like the idea of being hidden by default because then user don't need to code more things than the strictly necessary. I am thinking that probably with Route injection you always need, so for me it seems logical to have it there by default. I agree that the root should not return `200` by default, but let's stick with this default. User can always configure the `path` and the `return code` to expect.

Ladicek:
If you're fine with that, then I'm too :-)

lordofthejars:
Yes I think that if we can give a default to user, his UX will be better. Also in terms of boosters might help.

Ladicek:
Fixed the `@Target` annotation of `@AwaitRoute`, I think this would be ready to go.

lordofthejars:
Thank you very much.

Ladicek:
Thank you!

PR #840, opened 2017-11-06
improve route URL injection

### LINKED ISSUE #837: the @AwaitRoute mechanism probably doesn't deal nicely with self-signed HTTPS certs
A few weeks ago, I added a mechanism to await routes injected with `@RouteURL`. That mechanism is always applied and one can't opt-out, so it should better be reliable. However, I believe it won't deal nicely with self-signed HTTPS certificates, and maybe in some other cases.

I'm now thinking this mechanism should be opt-in; i.e., the `@AwaitRoute` annotations should be _outside_ of `@RouteURL` and the route awaiting mechanism should only kick in when the annotation is present.

Obviously I volunteer to investigate if the self-signed HTTPS cert suspicion is real and if yes, send a PR with solution proposed above.

### LINKED ISSUE #838: @RouteURL should have a parameter to customize the path portion of the URL
In my case, I often need to create URLs to the endpoints of tested application. Because `@RouteURL` injects a URL of the form `http://route.host/` (note the `/` at the end), I need to construct the URL by something like `routeUrl + "api/info"`. I can't do `routeUrl + "/api/info"`, as that would result in `http://route.host//api/info`, which is wrong.

I propose adding a `path` parameter to `@RouteURL` so that one can write `@RouteURL(value = "route-name", path = "/api/info")` and the injected URL would be directly `http://route.host/api/info`. In my case, I often need just a single endpoint in one test, that would be perfectly fine. In case someone needs to construct multiple URLs, they can either keep doing `routeUrl + "api/info"`, or define `@RouteURL(value = "route-name", path = "")` and do `routeUrl + "/api/info"`, which is easier on the eyes.

### LINKED ISSUE #839: @RouteURL shouldn't add a port number when it's the default for the protocol
The `@RouteURL` injection always adds a port number to the URL, even if it's the default for given protocol. E.g. instead of `http://route.host/`, I always get `http://route.host:80/`.

That's 100% correct, but unexpected. And it actually causes me one issue: in case the route leads to Keycloak and I use it to verify a token that was issued by that Keycloak server, the verification will most likely fail because the token was most likely issued by going through `https://sso.server/xxx` and not `https://sso.server:443/xxx`.

Ladicek:
This commit improves `@RouteURL` injection in these ways:
- The `@AwaitRoute` annotation is now external to `@RouteURL`, so
  that the entire route await mechanism is opt-in. This isn't
  a breaking change because the `@AwaitRoute` mechanism doesn't
  exist in any release.
- The `@RouteURL` annotation now has a parameter to customize
  the path portion of the injected URL.
- the `@RouteURL` injection now doesn't add a port number to the URL
  if it's the default value for given protocol. This means the URLs
  will be more inline with users expectations.

Fixes #837, fixes #838 and fixes #839.

### DIFF
--- a/docs/enrichers.adoc
+++ b/docs/enrichers.adoc
@@ -70,9 +70,9 @@ When running Arquillian Cube against a OpenShift's pod you can retrieve the rout
 @RouteURL("app-name")
 private URL url;
 ----
-When Arquillian Cube injects the route URL, it also checks that the URL is available.
+This annotation also has a `path` parameter which can change the path part of the injected URL.
+You can also use an additional annotation `@AwaitRoute` to wait until the route becomes available.
 That is, it responds with a known good HTTP status code in given timeout.
-This can be configured using the `await` parameter of the `@RouteURL` annotation.
 
 If the route is not resolvable, you need to set the `routerHost` setting to the IP address of the OpenShift router.
 You can configure it in arquillian.xml:
--- a/openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/enricher/AwaitRoute.java
+++ b/openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/enricher/AwaitRoute.java
@@ -1,6 +1,7 @@
 package org.arquillian.cube.openshift.impl.enricher;
 
 import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
 import java.util.concurrent.TimeUnit;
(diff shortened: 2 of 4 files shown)


lordofthejars:
@Ladicek I have tried to fix the conflict, but after fixing it I don't know why github has not detected the change. https://github.com/arquillian/arquillian-cube/pull/840/commits/9c9d2024de5bf555945502f0e8867497dd35e778 Can you fix the conflict, you can use my commit as guide.  Thank you very much.

Ladicek:
Thanks @lordofthejars, rebased.

PR #857, opened 2017-11-21
fix(#850): @OpenShiftResource supports yaml format

### LINKED ISSUE #850: @OpenShiftResources only support json format
##### Issue Overview

When using `@OpenShiftResouce` annotation, only `json` format is supported when also `yaml` should be supported.

##### Expected Behaviour

`json` and `yaml` formats supported

##### Current Behaviour

Only `json` supported.

lordofthejars:

@OpenShiftResource supports `yaml` format

- Adds snakeyaml to process yaml format in case of cannot be parsed as json

Fixes #850

### DIFF
--- a/openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/fabric8/F8OpenShiftAdapter.java
+++ b/openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/fabric8/F8OpenShiftAdapter.java
@@ -86,7 +86,6 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.StringTokenizer;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.logging.Level;
 import okhttp3.Response;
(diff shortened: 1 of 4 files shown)


lordofthejars:
@jwendell @alesj can you review this PR? I think it is fairly simple, but a review is always welcome.

lordofthejars:
I thought the same, but notice that you receive an inputstream because it should not be a always a file, it can come from hardcoded string or classpath resource, https://github.com/arquillian/arquillian-cube/blob/master/openshift/api/src/main/java/org/arquillian/cube/openshift/api/OpenShiftResource.java#L38

So I am not sure any other way. I know is a bit ugly but I don't see any other way.

alesj:
Yeah, wrt the stream, I was thinking more in the way of something like (pseudo code ...)
`Yaml yaml = new Yaml(stream); boolean isReallyYaml = yaml.isYaml();`
(same for json)

But np, as long as it works ... or if that's the best we can do.

iocanel:
The kubernetes-client allows you to easily load any kind of resource from an inputstream without much fuss using `client.load(is).createOrReplace()`.

So we have an opportunity here to remove some custom code (e.g. the resource handles etc).

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/fabric8/F8OpenShiftAdapter.java
```diff
@@ -473,17 +474,16 @@ protected OpenShiftResourceHandle createResourceFromStream(InputStream stream) t
             String content = IOUtils.toString(stream, StandardCharsets.UTF_8);
 
             try {
-                // TODO only works if JSON should work with YAML as well.
-                ModelNode json;
-                json = ModelNode.fromJSONString(content);
+                final ModelNode json = ModelNode.fromJSONString(content);
```
iocanel: This try/catch block can be replaced by `client.load(stream).createOrReplace()`. Where client is an instance of the kubernetes-client.

Potentiall allowing us to get rid of snakeyml, dmr and maybe commons-io?

lordofthejars: @iocanel I have updated the code with your suggestion. I have run tests and it works, but can you review?

iocanel:
@lordofthejars: The code looks great! 

I think that we can now ditch the snakeyaml dependency too and maybe dmr and commons-io.

lordofthejars:
`jboss-dmr` not because it is used in other places. The others true.

lordofthejars:
Done, if you agree now I can merge it.

lordofthejars:
@iocanel I have squashed the PR (to avoid merge conflicts in the future) but if you find something please comment here and I'll fix too.

PR #861, opened 2017-11-21
feat(#830): support oc or kubectl command to execute as a part of your tests.

### LINKED ISSUE #830: Provide a ‘oc’ command proxy
I often need to execute ‘oc’ commands (such as ‘oc create -f ..., or oc new-app ...) in my test. Most of the time it’s to execute the exact same instructions we give to our user and so check that the doc is still valid.

It would be great to have a utility class calling ‘oc’ behind the scene. For instance, method allowing to execute oc command, retrieve the result (status) and output (StrIng or byte[]).

I can also see more and more usage of such a feature with the service catalog (with the new bind and unblind commands).

dipak-pawar:
Now you can execute any `oc` or `kubectl` command as a part of your tests.

-  Adds `CommandExecutor` class to execute any command using spacelift

#### Usage
Add kubernetes-client dependency on classpath
```xml
      <dependency>
        <groupId>org.arquillian.cube</groupId>
        <artifactId>arquillian-cube-kubernetes</artifactId>
        <version>${project.version}</version>
        <scope>test</scope>
      </dependency> 
```
Write a test like following
```java
    @Test
    public void should_be_able_deploy_resources_using_oc() {
       // given
        final CommandExecutor commandExecutor = new CommandExecutor();

        // when
        final List<String> resources = commandExecutor.execCommand("oc create -f https://goo.gl/qaiGRJ");

        // then
        assertThat(resources).contains("service \"hello-world\" created", "deployment \"hello-world\" created");
    }
```

Fixes #830

### DIFF
--- a/docs/kubernetes.adoc
+++ b/docs/kubernetes.adoc
@@ -402,33 +402,7 @@ Let's see how can you execute `oc` or `kubectl` commands as a part of your test.
 [source, java]
 .OpenshiftAndK8sExample.java
 ----
-public class HelloWorldTest {
-
-    private CommandExecutor commandExecutor;
-
-    @Before
-    public void initCommandExecutor() {
-        commandExecutor = new CommandExecutor();
-    }
-
-    @Test
-    public void should_be_able_deploy_resources_using_oc() {
-        // when
-        final List<String> resources = commandExecutor.execCommand("oc create -f https://goo.gl/qaiGRJ");
-
-        // then
-        assertThat(resources).contains("service \"hello-world\" created", "deployment \"hello-world\" created");
-    }
-
-    @Test
-    public void should_be_able_get_namespace_using_kubectl() {
-        // when
-        final List<String> namespaces = commandExecutor.execCommand("kubectl get ns -o=name");
-
-        // then
-        assertThat(namespaces).contains("namespaces/default");
-    }
-}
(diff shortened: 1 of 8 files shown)


openshift/ftest-oc-proxy/src/test/java/org/arquillian/cube/openshift/ftest/HelloWorldTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+    @Before
+    public void initCommandExecutor() {
+        commandExecutor = new CommandExecutor();
+    }
+
+    @Test
+    public void should_be_able_deploy_resources_using_oc() {
+        // when
+        final List<String> resources = commandExecutor.execCommand("oc create -f https://goo.gl/qaiGRJ");
```
bartoszmajsak: Can't we have this file referred locally? Otherwise, we would always be online to run this test.

docs/kubernetes.adoc
```diff
@@ -397,6 +397,40 @@ public class HelloWorldTest {
 }
 ----
 
+Let's see how can you execute `oc` or `kubectl` commands as a part of your test.
+
+[source, java]
+.OpenshiftAndK8sExample.java
```
bartoszmajsak: Can we include already existing test instead of having it written down explicitly here? This will let us avoid updating both the code and documentation while evolving the code of this feature.

lordofthejars:
I agree with changes purposed by Bartosz. Nothing else to fix.

PR #868, opened 2017-11-23
feat(#867): provide an option to skip readiness await

### LINKED ISSUE #867: provide an option to skip readiness await
##### Issue Overview

The readiness await as implemented in Fabric8 Kubernetes Client isn't perfect. For example, we found this issue: https://github.com/fabric8io/kubernetes-client/issues/921

I believe that there should be an option to skip this readiness await, relying instead on `@AwaitRoute`. `@AwaitRoute` doesn't wait for state change of some (possibly even internal) Kubernetes/OpenShift resources, but instead waits for actual application HTTP response.

By default, awaiting readiness would be on, and `@AwaitRoute` is already opt-in. I don't propose changing current behavior.

##### Expected Behaviour

Able to turn off readiness await, to workaround Fabric8 Kubernetes Client issues.

##### Current Behaviour

Readiness await is always performed.

Ladicek:
Fixes #867.

### DIFF
--- a/docs/kubernetes.adoc
+++ b/docs/kubernetes.adoc
@@ -83,7 +83,8 @@ environment
 the environment
 | env.dependencies                    | List           | Any | Whitespace separated list of URLs to more dependency 
 kubernetes.json             
-| wait.timeout                        | Long (5mins)   | Any | The total amount of time to wait until the env is ready       
+| wait.enabled                        | Bool (true)    | Any | Whether to wait until the env is ready
+| wait.timeout                        | Long (5mins)   | Any | The total amount of time to wait until the env is ready
 | wait.poll.interval                  | Long (5secs)   | Any | The poll interval to use for checking if the environment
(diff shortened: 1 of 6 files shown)


lordofthejars:
My question is, could we wait until `kubernetes-client` implements the fix or do we need in cube? If we need in Cube but will be fixed later, how about creating an issue to remove the code when the issue is fixed on kubernetes-client?

Ladicek:
That's a good question. I have no visibility into Fabric8 Kubernetes client, but from my experience, I'd guess that fixing that will take weeks (or even months).

I'd personally consider the ability to disable readiness await a feature, no matter how it's used, and I wouldn't want to remove it. On the other hand, of course, the less configuration options, the better. So I'm kinda stuck here :-)

lordofthejars:
Ok it has sense if it takes months. I will review the code, and then merge it.

PR #870, opened 2017-11-24
Do not log events upon failure, we have "logs.copy" property for that

jwendell:
This piece of code was added when we did't have the ability to copy the
logs from pods. Now we have it, so, it's up to the user enable in order
to debug issues.

See https://github.com/jboss-openshift/ce-arq/commit/e68f78ae54b7a0fad4b81a0040faa1eebf61a0b6


lordofthejars:
Thanks :)

PR #875, opened 2017-11-27
fix(#851): Unify names of annotations to be more consistent with plural

### LINKED ISSUE #851: Use `@Repeatable` annotations in OpenShift annotations
##### Issue Overview

Some annotations such as `@Templates`, `@OpenShiftResources, ... does not using `@Repeatable` approach.

##### Expected Behaviour

Implement annotations with `@Repeatable`

##### Current Behaviour

Uses old way (prior to Java8)

lordofthejars:

Unify names of annotations to its plural instead of `wrapper` and use `@Repeatable` in these classes.

Fixes #851

### DIFF
--- a/openshift/api/src/main/java/org/arquillian/cube/openshift/api/AddRoleToServiceAccounts.java
+++ b/openshift/api/src/main/java/org/arquillian/cube/openshift/api/AddRoleToServiceAccounts.java
@@ -33,6 +33,6 @@ import java.lang.annotation.Target;
  */
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.TYPE)
-public @interface AddRoleToServiceAccountWrapper {
+public @interface AddRoleToServiceAccounts {
     AddRoleToServiceAccount[] value();
 }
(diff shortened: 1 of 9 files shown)


lordofthejars:
@Ladicek you can review the names and if you agree I'll merge it.

openshift/api/src/main/java/org/arquillian/cube/openshift/api/RoleBindings.java
```diff
@@ -24,6 +24,7 @@
 package org.arquillian.cube.openshift.api;
 
 import java.lang.annotation.ElementType;
+import java.lang.annotation.Repeatable;
```
Ladicek: Unused import?

openshift/api/src/main/java/org/arquillian/cube/openshift/api/AddRoleToServiceAccounts.java
```diff
@@ -33,6 +33,6 @@
  */
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.TYPE)
-public @interface AddRoleToServiceAccountWrapper {
+public @interface AddRoleToServiceAccounts {
```
Ladicek: Hmm I see why they called it a `Wrapper` instead of using simple plural. Maybe call it `AddRolesToServiceAccounts`?

lordofthejars: so plural roles and plural accounts

Ladicek: Right.

lordofthejars:
Done

PR #877, opened 2017-11-28
fix(#876): Set wait enables to true for openshift env to be ready

### LINKED ISSUE #876: Cube should wait for openshift by default to initialize environment
##### Issue Overview
Currently openshift configuration is not setting default wait enabled to true due to which we are getting NPE to inject resources in test 

##### Expected Behaviour
cube-openshift should wait for openshift resources to be in ready state

##### Current Behaviour
Cube openshift is not waiting for openshift resource to be in ready state
##### Steps To Reproduce
1. cd openshift/ftest-standalone 
2. mvn test.

dipak-pawar:
As a part of https://github.com/arquillian/arquillian-cube/pull/868/files, we are setting waitEnabled to true only for DefaultKubernetes configuration which is working fine for k8s extension. Check code [here](https://github.com/arquillian/arquillian-cube/pull/868/files#diff-6532ea423d1c6b77a83466b11a4b6240R150)

We should set it to `CubeOpenShiftConfiguration`

Fixes #876


Ladicek:
Ah I totally missed this! Thank you!

PR #878, opened 2017-11-28
fix(#869): update core to latest & remove unnecessary dependencies.

### LINKED ISSUE #869: ResourceProvider of type Url couldn't lookup for KuberntesServiceUrlResourceProvider if classpathcontains multiple resourceProviders of url type
##### Issue Overview
If user is trying to Inject Service url using @ArquillianResource with `@PortForward`. Arquillain cube couldn't lookup for required url

`
    @Named("hello-openshift-service")
    @PortForward
    @ArquillianResource
    URL url;
`

##### Expected Behaviour
Cube should be able to lookup for service url

##### Current Behaviour
Cube is returning null after doing lookup 

##### Steps To Reproduce
1. Clone this repository
2. `cd openshift/ftest-standalone`
3. Run mvn clean tests from terminal
4. Run tests from ftest-standalone from Intellij

Already opened issue for core - https://github.com/arquillian/arquillian-core/issues/140

dipak-pawar:
Fixes #869

### DIFF
--- a/openshift/ftest-standalone/pom.xml
+++ b/openshift/ftest-standalone/pom.xml
@@ -31,15 +31,5 @@
       <artifactId>assertj-core</artifactId>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>io.fabric8</groupId>
-      <artifactId>openshift-client</artifactId>
-      <version>2.6.3</version>
-    </dependency>
-    <dependency>
-      <groupId>io.fabric8</groupId>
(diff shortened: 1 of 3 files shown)


lordofthejars:
Dipak there is an error in this PR, please do not merge them without having a review before. Notice that you are changing an ftest to use internal classes, which this is not what a normal user would do it. Ftest should never contain the internal bundled classes, but the public ones (not uber ones)

PR #884, opened 2017-11-28
feat(#860): Set OpenShift resources at method level

### LINKED ISSUE #860: Set @OpenShiftResources at method level
##### Issue Overview

Set `@OpenShiftResources` at method level.

##### Expected Behaviour

Set `@OpenShiftResources` at method level so user can define resources for test method execution instead of all class.

##### Current Behaviour

You can only set annotation at class level.

lordofthejars:

You can set OpenShift resources using `@OpenShiftResources` annotation at method level, so resources are created before the test and deleted after the test.

- `@OpenShiftResources` at method level
- Observers for creating resources before and after each test method execution.

Fixes #860

### DIFF
--- a/openshift/ftest-openshift-resources-standalone/src/test/java/org/arquillian/cube/openshift/standalone/HelloWorldOpenShiftResourcesTest.java
+++ b/openshift/ftest-openshift-resources-standalone/src/test/java/org/arquillian/cube/openshift/standalone/HelloWorldOpenShiftResourcesTest.java
@@ -1,5 +1,7 @@
 package org.arquillian.cube.openshift.standalone;
 
+import io.fabric8.openshift.api.model.RouteList;
+import io.fabric8.openshift.client.OpenShiftClient;
 import java.io.IOException;
 import java.net.URL;
 import okhttp3.OkHttpClient;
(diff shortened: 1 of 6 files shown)


openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/resources/OpenShiftResourceFactory.java
```diff
@@ -153,6 +175,18 @@ public static void deleteResources(String resourcesKey, OpenShiftAdapter adapter
         }
     }
 
+    public static void deleteReosurces(Class<?> testClass, Method testMethod, OpenShiftAdapter client) {
```
jwendell: Typo

openshift/ftest-openshift-resources-standalone/src/test/java/org/arquillian/cube/openshift/standalone/HelloWorldOpenShiftResourcesTest.java
```diff
@@ -35,4 +41,12 @@ public void should_show_hello_world() throws IOException {
         assertThat(response.code()).isEqualTo(200);
         assertThat(response.body().string()).isEqualTo("Hello OpenShift!\n");
     }
+
+    @Test
+    @OpenShiftResource("classpath:hello-route-2.yaml")
+    public void should_register_extra_route() {
+        final RouteList routes = openShiftClient.routes().list();
+        assertThat(routes.getItems()).hasSize(2);
```
jwendell: It would be nice if we test some metadata about the route just created, instead of testing the number of routes. This way this test wouldn't break if we add other routes in the future (in the class level).

PR #885, opened 2017-11-29
Cube CI is using oc cluster up to run all ftests from k8s & openshift

### LINKED ISSUE #882: [Enhancement] Improve Cube CI to run all ftests from k8s & openshift
##### Issue Overview
Currently we have ftests for openshift & k8s which are running with
 ```java
@RunWith(ArquillianConditionalRunner.class)
@RequiresKubernetes
```

due to this we are skipping all ftests from kubernetes & openshift.

##### Expected Behaviour
All ftests should run fine.

##### Current Behaviour
ftests are skipped.

### PR DESCRIPTION
**Pull request by dipak-pawar**
Running `oc cluster up` on travis.

 - Starting cluster using `oc cluster up` to start openshift cluster.
- Using matrix to run tests on different cluster

Fixes #882

### DIFF
--- a/openshift/ftest/src/test/java/org/arquillian/cube/openshift/ftest/HelloPodDeploymentOpenShiftTestCase.java
+++ b/openshift/ftest/src/test/java/org/arquillian/cube/openshift/ftest/HelloPodDeploymentOpenShiftTestCase.java
@@ -9,7 +9,6 @@ import org.jboss.shrinkwrap.api.ShrinkWrap;
 import org.jboss.shrinkwrap.api.asset.EmptyAsset;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
 import org.junit.Assert;
-import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: openshift/ftest/src/test/java/org/arquillian/cube/openshift/ftest/HelloPodDeploymentOpenShiftTestCase.java [hunk 1]
```diff
@@ -9,11 +9,13 @@
...
 import org.jboss.shrinkwrap.api.spec.WebArchive;
 import org.junit.Assert;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 
 @RunWith(ArquillianConditionalRunner.class)
 @RequiresOpenshift
+@Ignore
 public class HelloPodDeploymentOpenShiftTestCase {
```
bartoszmajsak: We can put the explanation to `@Ignore` annotation.

The only thing I'm concerned about is - if we ignore it - it's equal to just deleting it in this case. Can't we fix it so we can run in the CI? Or maybe write another test which is less resource consuming?

bartoszmajsak:
Good stuff @dipak-pawar! Many thanks for such a big improvement with such a small change! :1st_place_medal:

openshift/ftest/src/test/java/org/arquillian/cube/openshift/ftest/HelloPodDeploymentOpenShiftTestCase.java
dipak-pawar: This is the only one test in which we are not using standalone. We are using `arquillian-junit-container` where we are building docker image as a part of tests from test resources & then deploy it. Let me try to use already built image & run this tests.

PR #888, opened 2017-11-29
fix(#887): Removes HttpClient module

### LINKED ISSUE #887: Removes httpclient module
##### Issue Overview

Removes httpclient module and substitute the places where it is used by `okhttp`

lordofthejars:

Removes HttpClient module and instead use okhttp.

- Removes HttpClient module
- Adapt code to use okhttp

Fixes #887


jwendell:
hmm, we are a heavy user of this...

alesj:
Yeah, I think this just needs to be re-written to use OKHttpClient, instead of ApacheHC.
As CE's testsuite uses this a lot.

lordofthejars:
Already done in PR

jwendell:
I think what Ales said was to change the implementation to use okhttp but keep the interface.

lordofthejars:
Interface was used just in one place, so not sure if it has much sense
maintaing abstraction. Also okhttp offers this kind of abstraction

alesj:
It's used externally - - the interface.
So CE won't have to change all tests.

The abstraction was there to unify http API usage.

lordofthejars:
Ok, but then why an abstraction of an http client should be provided by Cube?

alesj:
@lordofthejars yes, I agree, no need to port that code over to Cube.

PR #890, opened 2017-12-01
Add support for default docker configuration

### LINKED ISSUE #883: docker for mac is not supported by default
Hi,

I am running `org.arquillian.cube:arquillian-cube-docker:1.9.2` together with `org.jboss.arquillian.junit:arquillian-junit-standalone:1.1.13.Final` under mac os and docker for mac:

```
Client:
 Version:      17.09.0-ce
 API version:  1.32
 Go version:   go1.8.3
 Git commit:   afdb6d4
 Built:        Tue Sep 26 22:40:09 2017
 OS/Arch:      darwin/amd64

Server:
 Version:      17.09.0-ce
 API version:  1.32 (minimum version 1.12)
 Go version:   go1.8.3
 Git commit:   afdb6d4
 Built:        Tue Sep 26 22:45:38 2017
 OS/Arch:      linux/amd64
 Experimental: true]
```

My mac is configured to use the standard docker unix socket without TLS.

Trying to run arquillian out of the box does not work and leads to exceptions like `boot2docker: command not found`

docker for mac being the standard way to install docker on a mac, I would expect the framework to work with it without additional configuration.

In the current implementation, the following lines needs to be added to the `arquillian.xml` resource:
```
<?xml version="1.0"?>
<arquillian xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
            xmlns="http://jboss.org/schema/arquillian"
            xsi:schemaLocation="http://jboss.org/schema/arquillian
    http://jboss.org/schema/arquillian/arquillian_1_0.xsd">
    <extension qualifier="docker">
        <property name="serverUri">unix:///var/run/docker.sock</property>
        <property name="tlsVerify">false</property>
        <...>
    </extension>
</arquillian>
```

Docker for mac being the standard way to run docker on mac, could there be a way to support docker for mac by default, without extra configuration?

tjamet:
As discussed in #883, here is an implementation to consider the default docker configuration before running `docker-machine` or `boot2docker`

Fixes #883

### DIFF
--- a/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfigurationResolver.java
+++ b/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfigurationResolver.java
@@ -18,6 +18,8 @@ import java.io.File;
 import java.net.InetAddress;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.nio.file.FileSystems;
+import java.nio.file.Files;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Random;
@@ -58,6 +60,7 @@ public class CubeDockerConfigurationResolver {
      */
     public Map<String, String> resolve(Map<String, String> config) {
         config = resolveSystemEnvironmentVariables(config);
+        config = resolveSystemDefaultSetup(config);
         config = resolveDockerInsideDocker(config);
         config = resolveDownloadDockerMachine(config);
         config = resolveAutoStartDockerMachine(config);
@@ -193,6 +196,17 @@ public class CubeDockerConfigurationResolver {
 
         if (!config.containsKey(CubeDockerConfiguration.TLS_VERIFY) && isDockerTlsVerifySet()) {
             config.put(CubeDockerConfiguration.TLS_VERIFY, Boolean.TRUE.toString());
+        }
+
+        return config;
+    }
+    
+    private Map<String, String> resolveSystemDefaultSetup(Map<String, String> config) {
+        if (!config.containsKey(CubeDockerConfiguration.DOCKER_URI)) {
+            URI uri = URI.create(operatingSystem.getDefaultFamily().getServerUri());
+            if (Files.exists(FileSystems.getDefault().getPath(uri.getPath()))){
+                config.put(CubeDockerConfiguration.DOCKER_URI, operatingSystem.getDefaultFamily().getServerUri());
+            }
         }
 
         return config;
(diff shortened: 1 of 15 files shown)


lordofthejars:
I cannot see in the code where is automatically detected if unix sockets are available or not. I mean in theory in case of Windows and Mac what I would expect is something like if unix socket is available in current installation then use it, if not then use boot2docker/docker machine.

tjamet:
Unless I missed a commit in the PR, the logic is pretty much:
If there is no DOCKER_HOST set in the configuration nor in the environment:
- get the default socket for the current environment
- consider the default socket as a file socket
- If the path of the default socket exists, assume that is the docker socket and uses it
If there is neither docker host nor default socket available, go on with the boot2docker / docker-machine discovery

lordofthejars:
_If the path of the default socket exists, assume that is the docker socket and uses it
If there is neither docker host nor default socket available, go on with the boot2docker / docker-machine discovery_ it is implemented here, now I see that:
https://github.com/arquillian/arquillian-cube/pull/890/commits/633c013a31975baada00ae962d312c9f8ea7fb8f

Ok then there is some failures on build. Since some logic has changed regarding the defaults, the tests should be adapted as well.

tjamet:
Yes, there are the docker-machine tests to be updated, I am looking for a way to inject the default socket path in tests so that we can have deterministic tests for docker-machine regardless the environment we are running on (wether docker is available or not)

tjamet:
I have been able to test the changes on an integration on my side, and encountered a little issue in case docker for mac is installed but not up, the default socket exists but cannot reach an engine.
I am working to update the check for something more reliable and fallback to docker-machine in this case

lordofthejars:
Sure, no rush, thank you very much for your help.

tjamet:
Hi @lordofthejars I updated the PR to fallback on docker machine / boot2docker if the default docker is unavailable. What would you see as next steps?

rpau:
ping @lordofthejars

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfigurationResolver.java
```diff
@@ -196,6 +207,30 @@ private Machine getFirstMachine(Set<Machine> machines) {
...
         return config;
     }
+    
+    private Map<String, String> resolveSystemDefaultSetup(Map<String, String> config) {
+        if (!config.containsKey(CubeDockerConfiguration.DOCKER_URI)) {
+            final String defaultUri = operatingSystem.getDefaultFamily().getServerUri();
+            URI uri = URI.create(defaultUri);
+            if (Files.exists(FileSystems.getDefault().getPath(uri.getPath()))){
+
+                DockerClient client = defaultDocker.GetDefaultDockerClient(defaultUri);
```
lordofthejars: Can you change method to be Java style? (Camel case)

lordofthejars:
@rpau Last change before merge :)

docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfigurationResolver.java
tjamet: the `GetDefaultDockerClient` right?

tjamet:
thanks for your feedback @lordofthejars 
let me know if there is anything else to update :)

PR #891, opened 2017-12-01
fix: routing using xip.io

bartoszmajsak:
As `nip.io` is down let's try to run the tests using `xip.io` instead.


.travis.yml
```diff
@@ -52,7 +52,7 @@ install:
             chmod +x kubectl && sudo mv kubectl /usr/local/bin/
         client_tools="openshift-origin-client-tools-${OC_VERSION}-${COMMIT_ID}-linux-64bit" && curl -LO https://github.com/openshift/origin/releases/download/${OC_VERSION}/${client_tools}.tar.gz && \
           tar -xvzf ${client_tools}.tar.gz && sudo mv $PWD/${client_tools}/oc /usr/local/bin/ && rm -rf ${client_tools}.tar.gz
-        oc cluster up --version=${OC_VERSION} --routing-suffix='127.0.0.1.xip.io'
+        oc cluster up --version=${OC_VERSION} --routing-suffix="127.0.0.1.${OC_DOMAIN:-nip.io}"
```
lordofthejars: Cannot fins where are you setting `OC_DOMAIN` to `xip.io`

bartoszmajsak: It's set in build config in Travis https://travis-ci.org/arquillian/arquillian-cube/settings

lordofthejars: Ah ok, for this reason I couldn't find anything cool.

PR #898, opened 2017-12-12
fix: Openshift resource installer.

iocanel:
The resource installer for Openshift is currently unusable, as it doesn't override the `toImmutable()` method which returns an immutable instance of its super class.

This pull request fixes that and also adds some more logging.


lordofthejars:
One question, why this error has not detected in the functional test of Templates?

iocanel:
@lordofthejars: because its only visible if you try to pass a template parameters file.

lordofthejars:
Great, thanks for the find and the improvement.

PR #903, opened 2017-12-14
Fix #895: Creates integration between graphene and openshift

### LINKED ISSUE #895: Creates an integration between Graphene and OpenShift
##### Issue Overview

Creates an integration between Graphene and OpenShift so Graphene is configured using `route` as done in Cube Docker https://github.com/arquillian/arquillian-cube/tree/master/docker/drone

##### Expected Behaviour

Allow Graphene to be configured using the `route` definition of cluster

##### Current Behaviour

Cannot be done and you need to set the hardcode url.

hemanik:
Creates an integration between graphene and openshift, such that graphene can be configured as `route` definition of the cluster.

- Implements route resolution logic.
- Adds ftest for the same.

Fixes #895


openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/graphene/location/OpenshiftCustomizableURLResourceProvider.java
```diff
@@ -0,0 +1,97 @@
...
+
+        final String configuredUrl = grapheneConfiguration.getUrl();
+
+        if (configuredUrl != null && !configuredUrl.isEmpty()) {
+            return getRoute(new URL(configuredUrl).getHost()).get();
+        }
+        return getRoute().get();
+    }
+
+    private Optional<URL> getRoute(String routeName) {
```
lordofthejars: Instead of passing routeName, pass the URL (`new URL(configuredURL)`) then do the `getHost` thing. But if `getRoute` returns null, return the original `new URL(configuredURL)`. So in this way this method does not return an `Optional` and always returns a valid `URL` object.

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/graphene/location/OpenshiftCustomizableURLResourceProvider.java
```diff
@@ -0,0 +1,97 @@
...
+        final OpenShiftClient client = clientInstance.get();
+
+        final CubeOpenShiftConfiguration config = (CubeOpenShiftConfiguration) configurationInstance.get();
+
+        Route route = client.getClient().routes().inNamespace(config.getNamespace()).withName(routeName).get();
+
+        return route != null ? createUrlFromRoute(route) : Optional.empty();
+    }
+
+    private Optional<URL> getRoute() {
```
lordofthejars: You are doing nothing in the caller with `Optional` so I'll suggest returning directly the `URL`, then in the caller, you just need to check if it is null and if it is null just throw an exception saying that no route is defined.

lordofthejars:
Last thing before merging, can you update the documentation as well in OpenShift section?


Thank you very much :)

lordofthejars:
Thank you very much for this new awesome feature.

PR #913, opened 2017-12-19
Adds example for new JUnit Rules

### LINKED ISSUE #912: Write examples for new Arquillian JUnit Rule
Arquillian-core implemented JUnit Rule feature as a part of arquillian/arquillian-core#147.

Write few tests to run it with new Arquillian JUnit Rules

dipak-pawar:
Adds example for new JUnit arquillian Rule implemented as a part of arquillian/arquillian-core#147.

Fixes #912

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -63,7 +63,7 @@
   </issueManagement>
 
   <properties>
-    <version.arquillian_core>1.1.15.Final</version.arquillian_core>
+    <version.arquillian_core>1.2.0-SNAPSHOT</version.arquillian_core>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -63,7 +63,7 @@
   </issueManagement>
 
   <properties>
-    <version.arquillian_core>1.1.15.Final</version.arquillian_core>
+    <version.arquillian_core>1.2.0-SNAPSHOT</version.arquillian_core>
```
bartoszmajsak: Need to be changed to the released version before merging

PR #918, opened 2017-12-29
fix(#886): categorizes tests as per environment

### LINKED ISSUE #886: Create maven profiles to start docker tests or k8s/openshift tests on travis CI.
##### Issue Overview
There is no different profiles for different env. We can create parallel configuration by enabling `env in [docker, k8s/openshift]`. So just run k8s & openshift tests in parallely with travis matrix when env k8s is enabled & docker in another job.

This will reduce build time

dipak-pawar:
Categorizes tests as per environment

 -  Decoupled requirement and resolver/logic by using SPI instead of meta-annotation as it's done now.
 - moved annotations to its own module
 - skipping releasing this module
 - together with @RequireOpenshift using @Category(RequireOpenshift.class).
-  Running all unit tests with respective category tests using surefire and failsafe.

Fixes #886 

We can run all unit tests in one build job parallel with other categorized builds, once https://github.com/travis-ci/travis-ci/issues/2062 gets fixed. So we can get fast feedback if any one unit tests failed


docker/docker/src/test/java/org/arquillian/cube/docker/impl/client/BeforeStopContainerObserverTest.java
```diff
@@ -32,6 +35,7 @@
 import static org.mockito.Mockito.verify;
 
 @RunWith(MockitoJUnitRunner.class)
+@Category({RequiresDocker.class, RequiresDockerMachine.class})
```
bartoszmajsak: Do we really need these categories if we are running mockito-powered tests?

docker/docker/src/test/java/org/arquillian/cube/docker/impl/client/CubeConfigurationTest.java
```diff
@@ -25,6 +28,7 @@
 import org.hamcrest.collection.IsIterableContainingInOrder;
 import static org.junit.Assert.assertThat;
 
+@Category({RequiresDocker.class, RequiresDockerMachine.class})
```
bartoszmajsak: This looks like a unit test to me - why do we require docker category here?

pom.xml
```diff
@@ -358,6 +365,43 @@
...
+            <artifactId>maven-surefire-plugin</artifactId>
+            <version>${version.surefire}</version>
+            <executions>
+              <execution>
+                <id>default-test</id>
+                <goals>
+                  <goal>test</goal>
+                </goals>
+                <configuration>
+                  <groups>${env.type}</groups>
```
bartoszmajsak: Not really sure if that's needed. According to [the docs](http://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#groups) you should be able to use `-Dgroups` when running the build, so no need for a profile, really.

requirement/src/main/java/org/arquillian/cube/requirement/Requirements.java
```diff
@@ -26,4 +33,24 @@ public static void checkRequirement(Requires requires, Annotation context) throw
...
+
+        final Iterator<Requirement> requirements = requirementLoader.iterator();
+
+        while (requirement == null && requirements.hasNext()) {
+            Requirement req = requirements.next();
+            try {
+                req.getClass().getDeclaredMethod("check", context.annotationType());
+                requirement = req;
+            } catch (NoSuchMethodException e) {
+                // if method not found with required signature. Look for next implementation
```
bartoszmajsak: What if after iterating over all implementations found on the CP we still have `requirement` being `null`? Shall we fail in such a case?

pom.xml
```diff
@@ -65,6 +66,7 @@
   <properties>
     <version.arquillian_core>1.2.0.Final</version.arquillian_core>
     <version.junit>4.12</version.junit>
+    <version.surefire>2.19.1</version.surefire>
```
bartoszmajsak: Can't we use newer version `2.20.1`?

requirement/src/main/java/org/arquillian/cube/requirement/Requirements.java
```diff
@@ -26,4 +33,24 @@ public static void checkRequirement(Requires requires, Annotation context) throw
             }
         }
     }
+
+    private static Requirement getRequirement(Annotation context) {
```
bartoszmajsak: Maybe some better method name suggesting it's dynamically looking for SPI implementation?

requirement/src/main/java/org/arquillian/cube/requirement/Requirements.java
```diff
@@ -26,4 +33,24 @@ public static void checkRequirement(Requires requires, Annotation context) throw
...
+        Requirement requirement = null;
+
+        final ServiceLoader<Requirement> requirementLoader = ServiceLoader.load(Requirement.class);
+
+        final Iterator<Requirement> requirements = requirementLoader.iterator();
+
+        while (requirement == null && requirements.hasNext()) {
+            Requirement req = requirements.next();
+            try {
+                req.getClass().getDeclaredMethod("check", context.annotationType());
```
bartoszmajsak: I struggle to understand how this code is supposed to be working. We never really `check` if given implementation satisfies the requirement, but always just return the last passing this loop. And only then we call the actual `.check` in the caller method `checkRequirement`.  Will this always work?

requirement-spi/src/main/java/org/arquillian/cube/openshift/impl/requirement/RequiresOpenshift.java
```diff
@@ -4,11 +4,12 @@
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
+import org.arquillian.cube.spi.requirement.Requirement;
 import org.arquillian.cube.spi.requirement.Requires;
 
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ElementType.TYPE, ElementType.METHOD})
-@Requires(OpenshiftRequirement.class)
+@Requires(Requirement.class)
```
bartoszmajsak: This looks a bit odd now. WDYT @lordofthejars?

docker/docker/src/test/java/org/arquillian/cube/docker/impl/client/BeforeStopContainerObserverTest.java
dipak-pawar: Now if we decided to just run all docker related tests (unit and functional), currently there is no way, if we run tests using category, it will execute all unit tests (docker, k8s, openshift) and functional tests having `@Category(RequiresDocker.class)`, as for all unit tests we don't have any category.

So to run particular env(docker/k8s) unit tests I applied `@Category` for all unit tests

dipak-pawar: With this also you can run tests on host having docker or host having docker-machine without commenting anything if we fix how to get docker machine installed on host?

docker/docker/src/test/java/org/arquillian/cube/docker/impl/client/CubeConfigurationTest.java
dipak-pawar: same like above

docker/docker/src/test/java/org/arquillian/cube/docker/impl/client/BeforeStopContainerObserverTest.java
bartoszmajsak: I understand this reasoning, but this effectively means that we have to have a category always defined for every test or run the build without `groups` defined. I am not sure if that's intuitive.

In addition, it can be overlooked (unless we bring some check during the build to enforce having categories on each test), so if we don't have a build on our CI without `groups` defined those tests will never be run.

pom.xml
dipak-pawar: This works fine for `mvn clean test`. But when you try to use same configuration for `mvn clean install`. It won't work. It fails saying ` Cannot find default setter in class org.apache.maven.plugins.javadoc.options.Group`

bartoszmajsak: That's a very odd type (`javadoc`?) :) Does it also show up when you use latest Surefire?

requirement/src/main/java/org/arquillian/cube/requirement/Requirements.java
dipak-pawar: yes we should fail in such case. Good catch

pom.xml
dipak-pawar: yes, it's reproducing with latest `surefire`. ok I will try to overwrite it with `jboss-parent`

bartoszmajsak: Can you figure out if that's a known issue or limitation? Since we run `install` maybe it's the failsafe side which is causing the problem?

requirement/src/main/java/org/arquillian/cube/requirement/Requirements.java
dipak-pawar: @bartoszmajsak , We are looking for requirement with respective method definition `check(context.annotationType)`. If it haven't found it it will throw exception by saying `NoSuchMethodException` which we are catching & iterating over next. Here we are returning first requirement with correct method(`check(RequiresDocker)|check(RequiresKubernetes)`) on classpath, we are not returning last passing value. This should always work

bartoszmajsak: I this given case, yes. But what if you have two SPIs with the same method signature (unlikely, we are not preventing it from happening)?

bartoszmajsak: And I still believe we are returning the last matching one, not the first one.

bartoszmajsak: > Here we are returning first requirement with correct method

Right, but the main concern I explained above still remains.

pom.xml
dipak-pawar: no it's not the case. [maven-javadoc-plugin](https://maven.apache.org/plugins/maven-javadoc-plugin/examples/group-configuration.html) has `groups` as configuration property which is for groping packages. So it has conflict with this.

We have workaround for this as follows:
```xml
<properties>
    <surefire.groups/>
  </properties>
<plugin>
 <artifactId>maven-surefire-plugin</artifactId>
    <configuration>
      <groups>${surefire.groups}</groups>
    </configuration>
</plugin>
```

So with this we run run our tests as `mvn clean install -Dsurefire.groups=RequiresDocker`

requirement/src/main/java/org/arquillian/cube/requirement/Requirements.java
dipak-pawar: yeah but this is the limitation of spi. May be we can add weight to take precedence of one over another wdyt?

bartoszmajsak: That's not a limitation of SPI really, more this code to be honest. But on the other we don't need it yet. Worth documenting it though.

requirement-spi/src/main/java/org/arquillian/cube/openshift/impl/requirement/RequiresOpenshift.java
lordofthejars: Yeah really odd.

dipak-pawar: Can you suggest me something using which we can make it clean

kubernetes/ftest-kubernetes/src/test/java/HelloWorldIT.java
```diff
@@ -1,3 +1,4 @@
+import com.sun.org.apache.regexp.internal.RE;
```
lordofthejars: What is this import?

requirement-spi/src/main/java/org/arquillian/cube/openshift/impl/requirement/RequiresOpenshift.java
lordofthejars: I am really bad given names hehehe maybe `constraint`?

bartoszmajsak: I don't think it's even about the name, but the general construct. For now, maybe renaming is enough, but we might want to revisit our approach with annotations. Something tells me that we might need to improve the design

bartoszmajsak:
I'm fine with the changes.

@lordofthejars if it makes sense for you, let's create an issue to address `@Requires(Constraint.class)` construct - I believe we can get rid off it.

PR #926, opened 2018-01-03
Improve could not resolve route message

spolti:
Add a hint about the available routes when you specify a invalid route name, like the example below:

```
    @RouteURL("kie-app")
    private URL routeURL;
```

With this change a message similar to this will be printed within the stacktrace:

> Caused by: java.lang.IllegalArgumentException: Could not resolve route: kie-app. Available routes: [kie-app-kieserver, secure-kie-app-kieserver]

It could be helpful when you are unsure about the route name, and instead to configure the arquillian to do not remove the test project to check the route names,
it is easier to have it on the stacktrace.


dipak-pawar:
Thank you @spolti. Looks good to me.

PR #927, opened 2018-01-04
chore(#924): Abstract common logic to Kubernetes Assistant and extend it to OpenShift Assistant

### LINKED ISSUE #924: Integrate common parts of KubernetesAssistant and OpenShiftAssistant
##### Issue Overview

Integrate common parts of `KubernetesAssistant` and `OpenShiftAssistant`

##### Expected Behaviour

Have common bits of both assistants in the same class and not duplicated.

Now it comes to my mind two different approaches, one is creating an abstract class which both classes extends, or put all common parts on `KubernetesAssistant` and make `OpenShiftAssistant` extending `KubernetesAssistant`. I prefer the later one.

hemanik:

Abstract the common logic from the Kubernetes and OpenShift Assistants to Kubernetes Assistant and extend the same to OpenShift Assistant.

- Removes duplicate code logic from OpenShift Assistant.
- Fixes `canProvide` method for provider classes.

Fixes #924

### DIFF
--- a/kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/KubernetesAssistant.java
+++ b/kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/KubernetesAssistant.java
@@ -13,7 +13,6 @@ import io.fabric8.kubernetes.clnt.v3_1.KubernetesClient;
 import io.fabric8.kubernetes.clnt.v3_1.KubernetesClientException;
 import io.fabric8.kubernetes.clnt.v3_1.dsl.NamespaceListVisitFromServerGetDeleteRecreateWaitApplicable;
 import io.fabric8.kubernetes.clnt.v3_1.internal.readiness.Readiness;
-import io.fabric8.openshift.clnt.v3_1.OpenShiftClient;
 import io.github.lukehutch.fastclasspathscanner.FastClasspathScanner;
 import io.github.lukehutch.fastclasspathscanner.matchprocessor.FileMatchProcessor;
 import org.arquillian.cube.kubernetes.impl.portforward.PortForwarder;
@@ -58,16 +57,10 @@ public class KubernetesAssistant {
     private KubernetesAssistantDefaultResourceLocator kubernetesAssistantDefaultResourcesLocator;
     private Map<String, List<HasMetadata>> created = new LinkedHashMap<>();
 
-    KubernetesAssistant(KubernetesClient client, String namespace) {
+    public KubernetesAssistant(KubernetesClient client, String namespace) {
         this.client = client;
         this.namespace = namespace;
         this.kubernetesAssistantDefaultResourcesLocator = new KubernetesAssistantDefaultResourceLocator();
-    }
-
-    public KubernetesAssistant(io.fabric8.openshift.clnt.v3_1.OpenShiftClient openShiftClient, String namespace) {
-        this.client = openShiftClient;
-        this.client.adapt(OpenShiftClient.class);
-        this.namespace = namespace;
     }
 
     /**
@@ -99,7 +92,7 @@ public class KubernetesAssistant {
         if (defaultFileOptional.isPresent()) {
             deployApplication(applicationName, defaultFileOptional.get());
         } else {
-            log.warning("No default Kubernetes or OpenShift resources found at default locations.");
+            log.warning("No default Kubernetes resources found at default locations.");
         }
     }
 
@@ -246,7 +239,7 @@ public class KubernetesAssistant {
         }
     }
 
-    public List<? extends HasMetadata> deploy(String name, InputStream element) throws IOException {
+    protected List<? extends HasMetadata> deploy(String name, InputStream element) throws IOException {
         NamespaceListVisitFromServerGetDeleteRecreateWaitApplicable<HasMetadata, Boolean> declarations = client.load(element);
         List<HasMetadata> entities = declarations.createOrReplace();
 
@@ -467,7 +460,7 @@ public class KubernetesAssistant {
             });
     }
 
-    public List<Pod> getPods(String label) {
+    protected List<Pod> getPods(String label) {
         return this.client
             .pods()
             .inNamespace(this.namespace)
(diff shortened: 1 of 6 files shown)


hemanik:
The problem is if I execute the complete HelloWorldOpenShiftAssistantTest test class, the tests fail with the error message as follows:
```
java.lang.RuntimeException: Could not set value on field org.arquillian.cube.openshift.impl.client.OpenShiftAssistant org.arquillian.cube.openshift.standalone.HelloWorldOpenShiftAssistantTest.openShiftAssistant using org.arquillian.cube.kubernetes.impl.KubernetesAssistant@1dab9dd6
```
But when each test is executed separately, it works fine.

dipak-pawar:
@hemanik I have a fix for issue. Can you give me access to push your repo?

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/KubernetesAssistant.java
```diff
@@ -48,21 +49,27 @@
...
 
-    public KubernetesAssistant(io.fabric8.kubernetes.clnt.v3_1.KubernetesClient kubernetesClient, String namespace) {
-        this.client = kubernetesClient;
+    KubernetesAssistant(KubernetesClient client, String namespace) {
+        this.client = client;
         this.namespace = namespace;
         this.kubernetesAssistantDefaultResourcesLocator = new KubernetesAssistantDefaultResourceLocator();
     }
 
+    public KubernetesAssistant(io.fabric8.openshift.clnt.v3_1.OpenShiftClient openShiftClient, String namespace) {
```
dipak-pawar: I believe we don't need this constructor.

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/KubernetesAssistant.java
```diff
@@ -219,7 +226,13 @@ public void deployAll(String applicationName, Path directory) throws IOException
...
     }
 
-    private void deploy(InputStream inputStream) throws IOException {
+    /**
+     * Deploys application reading resources from specified InputStream
+     *
+     * @param inputStream  where resources are read
+     * @throws IOException
+     */
+    public void deploy(InputStream inputStream) throws IOException {
```
dipak-pawar: Do we really need to make it public?

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/KubernetesAssistant.java
```diff
@@ -233,7 +239,7 @@ private void deploy(InputStream inputStream) throws IOException {
         }
     }
 
-    private List<? extends HasMetadata> deploy(String name, InputStream element) throws IOException {
+    public List<? extends HasMetadata> deploy(String name, InputStream element) throws IOException {
```
dipak-pawar: I am not sure Is it really necessary to make it public for end user to use it

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/KubernetesAssistant.java
```diff
@@ -455,7 +460,7 @@ public void scale(final int replicas) {
             });
     }
 
-    private List<Pod> getPods(String label) {
+    public List<Pod> getPods(String label) {
```
dipak-pawar: It can be `protected`

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/OpenShiftRouteLocator.java
```diff
@@ -52,7 +52,7 @@ private CubeOpenShiftConfiguration getCubeOpenShiftConfiguration() {
         return config;
     }
 
-    private URL createUrlFromRoute(Route route) {
+    static URL createUrlFromRoute(Route route) {
```
dipak-pawar: I am not sure do we really need to make it static?

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/enricher/external/OpenShiftAssistantResourceProvider.java
```diff
@@ -14,15 +14,15 @@
...
+        return OpenShiftAssistant.class.getName().equals(type.getName());
     }
 
     @Override
     public Object lookup(ArquillianResource arquillianResource, Annotation... annotations) {
         OpenShiftAssistant openShiftAssistant = this.openShiftAssistantInstance.get();
 
         if (openShiftAssistant == null) {
-            throw new IllegalStateException("Unable to inject DockerClient into test.");
+            throw new IllegalStateException("Unable to inject OpenshiftAssistant into test.");
```
dipak-pawar: Are we injecting here or doing lookup?

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/KubernetesAssistant.java
hemanik: I think this gives just another means of deploying application, one example where I saw this could be used was for launcher-backend. So, I feel it makes sense to having it as public.

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/enricher/external/OpenShiftAssistantResourceProvider.java
hemanik: It is injected above.

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/OpenShiftRouteLocator.java
hemanik: I need to access it in OpenShiftAssistant, if I don't make it static, I'll have to create an instance of `OpenShiftRouteLocator` class.

PR #929, opened 2018-01-04
Configure Arquillian Naming Service after process templates

spolti:
This commit fixes the following issue:

If you define the following properties on arquillian.xml everything works as expected:

```
<property name="definitionsFile">src/test/resources/testrunner-pod.json</property>
<property name="proxiedContainerPorts">testrunner:9999,testrunner:9990,testrunner:8080</property>
```

The issue happens when there is no definitionsFile, the BuildablePodCube.java, the class where the routes are added in the naming service is not called.

With this fixes, all routes, even if there is multiple templates, will be properly configured after all resources are in place.


lordofthejars:
One comment, now I see that you've moved from BuildablePod logic to template logic, doesn't this mean that this logic is only executed in case of templates where before was called always?

spolti:
Do you think we should keep it on both places?

lordofthejars:
This is what I don't know, but it seems to me (I am not deeply familiar with this piece) that we are moving the functionality instead of fixing in one place where it is not executed.

spolti:
That's right, the problem is if there is no definitionsFile to be processed the Observer on TemplateContainerStarter.waitForDeployments (IMHO the routesAdd could be here instead BuildablePodCube) is not triggered, not sure why, I am still trying to figure it out, do you have any idea?

lordofthejars:
no idea in this part

spolti:
@lordofthejars  I updated the PR, it now will wait for all resources and create the routes properly with or not the definitionsFile, or if there is only the definitionsFile.

lordofthejars:
Seems good to me, @iocanel or  @jwendell can you take a quick review before I merge it?

Thank you very much.

jwendell:
is there an test for this already? if we have, it means test passed and I'm ok with it. Otherwise it would be good to add one that showcases the expected behavior.

spolti:
I'll rebase it and try to add some tests to it.
Thanks and sorry for the delay.

lordofthejars:
Sure, no worries :)

spolti:
Hey @lordofthejars / @jwendell  can you please take a look?

lordofthejars:
I have run the built twice but it is failing in all cases.

Issue #931, opened 2018-01-08
Use comma to separate multiple values in arquillian.xml


spisiakm:
##### Issue Overview

We ran into this issue when using Arquillian Cube to test Vert.x ConfigMap booster on OSO Pro. We're adding additional kubernetes resources through `env.dependencies` property in arquillian.xml. When a path to additional kubernetes/openshift resources contains a whitespace, it gets separated at this place, which results in an error - *FileNotFoundException*. Example: run a booster from a directory that contains a space in the name, for example `/home/Vert.x boosters/....` 

This could be resolved by using comma (and arbitrarily long string of whitespaces after that) to separate multiple values rather than just whitespaces.

##### Expected Behaviour

Additional kubernetes resources are located correctly even if the absolute path to them contains a whitespace.

##### Current Behaviour

If the absolute path contains a whitespace, it is split at this place, which results in an error.

##### Steps To Reproduce
1. `git clone git@github.com:openshiftio-vertx-boosters/vertx-configmap-booster-redhat.git` in any directory that contains a whitespace in the name
2. `mvn clean verify -Popenshift,openshift-it` from this directory. Note that you must be logged to an existing OpenShift instance.

lordofthejars:
Thank you very much.

Ladicek:
Frankly, I don't think this is a proper fix. I can put a comma into a file path and will be screwed again. IMHO, a proper fix would be either adding some form of escaping / quotation, or using XML elements for individual elements. I even think it would be possible to implement both options in a backwards-compatible way.

lordofthejars:
I agree that you can still put a comma in a path, but it is really strange to have a path with commas. Also, having whitespace as separator even escaping them is a good idea from the point of view of the user since checking the attribute you don't clearly know what is a new path and what's not.

Saying that of course, the best solution would be as you mention having XML elements there but then I think that what we are doing is a reparsing of an XML document. Something to discuss if this can be provided by core.

PR #932, opened 2018-01-08
ENVIRONMENT_DEPENDENCIES are now separated by comma

spisiakm:

https://github.com/arquillian/arquillian-cube/issues/931

- use a comma (and arbitrarily long string of whitespace chars) instead of just whitespaces to separate multiple values specified in `arquillian.xml` configuration file

### DIFF
--- a/kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/DefaultConfiguration.java
+++ b/kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/DefaultConfiguration.java
@@ -151,7 +151,7 @@ public class DefaultConfiguration implements Configuration {
                 .withWaitTimeout(getLongProperty(WAIT_TIMEOUT, map, DEFAULT_WAIT_TIMEOUT))
                 .withWaitPollInterval(getLongProperty(WAIT_POLL_INTERVAL, map, DEFAULT_WAIT_POLL_INTERVAL))
                 .withWaitForServiceList(
-                    Strings.splitAndTrimAsList(getStringProperty(WAIT_FOR_SERVICE_LIST, map, ""), "\\s+"))
+                    Strings.splitAndTrimAsList(getStringProperty(WAIT_FOR_SERVICE_LIST, map, ""), ",\\s*"))
                 .withAnsiLoggerEnabled(getBooleanProperty(ANSI_LOGGER_ENABLED, map, true))
                 .withKubernetesDomain(getStringProperty(DOMAIN, KUBERNETES_DOMAIN, map, null))
                 .withDockerRegistry(getDockerRegistry(map))
--- a/kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/DefaultConfiguration.java
+++ b/kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/DefaultConfiguration.java
@@ -132,7 +132,7 @@ public class DefaultConfiguration implements Configuration {
                     asUrlOrResource(getStringProperty(ENVIRONMENT_TEARDOWN_SCRIPT_URL, map, null)))
                 .withEnvironmentConfigUrl(getKubernetesConfigurationUrl(map))
                 .withEnvironmentDependencies(
-                    asURL(Strings.splitAndTrimAsList(getStringProperty(ENVIRONMENT_DEPENDENCIES, map, ""), ",\\s*")))
+                    asURL(Strings.splitAndTrimAsList(getStringProperty(ENVIRONMENT_DEPENDENCIES, map, ""), "\\s*,\\s*")))
                 .withNamespaceLazyCreateEnabled(
                     getBooleanProperty(NAMESPACE_LAZY_CREATE_ENABLED, map, DEFAULT_NAMESPACE_LAZY_CREATE_ENABLED))
                 .withNamespaceCleanupEnabled(getBooleanProperty(NAMESPACE_CLEANUP_ENABLED, map, true))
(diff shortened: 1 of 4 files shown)


spisiakm:
I also noticed that multiple values in `WAIT_FOR_SERVICE_LIST` property are also separated by just whitespaces. That probably needs to be changed as well.

lordofthejars:
Good, thank you very much, can you provide the fix for `WAIT_FOR_SERVICE_LIST ` and update the docs?

Thank you very much.

cescoffier:
👍

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/DefaultConfiguration.java
```diff
@@ -132,7 +132,7 @@ public static DefaultConfiguration fromMap(Map<String, String> map) {
                     asUrlOrResource(getStringProperty(ENVIRONMENT_TEARDOWN_SCRIPT_URL, map, null)))
                 .withEnvironmentConfigUrl(getKubernetesConfigurationUrl(map))
                 .withEnvironmentDependencies(
-                    asURL(Strings.splitAndTrimAsList(getStringProperty(ENVIRONMENT_DEPENDENCIES, map, ""), "\\s+")))
+                    asURL(Strings.splitAndTrimAsList(getStringProperty(ENVIRONMENT_DEPENDENCIES, map, ""), ",\\s*")))
```
dipak-pawar: How about `\\s*,\\s*` regex?

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/DefaultConfiguration.java
```diff
@@ -151,7 +151,7 @@ public static DefaultConfiguration fromMap(Map<String, String> map) {
                 .withWaitTimeout(getLongProperty(WAIT_TIMEOUT, map, DEFAULT_WAIT_TIMEOUT))
                 .withWaitPollInterval(getLongProperty(WAIT_POLL_INTERVAL, map, DEFAULT_WAIT_POLL_INTERVAL))
                 .withWaitForServiceList(
-                    Strings.splitAndTrimAsList(getStringProperty(WAIT_FOR_SERVICE_LIST, map, ""), "\\s+"))
+                    Strings.splitAndTrimAsList(getStringProperty(WAIT_FOR_SERVICE_LIST, map, ""), ",\\s*"))
```
dipak-pawar: How about `\\s*,\\s*` regex?

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/CubeOpenShiftConfiguration.java
```diff
@@ -164,7 +164,7 @@ public static CubeOpenShiftConfiguration fromMap(Map<String, String> map) {
                     asUrlOrResource(getStringProperty(ENVIRONMENT_TEARDOWN_SCRIPT_URL, map, null)))
                 .withEnvironmentConfigUrl(getKubernetesConfigurationUrl(map))
                 .withEnvironmentDependencies(
-                    asURL(Strings.splitAndTrimAsList(getStringProperty(ENVIRONMENT_DEPENDENCIES, map, ""), "\\s+")))
+                    asURL(Strings.splitAndTrimAsList(getStringProperty(ENVIRONMENT_DEPENDENCIES, map, ""), ",\\s*")))
```
dipak-pawar: same as above

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/CubeOpenShiftConfiguration.java
```diff
@@ -183,7 +183,7 @@ public static CubeOpenShiftConfiguration fromMap(Map<String, String> map) {
                 .withWaitTimeout(getLongProperty(WAIT_TIMEOUT, map, DEFAULT_WAIT_TIMEOUT))
                 .withWaitPollInterval(getLongProperty(WAIT_POLL_INTERVAL, map, DEFAULT_WAIT_POLL_INTERVAL))
                 .withWaitForServiceList(
-                    Strings.splitAndTrimAsList(getStringProperty(WAIT_FOR_SERVICE_LIST, map, ""), "\\s+"))
+                    Strings.splitAndTrimAsList(getStringProperty(WAIT_FOR_SERVICE_LIST, map, ""), ",\\s*"))
```
dipak-pawar: same as above

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/DefaultConfiguration.java
spisiakm: I don't mind, let me update it :)

lordofthejars: I am not sure about any of both expressions, I think that the comma part should be optional, so the property can be either `mykubernetes.json` or `mykubernetes.json, mysecondkubernetes.json`

dipak-pawar: With `\\s*,\\s*` regex following cases will be covered:
* mykubernetes.json 
* mykubernetes.json, mysecondkubernetes.json
* mykubernetes.json , mysecondkubernetes.json

spisiakm: Well, you're using `String.split` java method for separating multiple values by regex, so if there's just one single value, nothing happens and this value will be returned as is.

lordofthejars:
Now looks good to me, if @dipak-pawar agrees, then tomorrow I'll release Cube 2.0.0

lordofthejars:
Also @spisiakm can you fix the build? There are some test failing because of the change (`org.arquillian.cube.kubernetes.reporter.TakeKubernetesResourcesInformationTest`) 

Thanks.

spisiakm:
I'll fix that tomorrow morning.

dipak-pawar:
Thank you @spisiakm. Lgtm

PR #938, opened 2018-01-11
feat(#934): introduces template dsl for openshift assistant.

### LINKED ISSUE #934: Create Template DSL for OpenShiftAssistant
##### Issue Overview

Create Template DSL for `OpenShiftAssistant` so you can apply templates programmatically.

This DSL might look like: 

`openshiftAssistant.usingTemplate("path/to/template").parameter("param1", "value1").parameter("param2", "value2").deploy()`

##### Expected Behaviour

Deploy templates programmatically.

hemanik:
Creates Template DSL for OpenShiftAssistant so you can apply templates programmatically.

This DSL looks like:

`openshiftAssistant.usingTemplate("url/to/template").parameter("param1", "value1").parameter("param2", "value2").deploy()`

- Template DSL for OpenShift Assistant.
- ftest to showcase template DSL for OpenShift Assistant.

Fixes #934

### DIFF
--- a/openshift/ftest-openshift-assistant-template/pom.xml
+++ b/openshift/ftest-openshift-assistant-template/pom.xml
@@ -39,5 +39,17 @@
     </dependency>
   </dependencies>
 
+  <build>
+    <plugins>
+      <!-- This artifact shouldn't be deployed to maven repository -->
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-deploy-plugin</artifactId>
+        <configuration>
+          <skip>true</skip>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
 
 </project>
--- a/openshift/ftest-openshift-assistant-template/pom.xml
+++ b/openshift/ftest-openshift-assistant-template/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <artifactId>arquillian-cube-openshift-parent</artifactId>
     <groupId>org.arquillian.cube</groupId>
-    <version>1.12.1-SNAPSHOT</version>
+    <version>1.13.1-SNAPSHOT</version>
   </parent>
   <modelVersion>4.0.0</modelVersion>
 
--- a/openshift/ftest-openshift-assistant-template/pom.xml
+++ b/openshift/ftest-openshift-assistant-template/pom.xml
@@ -1,0 +1,43 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <artifactId>arquillian-cube-openshift-parent</artifactId>
+    <groupId>org.arquillian.cube</groupId>
+    <version>1.12.1-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>arquillian-cube-openshift-assistant-template-ftest</artifactId>
+
+  <properties>
+    <maven.compiler.target>1.8</maven.compiler.target>
+    <maven.compiler.source>1.8</maven.compiler.source>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.arquillian.cube</groupId>
+      <artifactId>arquillian-cube-requirement</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.arquillian.cube</groupId>
+      <artifactId>arquillian-cube-openshift</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.jboss.arquillian.junit</groupId>
+      <artifactId>arquillian-junit-standalone</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+
+
+</project>
--- a/openshift/ftest-openshift-assistant-template/src/test/java/org/arquillian/cube/openshift/standalone/HelloOpenShiftAssistantTemplateTest.java
+++ b/openshift/ftest-openshift-assistant-template/src/test/java/org/arquillian/cube/openshift/standalone/HelloOpenShiftAssistantTemplateTest.java
@@ -32,7 +32,7 @@ public class HelloOpenShiftAssistantTemplateTest {
     public void should_apply_template_programmatically() throws IOException {
 
         openShiftAssistant
-                .usingTemplate(String.valueOf(getClass().getClassLoader().getResource("hello-template.yaml")))
+                .usingTemplate(getClass().getClassLoader().getResource("hello-template.yaml"))
                 .parameter("RESPONSE", "Hello from Arquillian Template")
             .deploy();
 
--- a/openshift/ftest-openshift-assistant-template/src/test/java/org/arquillian/cube/openshift/standalone/HelloOpenShiftAssistantTemplateTest.java
+++ b/openshift/ftest-openshift-assistant-template/src/test/java/org/arquillian/cube/openshift/standalone/HelloOpenShiftAssistantTemplateTest.java
@@ -1,0 +1,51 @@
+package org.arquillian.cube.openshift.standalone;
+
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.Response;
+import org.arquillian.cube.openshift.impl.client.OpenShiftAssistant;
+import org.arquillian.cube.openshift.impl.requirement.RequiresOpenshift;
+import org.arquillian.cube.requirement.ArquillianConditionalRunner;
+import org.jboss.arquillian.test.api.ArquillianResource;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import java.io.IOException;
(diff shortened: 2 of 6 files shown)


openshift/ftest-openshift-assistant-template/src/test/java/org/arquillian/cube/openshift/standalone/HelloOpenShiftAssistantTemplateTest.java
```diff
@@ -0,0 +1,51 @@
...
+    @Test
+    public void should_inject_openshift_assistant() {
+        assertThat(openShiftAssistant).isNotNull();
+    }
+
+    @Test
+    public void should_apply_template_programmatically() throws IOException {
+
+        openShiftAssistant
+                .usingTemplate("https://gist.githubusercontent.com/lordofthejars/8781cacd4000a516695ad6c55b5815b3/raw/5151aeef0f5dd8823e2c581c3b7452f04a76af59/hello-template.yaml")
```
bartoszmajsak: This way we won't be able to test it when GH is down (which was the case few minutes ago). Does it matter we want it from some remote location?

openshift/ftest-openshift-assistant-template/pom.xml
```diff
@@ -0,0 +1,43 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <artifactId>arquillian-cube-openshift-parent</artifactId>
+    <groupId>org.arquillian.cube</groupId>
+    <version>1.12.1-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>arquillian-cube-openshift-assistant-template-ftest</artifactId>
```
bartoszmajsak: Makes we wonder if we really need a new module for that. Shouldn't we just add it to `ftest-openshift-assistant`? WDYT @lordofthejars

bartoszmajsak: Also, if we keep it this way we shouldn't release it to maven central - there is probably no point for anyone to consume `ftest` jars.

hemanik: > Shouldn't we just add it to ftest-openshift-assistant? 

I tried this initially, but that module already contains an openshift.json file, which is read by default from the classpath, though it gets overridden by the template we apply. But the output is kind of confusing, it first logs something like, `Reading from template file, no external resources found` etc., which I found quite misleading at the start.  

@bartoszmajsak, @lordofthejars - would like to know the best and most optimized way of handling this.

lordofthejars: Yes Hemani, you are right, leave it as a separate ftest. Basically the `openshift-assistant` ftest is testing that the default cube parameters also work with `OpenShiftAssistant` so yeah.

Regarding ftest modules, I think that for now there are some of them that are released to Maven central, so let's fix this problem as well

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/OpenShiftAssistant.java
```diff
@@ -30,6 +36,10 @@
 
     private OpenShiftAssistantDefaultResourcesLocator openShiftAssistantDefaultResourcesLocator;
 
+    private List<ParameterValue> pvs = new ArrayList<>();
```
lordofthejars: Can you rename to `parameterValueList`? Because here it is clear but then when I read the code I need to think what it means `pvs` )

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/OpenShiftAssistant.java
```diff
@@ -160,4 +170,37 @@ public DeploymentConfig deploymentConfig() {
             .withName(this.applicationName)
             .get();
     }
+
+    public OpenShiftAssistant usingTemplate(String templateURL) {
+        this.templateURL = templateURL;
+        return this;
+    }
+
+    public OpenShiftAssistant parameter(String param, String value) {
```
lordofthejars: Question? What's happen if user sets same parameter twice? Yes I know that both are inserted in the list, is this something ok, or we should replace the previous parameter with the new one like happens with HashMap?

@bartoszmajsak

openshift/ftest-openshift-assistant-template/pom.xml
bartoszmajsak: Can't we use another filename though for this test?

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/OpenShiftAssistant.java
```diff
@@ -160,4 +170,37 @@ public DeploymentConfig deploymentConfig() {
             .withName(this.applicationName)
             .get();
     }
+
+    public OpenShiftAssistant usingTemplate(String templateURL) {
```
lordofthejars: Instead of returning `OpenShiftAssistant` I would return a class called `OpenShiftAssistantTemplate` which receives as constructor the `client`, then this class would contain the method to add parameters and the `deploy` method.

In this way you are avoiding that someone mistakenly calls `deploy` method without setting any template.

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/OpenShiftAssistant.java
bartoszmajsak: Behavior-wise it's the same I guess? Last one wins. However I would use a map and only convert to `ParameterValue`s when really used.

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/OpenShiftAssistant.java
bartoszmajsak: why `List`? `parameterValues` won't work?

openshift/ftest-openshift-assistant-template/pom.xml
hemanik: This still doesn't help, because classpath is searched for configuration file first. And the output is as follows:

``` 
Applying kubernetes configuration from: file:/home/hemani/trello/arquillian-cube/openshift/ftest-openshift-assistant/target/test-classes/hello-openshift.json
Processing template. No parameters file has been specified, processing without external parameters!
```

bartoszmajsak: @lordofthejars Is this a bug then?

lordofthejars: No it is not a bug, by default arquillian "kube" loads some defaults- And this is what I wanted to test in that ftest, that defaults are loaded independently of openshiftassistant.

openshift/ftest-openshift-assistant-template/pom.xml
```diff
@@ -0,0 +1,43 @@
...
+      <artifactId>arquillian-junit-standalone</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+
```
bartoszmajsak: Can we skip this artifact from being released like we do with some of the other modules? https://github.com/arquillian/arquillian-cube/blob/master/openshift/ftest-openshift-graphene/pom.xml#L73-L83

@lordofthejars if you agree to clean it up for all the other ftest modules let's create an issue for it too

lordofthejars: https://github.com/arquillian/arquillian-cube/issues/951

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/OpenShiftAssistant.java
```diff
@@ -160,4 +160,15 @@ public DeploymentConfig deploymentConfig() {
...
             .get();
     }
+
+    /**
+     * Gets template URL used for deploying application.
+     *
+     * @param templateURL path to the template
+     * @return OpenShiftAssistantTemplate object.
+     */
+    public OpenShiftAssistantTemplate usingTemplate(String templateURL) {
```
bartoszmajsak: How about having a method which accepts `URL` next to String too?

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/OpenShiftAssistantTemplate.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class OpenShiftAssistantTemplate {
+
+    private final KubernetesClient client;
+
+    private String templateURL;
```
bartoszmajsak: If we call it `templateURL` why is it of a `String` type?

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/OpenShiftAssistantTemplate.java
```diff
@@ -0,0 +1,72 @@
...
+import java.io.InputStream;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class OpenShiftAssistantTemplate {
+
+    private final KubernetesClient client;
```
lordofthejars: Can we change this to `OpenShiftClient`, in `OpenShiftAssistant` it uses the openshift client so you can pass it directly and you save later (line 54) the `adapt` call

PR #941, opened 2018-01-12
Allow parallelization of networks

ivankelly:
If you're using networks and you want to parallelize your containers,
all the containers will end up on the same set of networks, which
could interfere with your test.

Even worse, as docker allows multiple networks with the same name, if
there's a problem with cleanup on a previous run, you may end up with
containers in different networks when you want them to be in the same
network.

This patch solves these problems by allowing the star pattern for
networks. It works the same as the container star pattern, and
networkMode and networks are updated within the container configuration.


lordofthejars:
https://github.com/arquillian/arquillian-cube/issues/942

lordofthejars:
Thank you very much for your contribution. We really appreciate this and improves Cube usage. Just one final thing before merging, can you update the docs as well?

ivankelly:
Sure, will do

ivankelly:
Updated. Do you push nightly snapshots, so I can use this without manually installing myself until there's a release.

lordofthejars:
No, we do not have nightly releases, can you wait until Monday/Tuesday or you need urgently?

ivankelly:
No urgent need. Monday or Tuesday is fine. Do you manually push snapshots, or is there a release next week?

bartoszmajsak:
@lordofthejars we can publish in bintray, right?

PR #943, opened 2018-01-12
Add doc for alias

ivankelly:
This is a very useful feature when using parallelized cubes with
networks. Only spotted it by digging in code.


lordofthejars:
Thank you so much.

PR #944, opened 2018-01-13
Allow docker configuration to be loaded from the classpath

ivankelly:
Allow users to specify the docker configuration file as a file on the
classpath, rather that on the filesystem. This means they won't have
to worry about absolute and relative paths, which can change in maven
depending on which flags are specified (forkCount for example).

It also allows configurations to be bundled into a jar and shared
between modules.

### DIFF
--- a/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfiguration.java
+++ b/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfiguration.java
@@ -1,6 +1,7 @@
 package org.arquillian.cube.docker.impl.client;
 
 import java.io.File;
+import java.io.InputStream;
 import java.io.IOException;
 import java.net.URI;
 import java.util.ArrayList;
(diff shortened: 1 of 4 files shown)


ivankelly:
Will add docs on monday, if the change is agreed to be good.

lordofthejars:
Looks good, I am just wondering one thing, but t is something that it is on my mind, should we add a new property as you do here or we could add a prefix called `classpath:` on current property so if it starts with `classpath:` means that the resource is coming from classpath. Something like it happens with Spring.

ivankelly:
I did think of that when I was thinking of not modifying arquillian at all for this (and update the URL handlers). It's a bit convoluted though. The property is dockerContainerFile, but you're not actually specifying a file, you're specifying a URL that may point to number of things. 

A separate property is simple, and keeps it clear where the content is coming from.

lordofthejars:
then let's update docs, and I merge it :)

thank you so much.

ivankelly:
@lordofthejars pushed docs. Any hint on when I could expect a published snapshot?

lordofthejars:
today you can expect a final release

ivankelly:
Excellent. Thank you

lordofthejars:
https://github.com/arquillian/arquillian-cube/issues/947

lordofthejars:
Released Cube 1.13.0

ivankelly:
Thanks alex

lordofthejars:
Thank you for using it.

PR #945, opened 2018-01-15
feat(#928): simplifies dependencies for Standalone Mode

### LINKED ISSUE #928: Simplify dependencies for standalone mode
##### Issue Overview

All the projects I have seen that are using Arquillian Cube, they are using standalone mode. This means that they always have to add two dependencies the core one and the cube (docker|restassured|kubernetes|openshift) ones.

##### Expected Behaviour

Create a project that contains both dependencies (core-standalone and cube(docker|restassured|kubernetes|openshift)) so from the UX point of view, only one jar is required.

Also, this simplifies the adoption since user don't need to understand anything about the internals of Arquillian (what does it means standalone) or use the wrong approach (they use the container mode but what they really meant is the standalone one).

##### Current Behaviour

Two dependencies are required.

hemanik:
Packages set of commonly used dependencies together in a jar module to be used as a single entity for improving user adoption.

- [x] Adds missing names to existing modules.
- [x] Adds `starter` modules for OpenShift, Kubernetes, and Docker Extensions.
- [x] Update functional tests to contain simplified dependencies.

Fixes #928

### DIFF
--- a/arquillian-cube-docker-starter/pom.xml
+++ b/arquillian-cube-docker-starter/pom.xml
@@ -1,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <groupId>org.arquillian.cube</groupId>
+    <artifactId>arquillian-cube-parent</artifactId>
+    <version>1.12.1-SNAPSHOT</version>
+  </parent>
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>arquillian-cube-docker-starter</artifactId>
+  <packaging>jar</packaging>
+
+  <name>Arquillian Cube Docker Starter</name>
+  <description>Optimized dependencyManagement for the Arquillian Cube Docker Project</description>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.arquillian.cube</groupId>
+      <artifactId>arquillian-cube-docker</artifactId>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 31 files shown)


arquillian-cube-docker-starter/pom.xml
```diff
@@ -0,0 +1,32 @@
...
+  <packaging>jar</packaging>
+
+  <name>Arquillian Cube Docker Starter</name>
+  <description>Optimized dependencyManagement for the Arquillian Cube Docker Project</description>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.arquillian.cube</groupId>
+      <artifactId>arquillian-cube-docker</artifactId>
+      <scope>test</scope>
```
dipak-pawar: Can you try with removing `<scope>test</scope>`?

pom.xml
```diff
@@ -184,6 +187,24 @@
         <version>${project.version}</version>
       </dependency>
 
+      <dependency>
+        <groupId>org.arquillian.cube</groupId>
+        <artifactId>arquillian-cube-openshift-starter</artifactId>
+        <version>${project.version}</version>
```
dipak-pawar: why we don't have `<scope>test</scope>`? In case if user forget to override scope while using this dependency in submodule.

docs/bom.adoc
```diff
@@ -107,6 +107,17 @@ Then include the individual modules as you see fit, by simply depending on the u
...
 
+==== Arquillian Cube Docker Starter (Single Unified Dependency for Cube Docker in Standalone Mode)
+
+[source, xml]
+----
+<dependency>
+    <groupId>org.arquillian.cube</groupId>
+    <artifactId>arquillian-cube-docker-starter</artifactId>
+    <scope>test</scope>
+</dependency>
```
bartoszmajsak: How about sourcing these dependencies from the actual POM where it's defined? This way if we ever rename it we won't need to care about changes in the docs too.

lordofthejars:
@hemanik can you resolve conflict? @dipak-pawar after the merge conflict we are going to merge right?

dipak-pawar:
@lordofthejars Go ahead with merge after resolving conflicts.

Issue #935, opened 2018-01-09
Create a testing namespace for each test method


lordofthejars:
##### Issue Overview

Create a testing (random) namespace for each test method instead of each test suite.

This behaviour must be disabled by default and be configured as property to create it or not.

##### Expected Behaviour

Create a testing (random) namespace for each test method execution.

##### Current Behaviour

Namespace is created for each test suite execution.

dipak-pawar:
Would it be nice if we make it configurable on classlevel as well? So depending on configuration provided in `arquillian.xml` we'll create namespace for class level or method level. wdyt @lordofthejars @bartoszmajsak?

lordofthejars:
Yes, by default class level, but definitely it is a good idea.

lordofthejars:
We have tried to implement this but after it is not easy at all. First because if we create a namespace for class/method then we need to avoid creating a namespace for suite. This is not hard, but for any reason that we couldn't determinate if we call `SessionManagerLifecycle` class before class/method then we get some privilege errors. Probably when you create the namespace in suite scope, between the scope and the real deployment of resources there is something that makes everything work as expected.

Since this issue is not urgent and there are no requests regarding this, and on the other side there are some issues that are more urgent, we decided to park it for the moment.

PR #950, opened 2018-01-17
feat(#933): implements creating resources for method level templates

### LINKED ISSUE #933: Add @Template annotation in test method scope
##### Issue Overview

Add `@Template` annotation in test method scope.

##### Expected Behaviour

Allow defining `@Template` at method level, so the template is applied before executing the method and everything applied from that template removed after the test method.

##### Current Behaviour

`@Template` only works at class level

dipak-pawar:
Implements `@Template` in method scope.

- deploying templates defined over method using `@Template` annotation
- deleting resources created by template in after test
- Refactoring

Fixes #933


openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/resources/MethodTemplateProcessor.java
```diff
@@ -0,0 +1,127 @@
...
+    public List<Template> getTemplates() {
+        return templates;
+    }
+
+    /**
+     * Instantiates the templates specified by @Template within @Templates
+     */
+    public List<List<? extends OpenShiftResource>> processTemplateResources() {
+        List<? extends OpenShiftResource> resources;
+        final List<List<? extends OpenShiftResource>> RESOURCES = new ArrayList<>();
```
lordofthejars: I know this code comes from old impl. but can we name this variable in lower case to not confuse as a constant?

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/resources/TemplateProcessor.java
```diff
@@ -0,0 +1,40 @@
...
+import org.arquillian.cube.openshift.api.model.DeploymentConfig;
+import org.arquillian.cube.openshift.api.model.OpenShiftResource;
+import org.arquillian.cube.openshift.impl.adapter.OpenShiftAdapter;
+import org.arquillian.cube.openshift.impl.utils.Operator;
+import org.arquillian.cube.openshift.impl.utils.ParamValue;
+import org.arquillian.cube.openshift.impl.utils.TemplateUtils;
+
+public interface TemplateProcessor {
+
+    List<List<? extends OpenShiftResource>> processTemplateResources();
```
lordofthejars: Can we flat this to be a single List of `OpenShiftResources` instead of list of lists?

dipak-pawar: We have list of generics due to which it's not possible. See ref [[1]](https://docs.oracle.com/javase/tutorial/extra/generics/wildcards.html), [[2]](https://stackoverflow.com/questions/6504208/another-java-generic-question/6504576#6504576)

dipak-pawar: I was wrong here. We can do it by contravariance( widening a reference)

dipak-pawar:
@lordofthejars: We are waiting for this from long time. As we are blocked on namespace issue #935 let's merge it.

lordofthejars:
Ok, let's merge and if there is something we can always open another PR.

PR #959, opened 2018-01-19
feat(#935): makes namespace configurable at test method  and class scope levels.

### LINKED ISSUE #935: Create a testing namespace for each test method

hemanik:

Enables creating a temporary random namespace at test method scope and class scope for OpenShift extension. 

This feature is disabled by default and can be enabled by setting the property  `namespace.method.scope.enabled` and `namespace.class.scope.enabled` for test method and class scope respectively in `arquillian.xml` .

By default, the namespace is created at test suite level. 

- creates properties `namespace.method.scope.enabled` and `namespace.class.scope.enabled` for configuring the namespace scope.
- updates CEEnvironmentProcessor to create a temporary namespace at different scope levels.
- updates a ftest to test the proposed changes. 

#### Missing

- [ ] documentation

Fixes #935


openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/CEEnvironmentProcessor.java
```diff
@@ -93,6 +98,11 @@ public void createEnvironment(@Observes(precedence = 10) BeforeClass event, Open
         CubeOpenShiftConfiguration configuration) {
         final TestClass testClass = event.getTestClass();
         log.info(String.format("Creating environment for %s", testClass.getName()));
+
+        if (configuration.isNamespaceClassScopeEnabled()) {
+            namespaceService.get().create(configuration.getNamespace() + "-" + testClass.getJavaClass().hashCode());
```
dipak-pawar: With this it's quite difficult for user to debug which namaspace is created for which TestClass if he has to debug it. Can't we make it easy such a that we can understand from created namespace for which class it is created?

hemanik: The namespace can only have lowercase alphanumeric characters or `-` and should be less than 66 characters. 
I'll update if you can suggest something, having the name there doesn't work in spite of converting it to lowercase and replacing all the other characters such as `.`  etc with `-` because the length exceeds the maximum character limit. 
So, this was the only way, I could think of creating unique namespaces.

hemanik: One way for class could be `testClass.getJavaClass().getSimpleName().toLowerCase()` but then I am not sure of something equivalent for test method.

dipak-pawar: ok as we are not sure about TestClass length, then we can show  
provide log message for it. e.g.
`Created namespace `itest-5454237` for class org.arquillian.cube.FooTest`
`Created namespace  `itest-5454267` for class:  org.arquillian.cube.FooTest method: should_show_hello_world`

bartoszmajsak: > One way for class could be testClass.getJavaClass().getSimpleName().toLowerCase() but then I am not sure of something equivalent for test method.

Not on this level, as this handles class events, but for `Before event` you get the test method easily and then `testMethod.getName()` is enough.

bartoszmajsak: I believe we can always try to use this pattern and check if we are not exceeding the constraint of 63 characters. In such a case we can rollback to something else and log as @dipak-pawar suggested.

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/CEEnvironmentProcessor.java
```diff
@@ -263,4 +264,13 @@ private void delay(OpenShiftAdapter client, final List<? extends OpenShiftResour
     private void additionalCleanup(OpenShiftAdapter client, Map<String, String> labels) throws Exception {
         client.cleanRemnants(labels);
     }
+
+    private void createUniqueNamespace(CubeOpenShiftConfiguration configuration, String uniqueKey) {
+        namespace = configuration.getNamespace() + "-" + uniqueKey.toLowerCase().replaceAll("[^a-z0-9-]", "-");
+        if (namespace.chars().count() > 63) {
```
dipak-pawar: While creating namespace for method, it'll be good if we think for class name as well, As two classes can have same method name. Also we should  log it to avoid confusion of user about from which class this method belongs to.

namespace = `configuration.getNamespace() + "_" +  testClass + "_" + methodName`. 

If it has more than 63 characters then we should fall back to some readable name(e.g. `configuration.getNamespace() + randomUniqueId`) and log it like "fallback to "itest_5876667" as `namespace-testclass-methodName` is greater than 63 characters, so user knows the reason about trimming.

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/CEEnvironmentProcessor.java
```diff
@@ -106,11 +117,14 @@ public void createOpenShiftResource(@Observes(precedence = 10) Before event, Ope
 
         log.info(String.format("Creating environment for %s method %s", testClass.getName(), testMethod));
 
-        OpenShiftResourceFactory.createResources(testClass.getJavaClass(), client, testMethod, cubeOpenShiftConfiguration.getProperties());
+        if (cubeOpenShiftConfiguration.isNamespaceMethodScopeEnabled()) {
+            createUniqueNamespace(cubeOpenShiftConfiguration, testMethod.getName());
```
lordofthejars: maybe in case of method should be something like `testClass.getSimpleName() + "_" + testMethod.getName()`

hemanik:
> I think we should handle if user specify namespace.use.existing and namespace.class.scope.enabled or namespace.method.scope.enabled in arquillian.xml 

After trying this, I am slightly unconvinced that if one wishes to you use an existing namespace, why would he want to recreate it and destroy it for each method or for a case if one wishes to use an existing namespace say just for a particular test, I think we should require something like an annotation of the form `@Namespace("namespace")`  for that method and not depend on Arquillian.xml property for that.
Or maybe I just perceived and misunderstood what @dipak-pawar you were trying to suggest.

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/CEEnvironmentProcessor.java
```diff
@@ -130,6 +147,21 @@ public void deleteOpenShiftResource(@Observes(precedence = -10) After event, Ope
...
         CubeOpenShiftConfiguration configuration) throws Exception {
         OpenShiftResourceFactory.deleteEnvironment(event.getTestClass(), client, configuration, classTemplateProcessor.getTemplates());
+        if (configuration.isNamespaceClassScopeEnabled()) {
+            namespaceService.get().delete(configuration.getNamespace() + "-" + event.getTestClass().getJavaClass().hashCode());
+        }
     }
 
+    private void createUniqueNamespace(CubeOpenShiftConfiguration configuration, String uniqueKey) {
+        namespace = configuration.getNamespace() + "-" + uniqueKey.toLowerCase().replaceAll("[^a-z0-9-]", "-");
+        if (namespace.chars().count() > 63) {
```
bartoszmajsak: Can we make `63` a constant with the name suggesting what does it really mean? Is that k8s limit?

hemanik:
@bartoszmajsak, this is pending for the reason https://github.com/arquillian/arquillian-cube/issues/935#issuecomment-362602005 as mentioned by @lordofthejars.

PR #965, opened 2018-02-05
feat(874): creates kubernetes resource annotation.

### LINKED ISSUE #874: Creates @KubernetesResource to set resources in K8s cluster
##### Issue Overview

Currently we have `@OpenShiftResource` annotation for setting resources for each test.
We could implement the same feature but for Kubernetes. 

@jwendell @alesj Do you think it is worth to implement it? I think we cannot generialize because the openshift ones requires the `OpenshiftClient`

hemanik:

Creates `KubernetesResource` annotation to populate Kubernetes resources into the cluster before test execution.

- introduces `KubernetesResource` annotation and a functional test for the same.
- refactors `list annotations` for OpenShift.
- abstracts common logic for resource resolver for Istio, Templates, OpenShift and Kubernetes Resources. 

Fixes #874


kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/annotations/KubernetesResource.java
```diff
@@ -0,0 +1,57 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation to be used to populate Kubernetes resources into the cluster.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE, ElementType.ANNOTATION_TYPE})
+@Repeatable(KubernetesResource.List.class)
```
lordofthejars: Thanks for following Istio approach, this is the way we should implement in all places.

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/resources/KubernetesResourcesApplier.java
```diff
@@ -0,0 +1,65 @@
...
+import org.jboss.arquillian.test.spi.event.suite.AfterClass;
+import org.jboss.arquillian.test.spi.event.suite.BeforeClass;
+
+import java.io.BufferedInputStream;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public class KubernetesResourcesApplier {
```
lordofthejars: Can we add support for method scope? So you can define a Kubernetes resource at the method level. In Istio we are not offering this because I am not sure that it has sense at all since it is a really high-level routing feature.

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/adapter/KubernetesAdapter.java
```diff
@@ -0,0 +1,72 @@
...
+package org.arquillian.cube.kubernetes.impl.adapter;
+
+import io.fabric8.kubernetes.api.model.v3_1.HasMetadata;
+import io.fabric8.kubernetes.clnt.v3_1.KubernetesClient;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.List;
+
+public class KubernetesAdapter implements Adapter {
```
lordofthejars: Why an adapter? Cannot we add this lines directly in the `@Observes` method? An adapter is usually used when you want to adapt one class to another class, but not sure this is what we are doing here.

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/resources/KubernetesResourceResolver.java
```diff
@@ -0,0 +1,35 @@
+package org.arquillian.cube.kubernetes.impl.resources;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
+
+public class KubernetesResourceResolver {
```
lordofthejars: Now let's submit as is, but can you open a new issue to reduce the technical debt of this and create a common class that can be used everywhere? `Istio`, `Templates`, `OpenShiftResources` and `KubernetesResources`?

hemanik: Handled as part of this PR itself.

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/resources/OpenShiftResourceFactory.java
```diff
@@ -182,7 +166,7 @@ public static void deleteTemplates(final String templateKeyPrefix, Class<?> test
         OpenShiftAdapter openshiftAdapter, CubeOpenShiftConfiguration configuration)
         throws Exception {
         StringResolver resolver;
-        String templateURL;
+        InputStream templateURL;
```
dipak-pawar: Can you rename variable too to make it more readable?

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/resources/KubernetesResourcesApplier.java
```diff
@@ -0,0 +1,138 @@
...
+    private void deleteResources(String resourcesKey, KubernetesClient kubernetesClient) {
+        List<KubernetesResourceHandle> list = resourcesMap.remove(resourcesKey);
+        if (list != null) {
+            for (KubernetesResourceHandle resource : list) {
+                resource.delete(kubernetesClient);
+            }
+        }
+    }
+
+    private void addResourceHandle(String resourcesKey, KubernetesResourceHandle handle) {
```
dipak-pawar: How about this
```java
    private void addResourceHandle(String resourcesKey, KubernetesResourceHandle handle) {
        List<KubernetesResourceHandle> list = resourcesMap.computeIfAbsent(resourcesKey, k -> new ArrayList<>());
        list.add(handle);
    }
```

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/fabric8/F8OpenShiftAdapter.java
```diff
@@ -427,18 +401,17 @@ private void verifyServiceAccounts(DeploymentConfig dc) throws Exception {
...
-                template.setLabels(new HashMap<String, String>());
-            }
-            template.getLabels().putAll(labels);
-            return templateHandle.process(values.toArray(new ParameterValue[values.size()]));
+
+        TemplateResource<Template, KubernetesList, DoneableTemplate> templateHandle =
+            client.templates().inNamespace(configuration.getNamespace()).load(templateURL);
+        Template template = templateHandle.get();
+        if (template.getLabels() == null) {
+            template.setLabels(new HashMap<String, String>());
```
dipak-pawar: This can be replaced with `template.setLabels(new HashMap<>());`

PR #973, opened 2018-02-09
feat (#969): OpenshiftResourceInstaller now uses the configuration ob…

### LINKED ISSUE #969: Openshift templates, should be passed a default set of parameters, like scripts do.
Sometimes that are parameters passed to the openshift template, that are relative to the target environment (e.g. the project name: https://github.com/syndesisio/syndesis/blob/master/app/deploy/syndesis-restricted.yml#L31)

To deal with this cases in the past, we used to arquillian managed property values as environment variables to scripts. (See: https://github.com/arquillian/arquillian-cube/blob/0d6d2d758f71bd8a474ce33281cfa0609a2708c1/kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/SessionManager.java#L513-L513) 

We should do the same in the OpenshiftResourceInstaller.

iocanel:
Fixes #969

### DIFF
--- a/openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/install/OpenshiftResourceInstaller.java
+++ b/openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/install/OpenshiftResourceInstaller.java
@@ -13,29 +13,34 @@ import org.arquillian.cube.kubernetes.api.Logger;
 import org.arquillian.cube.kubernetes.api.ResourceInstaller;
 import org.arquillian.cube.kubernetes.impl.install.DefaultResourceInstaller;
 import org.arquillian.cube.kubernetes.impl.visitor.CompositeVisitor;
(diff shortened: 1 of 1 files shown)


openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/install/OpenshiftResourceInstaller.java
```diff
@@ -77,5 +85,17 @@ public ImmutableOpenshiftResourceInstaller(KubernetesClient client, Configuratio
                 throw KubernetesClientException.launderThrowable(t);
             }
         }
+
+        private Map<String, String> createBasicTemplateParameters() {
+            Map<String, String> env = new HashMap<>();
```
lordofthejars: Should we add system properties as well or it has no sense at all?

PR #974, opened 2018-02-12
chore(#968): moves parameter template.parameters.file to configuration.

### LINKED ISSUE #968: Openshift resource installer should read the `template.parameters.file` from the confiuration
Currently, it reads that from system properties. It should be part of the configuration object, so that it supports all 3 styles of passing configuration:

- arquillian.xml
- system properties
- environment variables.

hemanik:
Make parameter `template.parameters.file` configurable from 

- arquillian.xml
- system properties
- environment variables. 

- Moves `template.parameters.file` from openshift resource installer to configuration.  

Fixes #968


openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/CubeOpenShiftConfiguration.java
```diff
@@ -75,6 +76,7 @@
     private final String templateURL;
     private final String templateLabels;
     private final String templateParameters;
+    private final String templateParametersFile;
```
iocanel: CubeOpenshiftConfiguration already has template paramters. We don't need templatesParamtersFile too.
I'd say that OpenshiftResourceInstaller should depracted templateParmatersFile and just use templatesPrameters from configuration.

iocanel:
Now if, we remove the duplicating property, the PR will become a subset of #973. So, I think that we should close this one.

lordofthejars:
I close the PR since it becoms a subset of #973

Issue #964, opened 2018-02-02
Login to openshift/Kubernetes cluster before starting execution of tests


dipak-pawar:
Investigate is there any option for authentication with proper permission to openshift/kubernetes cluster while running arquillian cube tests

Parent task https://github.com/arquillian/arquillian-organization/issues/26

dipak-pawar:
We can run tests using system properties `-Dkubernetes.master=cluster_url` and `-Dkubernetes.auth.token=TOKEN` to authenitcate to openshift/k8s cluster before executing tests

lordofthejars:
But how about when using from IDE?

dipak-pawar:
We do have way to provide configuration in `arquillian.xml`, but I believe we should not expose token from arquillian.xml.

currently arquillian is looking for any system property with name `kubernetes.auth.token` and `kubernete.master` or env variable with `KUBERNETES_AUTH_TOKEN` and `KUBERNETES_MASTER`. So either set these values somehow before executing tests.

Do you have anything in mind other than this?

lordofthejars:
No, just thinking that would be great to fix it somehow, but probably it means changing how you run tests in Che.

dipak-pawar:
Also most important thing is you need just one time login to cluster, after that you don't need any cluster url or token, because cube will use cluster's current context stored in `~/.kube/config` on client,

lordofthejars:
exactly, so maybe we could do something like in cube we detect if we are in IDE (if we are in IDE probably checking classpath), then we check if there is the authentication file, if not then we can show a console message where user can introduce the token to authenticate. It is like a wizard but in console and just in case of being inside IDE.

dipak-pawar:
Good Idea @lordofthejars. @bartoszmajsak Do you have any thoughts?

bartoszmajsak:
In the context of OSIO switching clusters is not what we should promote (even if possible for the user to do). We have two namespaces we can potentially use to run our tests on:

- `user`
- `user`-stage

where `user` is your username, e.g. `bartosz-1` in my case.

dipak-pawar:
I am not sure should we use `user-stage` namespace to run our tests, because it's created by `devtools-sre ` and I am not sure about it's purpose. But definitely we can run tests in `user` namespace.

Now about switching cluster context, when you are inside osio, I think different user may belongs to different cluster. So We should ask for user which cluster are you on? or atleast using some api we need to fetch it to use it for single time login

[...]

PR #985, opened 2018-02-16
Now using OuthToken or username and password to authenticate remote cluster

### LINKED ISSUE #982: Authentication parameters are ignored in OpenShift
##### Issue Overview
Parameters such as `authToken` are ignored or does not work properly when used against remote clusters such as OSIO

##### Expected Behaviour

Log into the remote cluster using auth token provided in the configuration file.

##### Current Behaviour

Token is ignored and if you have not done before an `oc login`, test fails.

dipak-pawar:
Now using OuthToken or username and password to authenticate remote openshift cluster

- Adds `name`, `password`, `outhToken`, `trustCerts`, `apiVersion` propertis in configuration.
- Adds support for above properties for Kubernetes
- Adds documentation

Fixes #982 , #964


docs/kubernetes.adoc
```diff
@@ -74,7 +74,12 @@ You can configure Kubernetes by using any of the next configuration properties i
 |===
 | Options                             | Type           | Env | Description                                                   
 | kubernetes.master                   | URL            | Any | The URL to the Kubernetes master                               
-| kubernetes.domain                   | String         | OSE | Domain to use for creating routes for services                 
+| name                                | String         | Any | Username to log in server
```
lordofthejars: Why not username?

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/ClientCreator.java
```diff
@@ -34,15 +36,34 @@
     private InstanceProducer<KubernetesClient> producer;
 
     public void createClient(@Observes Configuration config) {
-        if (config.getMasterUrl() != null) {
+
+        if (config.getMasterUrl() == null) {
             producer.set(new DefaultKubernetesClient(new ConfigBuilder()
```
lordofthejars: Create `ConfigBuilder` outside `if` then depending on the values you populate the `ConfigBuilder`, i.e if(master!=null) then set masterUrl; if (token !=null) then set token;...

And then there is no if/else just ifs and at the end you construct the `DefaultKuebernetesClient` and inject in `InstanceProducer`

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/OpenShiftClientCreator.java
```diff
@@ -24,26 +25,37 @@ public void createClient(@Observes AfterStart afterStart, Configuration conf) {
         CubeOpenShiftConfiguration configuration = (CubeOpenShiftConfiguration) conf;
         System.setProperty("KUBERNETES_TRUST_CERT", "true");
         // override defaults for master and namespace
-        final Config config = new ConfigBuilder()
+        final ConfigBuilder configBuilder = new ConfigBuilder()
```
lordofthejars: Can you undo the changes of this class? My PR comes with a fix which is a totally different approach by reusing same client instead of instantiating a new one.

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/fabric8/F8OpenShiftAdapter.java
```diff
@@ -112,7 +112,7 @@ static OpenShiftConfig toOpenShiftConfig(CubeOpenShiftConfiguration configuratio
             .withMasterUrl(configuration.getKubernetesMaster())
             .withTrustCerts(configuration.isTrustCerts());
 
-        if (configuration.hasOpenshiftBasicAuth()) {
+        if (configuration.hasBasicAuth()) {
```
lordofthejars: Same for this class, my PR comes with the final solution by reusing the instance.

docs/kubernetes.adoc
```diff
@@ -74,7 +74,12 @@ You can configure Kubernetes by using any of the next configuration properties i
...
+| password                            | String         | Any | Password to log in server
```
lordofthejars: How about calling parameters `kubernetes.user.name` `kubernetes.user.password` and `kubernetes.user.token`? Any thoughts about this @bartoszmajsak

bartoszmajsak: Do we want to be specific here? Or the idea is to have generic concept of `user` and `pwd` to be applied depends on the target cluster/server/cloud provider (you name it) :)

bartoszmajsak: It might also be confusing, as now OpenShift is k8s, but not everyone knows that?

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/api/Configuration.java
```diff
@@ -54,6 +54,12 @@
     String LOGS_COPY = "logs.copy";
     String LOGS_PATH = "logs.path";
 
+    String USERNAME = "name";
```
bartoszmajsak: `username` to match?

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/ClientCreator.java
```diff
@@ -34,15 +36,34 @@
...
+                    }
+                });
+
+            if (Strings.isNotNullOrEmpty(config.getToken())) {
+                configBuilder.withOauthToken(config.getToken());
+            }
+
+            if (Strings.isNotNullOrEmpty(config.getUsername()) && Strings.isNotNullOrEmpty(config.getPassword())) {
+                configBuilder.withUsername(config.getUsername());
+                configBuilder.withPassword(config.getPassword());
```
bartoszmajsak: Detail - if it's fluent  - let's chain.

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/CubeOpenShiftConfiguration.java
```diff
@@ -285,11 +269,12 @@ public int getRouterSniPort() {
...
 
-        if ((token == null || token.isEmpty()) && (client != null)) {
-            token = client.getClientExt().getConfiguration().getOauthToken();
+        if ((super.getToken() == null || super.getToken().isEmpty()) && (client != null)) {
+          String token = client.getClientExt().getConfiguration().getOauthToken();
+          setToken(token);
         }
 
-        return token;
+        return super.getToken();
```
bartoszmajsak: Why `super.` here? Do we have any extra logic above to be concerned about?

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/api/Configuration.java
dipak-pawar: We are looking for system property and env variable. If we used `username`, then `System.getenv("USERNAME")` returns your host name. 

I tried `user.name` as well, but `System.getProperty("user.name")` returns your host name. thats why I am using `name`

bartoszmajsak: How about we prefix everything with `arq.cube` to have our own namespace? @lordofthejars wdyt?

docs/kubernetes.adoc
dipak-pawar: We are looking for system property and env variable. If we used username, then `System.getenv("USERNAME")` returns your host name.

I tried user.name as well, but `System.getProperty("user.name")` returns your host name. thats why I am using name

lordofthejars: ok, then leave it as is.

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/api/Configuration.java
lordofthejars: I don't know. I am neither so sure about it, maybe leave it as Diapk is the best approach now.

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/CubeOpenShiftConfiguration.java
dipak-pawar: Using `super.` covers both scenario. i.e.
1. When there is no token, with client not null, it should get token from config and set it. and then return it.
2.  If token is available in parent, then return it.

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/OpenShiftClientCreator.java
dipak-pawar: @lordofthejars I cannot see any parameters for authentication in your PR

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/api/Configuration.java
dipak-pawar: I will keep it the same for now.

lordofthejars:
About parameter names, I think we need to use some time in next cabal to unify all parameters docker, openshift and kubernetes, and we create an issue, and when we want to release  2.0.0 of Cube, then let's refactor all of them.

lordofthejars:
Good job @dipak-pawar Now we have fixed an old bug so now we are able to authenticate using token

PR #989, opened 2018-02-19
fix(#978): Adds openshift cluster 3.7.0 and 3.7.1 in CI

### LINKED ISSUE #978: Add support to run integration tests on openshift cluster v3.7.0 and v3.7.1 in Travis.

dipak-pawar:

- Adds oc version 3.7.0 an 3.7.1

Fixes #978

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,42 @@ before_install:
 
 install:
   - |
+      # OpenShift >= 3.7 requires nsenter https://github.com/openshift/origin/issues/17638
+      if [ ${OC_VERSION} == "v3.7.1" ] || [ ${OC_VERSION} == "v3.7.0" ] ; then
+          if ! which nsenter > /dev/null; then
+            echo "Did not find nsenter. Installing it."
+            NSENTER_BUILD_DIR=$(mktemp -d /tmp/nsenter-build-XXXXXX)
+            pushd ${NSENTER_BUILD_DIR}
+            curl https://www.kernel.org/pub/linux/utils/util-linux/v2.31/util-linux-2.31.tar.gz | tar -zxf-
+            cd util-linux-2.31
+            ./configure --without-ncurses
+            make nsenter
+            sudo cp nsenter /usr/local/bin
+            rm -rf "${NSENTER_BUILD_DIR}"
+            popd
+          fi
+          # OpenShift >= 3.7 requires systemd-run (but Ubuntu Trusty still uses Init)
+          # Should be in Kubernetes to correctly detect if systemd-run is present -
+          # Hack it till then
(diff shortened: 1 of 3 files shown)


bartoszmajsak:
The only one concern I have right now is that the build time will increase quite a bit. With 5 jobs (all started more or less in parallel) we were averaging around 16 minutes per build by looking at the recent trend. With 9 I'm afraid it will be +10 minutes, as I doubt we can have all 9 jobs executed in parallel. 

Can we prune some of the combinations then? WDYT @dipak-pawar @lordofthejars?

dipak-pawar:
@bartoszmajsak, Currently we are starting same cluster for k8s and openshift category, like following
```bash
OC_VERSION="v3.6.0", CATEGORY="RequiresKubernetes"
OC_VERSION="v3.6.0" CATEGORY="RequiresOpenshift"
OC_VERSION="v3.6.1" CATEGORY="RequiresKubernetes"
OC_VERSION="v3.6.1" CATEGORY="RequiresOpenshift"
OC_VERSION="v3.7.0" CATEGORY="RequiresKubernetes"
OC_VERSION="v3.7.0" CATEGORY="RequiresOpenshift"
OC_VERSION="v3.7.1" CATEGORY="RequiresKubernetes"
OC_VERSION="v3.7.1" CATEGORY="RequiresOpenshift"
```
For now may be we can think to run k8s and openshift tests on single node if time required to finish it is more than 5 minutes with current build(15 minutes  + 5  minutes)

```bash
OC_VERSION="v3.6.0", CATEGORY="RequiresKubernetes,RequiresOpenshift"
OC_VERSION="v3.6.1" CATEGORY="RequiresKubernetes,RequiresOpenshift"
OC_VERSION="v3.7.0" CATEGORY="RequiresKubernetes,RequiresOpenshift"
OC_VERSION="v3.7.1" CATEGORY="RequiresKubernetes,RequiresOpenshift"
```

I will check what will be time required to finish if we combine two categories

foo.sh
```diff
@@ -0,0 +1,3 @@
+if [ ${OC_VERSION} == "v3.7.1" ] ; then 
+  echo $OC_VERSION
```
bartoszmajsak: Why do we need that? :)

.travis.yml
```diff
@@ -43,6 +41,42 @@ before_install:
 
 install:
   - |
+      # OpenShift >= 3.7 requires nsenter https://github.com/openshift/origin/issues/17638
```
bartoszmajsak: How about externalizing it to its own file?

foo.sh
dipak-pawar: my bad, it was added mistakenly

PR #990, opened 2018-02-19
chore(#977): updates logger message to include FMP warning.

### LINKED ISSUE #977: Log when no kubernetes/openshift resources are found
Log when no kubernetes/openshift resources are found that if you are using FMP, then first you need to run Fabric8 Maven plugin first.

hemanik:
Logs warning to run fabric8-maven-plugin first, if FMP used and no Kubernetes resource found.

- Updates logger message.

Fixes #977

### DIFF
--- a/kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/SessionManager.java
+++ b/kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/SessionManager.java
@@ -163,8 +163,8 @@ public class SessionManager implements SessionCreatedListener {
                         resources.addAll(resourceInstaller.install(configUrl));
                     }
                 } else {
-                    log.warn(
-                        "Did not find any kubernetes configuration. If you are using fabric8-maven-plugin, ensure `mvn package fabric8:build` is run first.");
+                    log.warn("Did not find any kubernetes/openshift configuration files. If you are using fabric8-maven-plugin, "
+                        + "ensure mvn package fabric8:resource fabric8:build is run first to generate resources.");
                 }
 
                 List<HasMetadata> resourcesToWait = new ArrayList<>(resources);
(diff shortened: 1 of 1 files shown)


kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/SessionManager.java
```diff
@@ -165,7 +163,8 @@ public void createEnvironment() {
                         resources.addAll(resourceInstaller.install(configUrl));
                     }
                 } else {
-                    log.warn("Did not find any kubernetes configuration.");
+                    log.warn(
+                        "Did not find any kubernetes configuration. If you are using fabric8-maven-plugin, ensure `mvn package fabric8:build` is run first.");
```
dipak-pawar: I believe `mvn package fabric8:build` is not generating k8s/openshift configuration files, where it's building docker image. I couldn't find generated resources in target. Refer [build_log](https://gist.github.com/aa1611fc8e9fc46e41a574c6b0b0d97a).

When I tried `mvn package fabric8:resource`, then it created resources configuration. Refer [resource_log](https://gist.github.com/727912faa6872b4ad641fb324cee2118
)

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/SessionManager.java
```diff
@@ -165,7 +163,8 @@ public void createEnvironment() {
                         resources.addAll(resourceInstaller.install(configUrl));
                     }
                 } else {
-                    log.warn("Did not find any kubernetes configuration.");
+                    log.warn(
+                        "Did not find any kubernetes configuration. If you are using fabric8-maven-plugin, ensure `mvn package fabric8:resource fabric8:build` is run first.");
```
dipak-pawar: How about adding  `Did not find any kubernetes/openshift configuration.` instead of  `Did not find any kubernetes configuration.` to avoid confusing for end user.

lordofthejars:
One of the problems I see, and not sure how to solve it is that this message might appear even when you are using for example `OpenShiftAssistant` or for example when using `@KuberntesResource` at test level instead of global level. So I am not sure how to proceed in terms of what to put in the message. cc/ @bartoszmajsak

PR #994, opened 2018-02-21
fix(#984): reads configuration to initialize K8Client in Kubernetes/Openshift Requirement.

### LINKED ISSUE #984: KubernetesRequirement and OpenshiftRequirement doensn't look for Remote cluster if specified
Currently we do have [KubernetesRequirement](https://github.com/arquillian/arquillian-cube/blob/master/kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/requirement/KubernetesRequirement.java#L18) and [OpenshiftRequirement](https://github.com/arquillian/arquillian-cube/blob/master/openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/requirement/OpenshiftRequirement.java#L18) which checks cluster version, which is using kubernetes client to check version. But client is not initialized with configuration master url.

We should check configuration master url and initialize K8Client with configuration master url to work it properly in other env like osio che and pipeline if someone runs tests with `@RunWith(ArquillianConditionalRunner.class)`

hemanik:
1. Reads extension configuration to initialize K8Client with Remote Cluster in Kubernetes/Openshift Requirement.
2. Fixes IllegalArgumentException due to the addition of multiple org.jboss.shrinkwrap.api.Archive deployments with the same name: DEFAULT.

- reads extension configuration from `Arquillian Descriptor` exposed from [Arquillian Core](https://github.com/arquillian/arquillian-core/pull/165) in Kubernetes and OpenShift Requirement. 
- extracts ConfigBuilder logic to ClientConfigurator class.
- Updates Arquillian Core Version.
- Overrides AnnotationDeploymentScenarioGenerator service with the registered ExternalDeploymentScenarioGenerator service.
 
Fixes  https://github.com/arquillian/arquillian-core/issues/166,  #984

### DIFF
--- a/kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/ClientConfigurator.java
+++ b/kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/ClientConfigurator.java
@@ -1,0 +1,36 @@
+package org.arquillian.cube.kubernetes;
+
+import io.fabric8.kubernetes.api.builder.v3_1.TypedVisitor;
+import io.fabric8.kubernetes.clnt.v3_1.ConfigBuilder;
+import org.arquillian.cube.impl.util.Strings;
+import org.arquillian.cube.kubernetes.api.Configuration;
+
+public class ClientConfigurator {
+
+    public static ConfigBuilder getConfigBuilder(Configuration config) {
+        final ConfigBuilder configBuilder = new ConfigBuilder()
+            .withNamespace(config.getNamespace())
+            .withApiVersion(config.getApiVersion())
+            .withTrustCerts(config.isTrustCerts())
+            .accept(new TypedVisitor<ConfigBuilder>() {
+                @Override
+                public void visit(ConfigBuilder b) {
+                    b.withNoProxy(b.getNoProxy() == null ? new String[0] : b.getNoProxy());
+                }
+            });
+
+        if (Strings.isNotNullOrEmpty(config.getMasterUrl().toString())) {
+            configBuilder.withMasterUrl(config.getMasterUrl().toString());
+        }
+
+        if (Strings.isNotNullOrEmpty(config.getToken())) {
+            configBuilder.withOauthToken(config.getToken());
+        }
+
+        if (Strings.isNotNullOrEmpty(config.getUsername()) && Strings.isNotNullOrEmpty(config.getPassword())) {
+            configBuilder.withUsername(config.getUsername())
+                .withPassword(config.getPassword());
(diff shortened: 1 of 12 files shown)


kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/ClientConfigurator.java
```diff
@@ -0,0 +1,36 @@
+package org.arquillian.cube.kubernetes;
+
+import io.fabric8.kubernetes.api.builder.v3_1.TypedVisitor;
+import io.fabric8.kubernetes.clnt.v3_1.ConfigBuilder;
+import org.arquillian.cube.impl.util.Strings;
+import org.arquillian.cube.kubernetes.api.Configuration;
+
+public class ClientConfigurator {
```
dipak-pawar: I don't see any Client Configuration, where as I can see here `ClientConfigBuilder`.
This name is not relevant to what is it doing in the class. Also How about making this class to use something like
```java
new ClientConfigBuilder()
  .configuration(config)
  .build()
```

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/requirement/KubernetesRequirement.java
```diff
@@ -34,4 +43,16 @@ public void check(RequiresKubernetes context) throws UnsatisfiedRequirementExcep
                 "Error while checking kubernetes version: [" + e.getMessage() + "]");
         }
     }
+
+    private Config getConfig() {
```
dipak-pawar: I can see same code used in openshift as well except `extension name`. How about moving it to separate class, and use it both the places, to avoid duplication.

openshift/ftest-standalone/src/test/resources/arquillian.xml
```diff
@@ -6,6 +6,15 @@
 
   <extension qualifier="openshift">
     <property name="env.config.resource.name">hello-openshift.json</property>
+
+    <!-- configuring remote cluster -->
+
+    <!--
+    <property name="namespace.use.existing">existing-namespace</property>
+    <property name="kubernetes.master">https://api.starter-us-east-2.openshift.com</property>
+    <property name="cube.auth.token">token</property>
```
dipak-pawar: we don't need token and username/password at the same time, maybe OR will help in between for user to understand if we are going to keep this comment

dipak-pawar:
@hemanik Can you rebase and update this with latest core?

PR #995, opened 2018-02-22
fix(#988): creating container if it has destroyed state

### LINKED ISSUE #988: Allow DockerCube to create() a container when DESTROYED
##### Issue Overview

Due to our testsetup (arquillian, junit, multi-threaded surefire) we create one (subclass of) DockerCube, representing our testcontainer, which handles a number of testclasses by repeatedly creating, starting, stopping, destroying that testcontainer. However, since #957 / #958 calling create on a destroyed DockerCube does nothing due to the state check on 
https://github.com/arquillian/arquillian-cube/pull/958/files#diff-68554bf6fbadc1793c97b89f358bc69fR101

I can't think of a good reason why a destroyed container can't be recreated. Is it possible to modify the state-check so it also accepts DESTROYED when calling create() ?

If I overwrite the state value in our DockerCube subclass to set it to BEFORE_CREATE if it is DESTROYED our tests run as expected.

In essence, something like this happens:
DockerCube cube = new DockerCube();

cube.create();    // testclass 1
cube.start();
...
cube.stop();
cube.destroy();

cube.create();    // testclass 2
cube.start();
...

##### Expected Behaviour
Calling create() on a DESTROYED DockerCube recreates the container so that we can start() it once again.

##### Current Behaviour
Calling create() on a DESTROYED DockerCube does nothing, causing the subsequent start() to result in a "container not found".

##### Steps To Reproduce
I have no easily reproducible steps because the setup is complicated, meaning it would take time, and I think the problem is easily describable.

##### Additional Information
As this is more of an enhancement request I don't think any additional information is necessary, but if you would like more information please ask me :smiley:

dipak-pawar:

- Allow DockerCube to create docker container if it's destroyed

Fixes #988

### DIFF
--- a/docker/docker/src/main/java/org/arquillian/cube/docker/impl/model/DockerCube.java
+++ b/docker/docker/src/main/java/org/arquillian/cube/docker/impl/model/DockerCube.java
@@ -98,23 +98,22 @@ public class DockerCube extends BaseCube<CubeContainer> {
 
     @Override
     public void create() throws CubeControlException {
-        if (state != State.BEFORE_CREATE) {
-            return;
-        }
-        try {
-            lifecycle.fire(new BeforeCreate(id));
+        if (state == State.BEFORE_CREATE || state == State.DESTROYED) {
+            try {
+                lifecycle.fire(new BeforeCreate(id));
 
-            log.fine(String.format("Creating container with name %s and configuration %s.", id, configuration));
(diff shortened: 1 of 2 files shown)


dipak-pawar:
> Did you try steps leading to this issue reported in #998?

I haven't tried. But I believe this is what we are testing in unit test which I added as a part of this PR

PR #1002, opened 2018-02-26
chore(#981): enables printing k8s/openshift configuration parameters to console.

### LINKED ISSUE #981: Print in console configuration parameters of K8s/OS
##### Issue Overview

Print in console configuration parameters of K8s/OS

##### Expected Behaviour

In Cube Docker we are printing in console when the built is started all the configuration parameters that are used such as docker host, port, ... it will be great to have the same in Kubernetes/OpeShift configuration objects, so we can know in any moment which parameters are used.

##### Current Behaviour

No conf is printed

hemanik:
Enabling printing Kubernetes and OpenShift configuration parameters in Console.

- defines toString() methods for Kubernetes and OpenShift Configuration classes.
- makes calls to Kubernetes and OpenShift Configuration classes' toString() methods from respective configuration factory classes.

Fixes #981

### DIFF
--- a/kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/DefaultConfiguration.java
+++ b/kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/DefaultConfiguration.java
@@ -451,81 +451,81 @@ public class DefaultConfiguration implements Configuration {
     @Override
     public String toString() {
 
-        String SEP = System.getProperty("line.separator");
+        String lineSeparator = System.lineSeparator();
         StringBuilder content = new StringBuilder();
 
-        content.append("CubeKubernetesConfiguration: ").append(SEP);
+        content.append("CubeKubernetesConfiguration: ").append(lineSeparator);
         if (namespace != null) {
-            content.append("  ").append(NAMESPACE).append(" = ").append(namespace).append(SEP);
+            content.append("  ").append(NAMESPACE).append(" = ").append(namespace).append(lineSeparator);
         }
         if (masterUrl != null) {
-            content.append("  ").append(MASTER_URL).append(" = ").append(masterUrl).append(SEP);
+            content.append("  ").append(MASTER_URL).append(" = ").append(masterUrl).append(lineSeparator);
         }
         if (scriptEnvironmentVariables != null) {
-            content.append("  ").append(ENVIRONMENT_SCRIPT_ENV).append(" = ").append(scriptEnvironmentVariables).append(SEP);
+            content.append("  ").append(ENVIRONMENT_SCRIPT_ENV).append(" = ").append(scriptEnvironmentVariables).append(lineSeparator);
         }
         if (environmentSetupScriptUrl != null) {
-            content.append("  ").append(ENVIRONMENT_SETUP_SCRIPT_URL).append(" = ").append(environmentSetupScriptUrl).append(SEP);
+            content.append("  ").append(ENVIRONMENT_SETUP_SCRIPT_URL).append(" = ").append(environmentSetupScriptUrl).append(lineSeparator);
         }
 
         if (environmentTeardownScriptUrl != null) {
-            content.append("  ").append(ENVIRONMENT_TEARDOWN_SCRIPT_URL).append(" = ").append(environmentTeardownScriptUrl).append(SEP);
+            content.append("  ").append(ENVIRONMENT_TEARDOWN_SCRIPT_URL).append(" = ").append(environmentTeardownScriptUrl).append(lineSeparator);
         }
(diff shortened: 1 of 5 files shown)


kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/DefaultConfiguration.java
```diff
@@ -447,4 +447,86 @@ public boolean isTrustCerts() {
     protected void setToken(String token) {
         this.token = token;
     }
+
+    @Override
+    public String toString() {
+
+        String SEP = System.getProperty("line.separator");
```
lordofthejars: There is a method called `System.lineSepartor()` that returns the same. Also name the variable `lineSeparator` so it does not confuse a reader to think that this is a const

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/DefaultConfigurationFactory.java
```diff
@@ -41,6 +41,8 @@ protected static void configureProtocolHandlers(Map<String, String> conf) {
     public C create(ArquillianDescriptor arquillian) {
         Map<String, String> config = arquillian.extension(KUBERNETES_EXTENSION_NAME).getExtensionProperties();
         configureProtocolHandlers(config);
-        return (C) DefaultConfiguration.fromMap(config);
+        final DefaultConfiguration configuration = DefaultConfiguration.fromMap(config);
+        System.out.println(configuration);
```
lordofthejars: Two things to check here:

1) Are we using `sysout` in Docker part as well? Now I don't remember exactly if we changed to `logger` or not.
2) Have you checked if at this stage in case of using random namespace (automatically generated) is already set or not? I think (but I have not tried) that in that specific case the namespace will not be shown. So I suggest moving this printing in a stage where it has been already calculated and set. (For example `BeforeSuite` event). But something to check before doing the change.

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/CubeOpenShiftConfiguration.java
```diff
@@ -334,4 +335,135 @@ public long getHttpClientTimeout() {
...
+
+    @Override
+    public String toString() {
+
+        String SEP = System.getProperty("line.separator");
+        StringBuilder content = new StringBuilder();
+
+        content.append("CubeOpenShiftConfiguration: ").append(SEP);
+
+        if (getNamespace() != null) {
```
lordofthejars: Can you remove all parameters that are already set in kubernetes config class, so the process of stringify common parameters are done in the same place.

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/client/CubeOpenShiftConfigurationFactory.java
```diff
@@ -32,6 +32,9 @@ public CubeOpenShiftConfiguration create(ArquillianDescriptor arquillian) {
         config.putAll(arquillian.extension(OPENSHIFT_EXTENSION_NAME).getExtensionProperties());
 
         configureProtocolHandlers(config);
-        return CubeOpenShiftConfiguration.fromMap(config);
+
+        final CubeOpenShiftConfiguration openShiftConfiguration = CubeOpenShiftConfiguration.fromMap(config);
+        System.out.println(openShiftConfiguration);
```
lordofthejars: Same as Kubernetes comment.

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/DefaultConfigurationFactory.java
hemanik: > 1. Are we using `sysout` in Docker part as well? Now I don't remember exactly if we changed to `logger` or not.

It is still using `sysout` - https://github.com/arquillian/arquillian-cube/blob/4bed6df837f24d06813b6047cd223c81f508f6a6/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/CubeDockerConfigurator.java#L75 
Should this be changed to `logger`? 

> 2. Have you checked if at this stage in case of using random namespace (automatically generated) is already set or not? I think (but I have not tried) that in that specific case the namespace will not be shown. So I suggest moving this printing in a stage where it has been already calculated and set. (For example BeforeSuite event). But something to check before doing the change.

Yes, I checked. The namespace is set in the case when it is automatically generated.

screenshot from 2018-02-26 16-07-44

lordofthejars:
Good work @hemani, thank you very much.

PR #1007, opened 2018-02-28
fix(#1006): adds state checks in BuildablePodCube stop and destroy lifecycle events.

### LINKED ISSUE #1006: Ftests throw CubeControlException: Could not stop hello-openshift
##### Issue Overview
When executing `HelloPodOpenShiftITCase` or `HelloPodDeploymentOpenShiftITCase` ftest the following exception is thrown.
```
Exception in thread "Thread-1" org.arquillian.cube.spi.CubeControlException: Could not stop hello-openshift
	at org.arquillian.cube.spi.CubeControlException.failedStop(CubeControlException.java:23)
	at org.arquillian.cube.openshift.impl.model.BuildablePodCube.stop(BuildablePodCube.java:153)
	at org.arquillian.cube.impl.client.ForceStopDockerContainersShutdownHook.lambda$attachShutDownHookForceStopDockerContainers$0(ForceStopDockerContainersShutdownHook.java:17)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.NullPointerException
	at org.jboss.arquillian.core.impl.ManagerImpl.fire(ManagerImpl.java:128)
	at org.jboss.arquillian.core.impl.ManagerImpl.fire(ManagerImpl.java:105)
	at org.jboss.arquillian.core.impl.EventImpl.fire(EventImpl.java:62)
	at org.arquillian.cube.openshift.impl.model.BuildablePodCube.stop(BuildablePodCube.java:142)
	... 2 more
```

##### Expected Behaviour
After the session is destroyed the pods should be gracefully stopped and destroyed without any exceptions or errors.

##### Current Behaviour
`CubeControlException: Could not stop hello-openshift` is caused due to NPE.

##### Steps To Reproduce
1. execute  `HelloPodOpenShiftITCase` or `HelloPodDeploymentOpenShiftITCase` ftest.

hemanik:

Adds state checks before `stop` and `destroy` events for BuildablePodCube to fix NPE.

- Adds state checks before `stop` and `destroy` events for BuildablePodCube
- Adds Unit Tests for the Lifecycle Events. 

Fixes: #1006

### DIFF
--- a/openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/CubeOpenshiftExtension.java
+++ b/openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/CubeOpenshiftExtension.java
@@ -1,26 +1,37 @@
 package org.arquillian.cube.openshift.impl;
 
 import org.arquillian.cube.impl.client.enricher.StandaloneCubeUrlResourceProvider;
-import org.arquillian.cube.kubernetes.api.*;
+import org.arquillian.cube.kubernetes.api.ConfigurationFactory;
+import org.arquillian.cube.kubernetes.api.FeedbackProvider;
+import org.arquillian.cube.kubernetes.api.KubernetesResourceLocator;
+import org.arquillian.cube.kubernetes.api.NamespaceService;
+import org.arquillian.cube.kubernetes.api.ResourceInstaller;
 import org.arquillian.cube.kubernetes.impl.DefaultConfigurationFactory;
 import org.arquillian.cube.kubernetes.impl.enricher.KuberntesServiceUrlResourceProvider;
 import org.arquillian.cube.kubernetes.impl.feedback.DefaultFeedbackProvider;
 import org.arquillian.cube.kubernetes.impl.install.DefaultResourceInstaller;
 import org.arquillian.cube.kubernetes.impl.locator.DefaultKubernetesResourceLocator;
 import org.arquillian.cube.kubernetes.impl.namespace.DefaultNamespaceService;
-import org.arquillian.cube.openshift.impl.client.*;
+import org.arquillian.cube.openshift.impl.client.CubeOpenShiftConfigurationFactory;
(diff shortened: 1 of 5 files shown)


openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/CubeOpenshiftExtension.java
```diff
@@ -69,7 +70,7 @@ public void register(ExtensionBuilder builder) {
         if (Validate.classExists("org.jboss.arquillian.container.test.spi.client.deployment.AuxiliaryArchiveAppender")
             && doesNotContainStandaloneExtension()) {
             builder.service(AuxiliaryArchiveAppender.class, UtilsArchiveAppender.class);
-            builder.service(DeploymentScenarioGenerator.class, ExternalDeploymentScenarioGenerator.class);
+            builder.override(DeploymentScenarioGenerator.class, AnnotationDeploymentScenarioGenerator.class, ExternalDeploymentScenarioGenerator.class);
```
dipak-pawar: I am not sure why are we overriding this as a part of this PR. wouldn't it be a good to add it when you are updating version of core in cube, so we can relate it in the future?

hemanik: No strong opinion about it. But as discussed with @lordofthejars, since it was a small change we thought of adding it as part of this PR since both the issues were related to the same tests.

hemanik: Can move it to https://github.com/arquillian/arquillian-cube/pull/994 or create a new PR for this? @dipak-pawar, what do you suggest?

dipak-pawar: I would move it to #994, as we found it there.

PR #1013, opened 2018-03-13
Integrate FMP with cube

### LINKED ISSUE #806: [Arquillian OpenShift] Creates an integration with Fabric8 plugin
##### Issue Overview

Creates an integration with Fabric8 plugin because when Fabric8 maven plugin is run it takes the default namespace configured (or the one that it is active in `oc` configuration) and do some stuff like creating ImageStreams or configuring builds.

The problem is that when running with Cube, you need to specify the namespace where Fabric8 has done all the configuration steps. That's fine in most cases, but if you want to use the random namespace generation provided by Cube, then the test will fail because this namespace created by Cube was not the one configured by Fabric8.

##### Expected Behaviour

The solution would be integrating Fabric8 plugin with Cube by either:

- we can re-use logic of fabric8-maven-plugin to generate those files create s2i image and register it in openshift registry this can be done under the hood based on e.g `@AutogenerateOpenshiftResources` annotation put on the test class
 - we can provide Shrinkwrap descriptors so you can either create default files like fmp does but also open it up for customization "shrinkwrap style "this would mean we have to flesh out what kind of `@Deployment` methods we would like to have to make it simple and powerful enough

My recommendation would be as first try, the first approach.

##### Current Behaviour

No integration

##### Additional Information

This bug comes from https://github.com/arquillian/arquillian-cube/issues/743

dipak-pawar:
This PR adds fmp integration with cube

- Introduces two properties i.e. `fmp.build` and `fmp.path` to configure arquillian cube.
- Generating resources for fmp using embedded maven.
- By default and from `mvn build`, this feature will be disabled.
- Adds IT 

Fixes #806


kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/SessionManager.java
```diff
@@ -220,6 +231,15 @@ public void start() {
         }
     }
 
+    private boolean isRunningFromMaven() {
```
lordofthejars: Do you think is it worth to move this logic into a "util" class to be reused somewhere else? Or for now leave it as is?

kubernetes/kubernetes/src/test/java/org/arquillian/cube/kubernetes/ResourceGeneratorBuilderIT.java
```diff
@@ -0,0 +1,40 @@
...
+import org.junit.experimental.categories.Category;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@Category(RequiresOpenshift.class)
+@RequiresOpenshift
+public class ResourceGeneratorBuilderIT {
+
+    @ClassRule
+    public static final GitClone gitClone = new GitClone("https://github.com/snowdrop/spring-boot-http-booster");
```
bartoszmajsak: Can't we make it local as a bundle? We know that at this point of time this repo has all we need, but it might not be the case in the future. So if we really want to use git for this test we should lock it down to a known state.

But then it opens a question - do we really need to use git for this tests? Can't we have some super tiny project as test-resource folder instead? Did you consider such approach?

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/ResourceGeneratorBuilder.java
```diff
@@ -0,0 +1,44 @@
+package org.arquillian.cube.kubernetes;
+
+import java.nio.file.Path;
+import org.jboss.shrinkwrap.resolver.api.maven.embedded.BuiltProject;
+import org.jboss.shrinkwrap.resolver.api.maven.embedded.EmbeddedMaven;
+
+public class ResourceGeneratorBuilder {
```
bartoszmajsak: Should it really belong to `kubernetes` module?

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/SessionManager.java
```diff
@@ -208,6 +210,15 @@ public void start() {
         Logger log = session.getLogger();
         log.status("Using Kubernetes at: " + client.getMasterUrl());
         createNamespace();
+
+        // generate fabric8 resources using build and resource goals
```
bartoszmajsak: Maybe instead of a comment here explaining what is used and how we could rework `ResourceGeneratorBuilder` to be more specific of what it really does? As it's anyway very much related to `fabric8` plugin we could equally well rename the class too.

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/SessionManager.java
```diff
@@ -220,6 +231,15 @@ public void start() {
         }
     }
 
+    private boolean isRunningFromMaven() {
+        try {
+            SessionManager.class.getClassLoader().loadClass("org.apache.maven.surefire.booter.ForkedBooter");
```
bartoszmajsak: Does it also work when we don't fork surefire/failsafe test execution? See http://maven.apache.org/surefire/maven-surefire-plugin/examples/fork-options-and-parallel-execution.html for more details

openshift/ftest-openshift-fmp/pom.xml
```diff
@@ -0,0 +1,91 @@
...
+	<modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <artifactId>arquillian-cube-openshift-parent</artifactId>
+    <groupId>org.arquillian.cube</groupId>
+    <version>1.15.3-SNAPSHOT</version>
+    <relativePath>../pom.xml</relativePath>
+  </parent>
+
+  <name>Arquillian Cube FMP Integration Functional Test</name>
```
bartoszmajsak: Maybe instead of `FMP` we can expand this abbreviation? At least in the name and description. I think we are fine with `artifactId`.

openshift/ftest-openshift-fmp/pom.xml
```diff
@@ -0,0 +1,91 @@
...
+
+  <artifactId>arquillian-cube-openshift-fmp-ftest</artifactId>
+
+	<properties>
+		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+		<project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+		<java.version>1.8</java.version>
+		<version.spring.boot>2.0.0.RELEASE</version.spring.boot>
+    <version.fmp>3.5.38</version.fmp>
```
bartoszmajsak: Here too - let's be more verbose in the name (plus formatting)

openshift/ftest-openshift-fmp/pom.xml
```diff
@@ -0,0 +1,91 @@
...
+    <version>1.15.3-SNAPSHOT</version>
+    <relativePath>../pom.xml</relativePath>
+  </parent>
+
+  <name>Arquillian Cube FMP Integration Functional Test</name>
+
+  <artifactId>arquillian-cube-openshift-fmp-ftest</artifactId>
+
+	<properties>
+		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
```
bartoszmajsak: Aren't those things inherited from parent `pom.xml`?

openshift/ftest-openshift-fmp/src/test/java/org/arquillian/cube/OpenshiftIT.java
```diff
@@ -0,0 +1,41 @@
...
+    private URL baseURL;
+
+    @Before
+    public void setup() throws Exception {
+        RestAssured.baseURI = baseURL.toString();
+    }
+
+    @Test
+    public void testGreetingEndpoint() {
+        given()
```
bartoszmajsak: We don't need `given()` block if we do not set any test fixture. Makes it confusing this way.

docs/kubernetes.adoc
```diff
@@ -120,6 +120,8 @@ filesystem - as individual files, one for each pod. Filenames will be "ClassName
 If the pod has multiple containers, one log file for each container will be created. Kubernetes events (`kubectl get events`)
 will also be captured if this flag is enabled. Filenames will end with `-KUBE_EVENTS.log`
 | logs.path | String | Any | Directory where to save the pods logs. Defaults to "target/surefire-reports".
+| fmp.path  | String | Any | Path to pom.xml where fmp plugin configuration is present. Defaults to "pom.xml".
+| fmp.build | Bool (false) | Any | Whether to generate resources using fmp when running test from IDE
```
bartoszmajsak: How about giving documentation a bit more love and document this feature on its own too? Also as noted below - does "fmp" really belongs to "k8s" module? Shouldn't it be a module on its own? WDYT @lordofthejars?

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/ResourceGeneratorBuilder.java
```diff
@@ -0,0 +1,44 @@
...
+            .forProject(pom.toFile())
+            .setQuiet()
+            .useDefaultDistribution()
+            .setDebugLoggerLevel()
+            .setGoals(goals)
+            .addProperty("fabric8.namespace", namespace)
+            .ignoreFailure()
+            .build();
+
+        System.out.println(builtProject.getMavenLog());
```
bartoszmajsak: Do we want to always print it out? Can we have it configurable for the user?

docs/kubernetes.adoc
lordofthejars: Yes I think that FMP deserves its own section with an example as well.

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/ResourceGeneratorBuilder.java
lordofthejars: We can configure but I will say `true` by default so the user can see what is happening under the cover.

kubernetes/kubernetes/src/test/java/org/arquillian/cube/kubernetes/ResourceGeneratorBuilderIT.java
dipak-pawar: Good idea.Tiny project as test-resource folder would work for this.

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/ResourceGeneratorBuilder.java
dipak-pawar: I was thinking the same, but not sure should we create separate module for single class.

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/SessionManager.java
dipak-pawar: yes it works in both the cases

docs/fabric8-maven-plugin.adoc
```diff
@@ -0,0 +1,42 @@
+= Cube Integration Fabric8 Maven Plugin
```
bartoszmajsak: Shouldn't a chapter start with `==` ? Did you render the doc to see how it fits to the rest of the content?

I'm getting:

```
asciidoctor: ERROR: docs/fabric8-maven-plugin.adoc: line 1: only book doctypes can contain level 0 sections
```

docs/fabric8-maven-plugin.adoc
```diff
@@ -0,0 +1,42 @@
...
+
+If you are using https://maven.fabric8.io/[fabric8-maven-plugin] to build images, generate resources and deploy your application to cluster, then
+you should use the same approach for integration testing using cube.
```
bartoszmajsak: Strange line break

docs/fabric8-maven-plugin.adoc
```diff
@@ -0,0 +1,42 @@
...
+
+However there is way to build images and generate resources using `fabric8-maven-plugin` as a part of test execution cycle.
```
bartoszmajsak: > there is a way

>  as a part of test execution cycle.

I would drop `cycle` from here.

docs/fabric8-maven-plugin.adoc
```diff
@@ -0,0 +1,42 @@
...
+If you enable `cube.fmp.build` property, cube will build image and generate resources for you by running embedded maven build as per options provided in
+`arquillian.xml` using `mvn package fabric8:build fabric8:resource -Dfabric8.namespace=${namespace_configured_for_cube}`.
+
+You can use following configuration options with kubernetes and openshift extension to enable fabric8 maven plugin build in test execution.
+
+*Configuration Options*
+[cols="2,1,1,3", options="header"]
+|===
+| Options                             | Type           | Env | Description
+| cube.fmp.build | Bool (false) | Any | Whether to generate resources using fmp when running test from maven or IDE
```
bartoszmajsak: `Maven` uppercase I believe, as it's its own name.

docs/fabric8-maven-plugin.adoc
bartoszmajsak: Also note, that just after that there is another section included called `fabric8` - shouldn't we merge those?

docs/fabric8-maven-plugin.adoc
```diff
@@ -0,0 +1,42 @@
...
+        <artifactId>fabric8-maven-plugin</artifactId>
+        <version>${version.fabric8.maven.plugin}</version>
+      </plugin>
+    </plugins>
+  </build>
+```
+- If you have multiple profiles defined in your pom.xml, you can enable profile in which you have
+fabric8-maven-plugin dependency by using `cube.fmp.profiles` in arquillian.xml.
+
+You can see the example using `fabric8-maven-plugin` at: https://github.com/dipak-pawar/arquillian-cube/tree/master/openshift/ftest-openshift-fabric8-maven-plugin
```
bartoszmajsak: Change this link already - it will become valid when we merge.

fabric8-maven-plugin-build/src/main/java/org/fabric8/maven/plugin/build/Fabric8MavenPluginResourceGeneratorBuilder.java
```diff
@@ -0,0 +1,126 @@
...
+        return this;
+    }
+
+    public Fabric8MavenPluginResourceGeneratorBuilder withProperties(List<String> propertiesPairs) {
+        return withProperties(propertiesPairs.toArray(new String[propertiesPairs.size()]));
+    }
+
+    public Fabric8MavenPluginResourceGeneratorBuilder withProperties(String... propertiesPairs) {
+        if (propertiesPairs.length % 2 != 0) {
+            throw new IllegalArgumentException("Expecting even amount of variable name - value pairs to be passed. Got "
```
bartoszmajsak: Minor detail here: we could do `String.format` here.

fabric8-maven-plugin-build/src/main/java/org/fabric8/maven/plugin/build/Fabric8MavenPluginResourceGeneratorBuilder.java
```diff
@@ -0,0 +1,126 @@
...
+        final ConfigurationDistributionStage distributionStage = EmbeddedMaven
+            .forProject(pom.toFile())
+            .setQuiet(quietMode)
+            .useDefaultDistribution()
+            .setDebug(mvnDebugOutput)
+            .setDebugLoggerLevel()
+            .setGoals(goals)
+            .addProperty("fabric8.namespace", namespace);
+
+        //Remove this after fixing issue  https://github.com/eclipse/che/issues/9105
```
bartoszmajsak: If we intend to merge it before the issue in Che is fixed, please mark it with `TODO` or `FIXME` and create issue in Cube so we won't forget about it.

fabric8-maven-plugin-build/src/test/java/org/fabric8/maven/plugin/build/ResourceGeneratorBuilderIT.java
```diff
@@ -0,0 +1,58 @@
...
+
+@Category(RequiresOpenshift.class)
+@RequiresOpenshift
+public class ResourceGeneratorBuilderIT {
+
+    @Rule
+    public TemporaryFolder temporaryFolder = new TemporaryFolder();
+
+    @Test
+    public void should_generate_build_and_resources() throws IOException {
```
bartoszmajsak: Minor detail: I'm not sure if we generate the build per se. Might be a bit misleading name later.

fabric8-maven-plugin-build/src/test/resources/spring-boot-http-booster/pom.xml
```diff
@@ -0,0 +1,141 @@
...
+
+  <artifactId>spring-boot-rest-http</artifactId>
+  <version>1.5.10-5-SNAPSHOT</version>
+
+  <name>Spring Boot - HTTP Booster</name>
+  <description>Spring Boot - HTTP Booster</description>
+
+
+  <properties>
+    <arquillian-cube.version>1.15.3-SNAPSHOT</arquillian-cube.version>
```
bartoszmajsak: TODO: we will have to fix it somehow when we merge with master :)

fabric8-maven-plugin-build/src/test/resources/spring-boot-http-booster/pom.xml
```diff
@@ -0,0 +1,141 @@
...
+    <groupId>io.openshift.booster</groupId>
+    <artifactId>spring-boot-booster-parent</artifactId>
+    <version>1.5.10-15</version>
+  </parent>
+
+  <artifactId>spring-boot-rest-http</artifactId>
+  <version>1.5.10-5-SNAPSHOT</version>
+
+  <name>Spring Boot - HTTP Booster</name>
+  <description>Spring Boot - HTTP Booster</description>
```
bartoszmajsak: Maybe add the description why we really have this project here.

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/DefaultConfiguration.java
```diff
@@ -535,6 +592,19 @@ public String toString() {
         }
         content.append("  ").append(TRUST_CERTS).append(" = ").append(trustCerts).append(lineSeparator);
 
+        content.append("  ").append(FMP_BUILD).append(" = ").append(fmpBuildEnabled).append(lineSeparator);
+        content.append("  ").append(FMP_BUILD_DISABLE_FOR_MAVEN).append(" = ").append(fmpBuildForMavenDisable).append(lineSeparator);
+        content.append("  ").append(FMP_POM_PATH).append(" = ").append(fmpPomPath).append(lineSeparator);
+        content.append("  ").append(FMP_DEBUG_OUTPUT).append(" = ").append(fmpDebugOutput).append(lineSeparator);
+        content.append("  ").append(FMP_LOGS).append(" = ").append(fmpLogsEnabled).append(lineSeparator);
```
bartoszmajsak: We have a lot of identical lines where the only difference is key and value as param. Maybe worth making a method for it?

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/locator/DefaultKubernetesResourceLocator.java
```diff
@@ -25,6 +33,41 @@ public URL locate() {
         return null;
     }
 
+    @Override
+    public URL locateFromTargetDir() {
+        for (String resource : getResourceNames()) {
+            for (String suffix : getAllowedSuffixes()) {
```
bartoszmajsak: This looks to me a bit too heavy for the problem we are aiming to solve. Could you try to just use `Files.walk` with filtering based on what ` getResourceNames()` and ` getAllowedSuffixes()` contain? These can be sets which we can use for the conditions.

docs/fabric8-maven-plugin.adoc
dipak-pawar: yes that's correct, it will become valid when we merge.

kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/DefaultConfiguration.java
dipak-pawar: I will make another PR for it.

bartoszmajsak: Sure, makes sense.

fabric8-maven-plugin-build/src/main/java/org/fabric8/maven/plugin/build/Fabric8MavenPluginResourceGeneratorBuilder.java
```diff
@@ -0,0 +1,124 @@
...
+            .setQuiet(quietMode)
+            .useDefaultDistribution()
+            .setDebug(mvnDebugOutput)
+            .setDebugLoggerLevel()
+            .setGoals(goals)
+            .addProperty("fabric8.namespace", namespace);
+
+        // TODO: https://github.com/arquillian/arquillian-cube/issues/1017
+        if (System.getenv("JAVA_HOME") == null) {
+            distributionStage.addShellEnvironment("JAVA_HOME", "/usr/lib/jvm/java-1.8.0");
```
bartoszmajsak: We might want to log some warning here too? WDYT? If it will happen for someone who doesn't have java_home but it's still not in the given location such warning could help.

e.g. 

> No `JAVA_HOME` defined. Defaulting to `/usr/lib/jvm/java-1.8.0`. If that's not where it should be, please define `JAVA_HOME` environment variable and re-run this test. See https://git.io/vxWo9 for reasons.

fabric8-maven-plugin-build/src/test/resources/spring-boot-http-booster/pom.xml
```diff
@@ -0,0 +1,140 @@
...
+  </parent>
+
+  <artifactId>spring-boot-rest-http</artifactId>
+  <version>1.5.10-5-SNAPSHOT</version>
+
+  <name>Spring Boot - HTTP Booster</name>
+  <description>Booster for FMP Integration Test with Cube</description>
+
+  <properties>
+    <arquillian-cube.version>1.15.3-SNAPSHOT</arquillian-cube.version>
```
bartoszmajsak: Did you try potential fixes we discussed?

dipak-pawar: Yes I tried. Currently we are setting `arquillian-cube.version` from failsafe plugin https://github.com/arquillian/arquillian-cube/pull/1013/commits/9fff3558f204a81159e3784f35589255a7cca94c#diff-0c972d5eaf7f2ea01468a7f9162f3ae5R52, which we are passing it to embedded maven build from [test](https://github.com/arquillian/arquillian-cube/pull/1013/commits/9fff3558f204a81159e3784f35589255a7cca94c#diff-519e7e2fb47df75b6fe949bceb6c7687R39). So we are good for maven build.

bartoszmajsak: But then we don't really need this to be set up here or we set it to be released version (soon to be :)) so it will work regardless. I think I like the second version more. Then we have to update the comment in the test that only if you want to run it from the IDE against latest code changes you have to change it.

dipak-pawar: I think we don't need this version to be set in this project as we have default value to `1.15.3-SNAPSHOT` in [test](https://github.com/arquillian/arquillian-cube/pull/1013/commits/9fff3558f204a81159e3784f35589255a7cca94c#diff-519e7e2fb47df75b6fe949bceb6c7687R39). So I can change it to `1.15.3` and update this property in pom.xml to `<arquillian-cube.version/>

bartoszmajsak:
Great work @dipak-pawar. This deserves a celebration gif!

triumph-colbert-report

bartoszmajsak:
@dipak-pawar why are we ignoring the test now?

PR #1021, opened 2018-03-22
chore(#1003): updates documentation regarding permission conflicts.

### LINKED ISSUE #1003: Document different permissions available for users when trying to access different namespaces.
##### Issue Overview
Depending on the permissions available for different users like developer/admin, users can query/create a namespace.

For instance, when you try to run Kubernetes ftest as a developer, the ftests fail as developer user does not have sufficient access to create a temporary namespace. So, you either need to login as an admin user or set an existing/current namespace. 

##### Expected Behaviour

The permissions available to different users should be documented.

##### Current Behaviour

This is not documented anywhere and it gets confusing when and which properties should be set mandatorily and when they can we avoided.

##### Steps To Reproduce
1. oc login as developer 
2. execute kubernetes ftest

hemanik:

Documents necessary configuration required to avoid permission conflicts for unauthorized actions. 

Example: User `developer` trying to create a namespace.

```
io.fabric8.kubernetes.clnt.v3_1.KubernetesClientException: Failure executing: GET at: https://192.168.42.78:8443/api/v1/namespaces/itest-9adb622d. Message: Forbidden! User developer/192-168-42-78:8443 doesn't have permission. User "developer" cannot get namespaces in project "itest-9adb622d".
```

- Adds note in the documentation regarding the permissions conflict and how to avoid the same.

Fixes #1003


docs/kubernetes.adoc
```diff
@@ -240,6 +240,31 @@ a possible location of resources needed to be accessed during tests, usually wit
 you are testing made some changes in state of services in another namespaces, you can specify the secondary
 namespace in a field of @Named annotation to reach them.
 
+[NOTE]
+====
+Necessary configuration to avoid permission conflicts for creating namespaces.
+
+1. If using `MiniShift` to spin up a local Kubernetes cluster, login in as `admin` to create the default namespace or
```
dipak-pawar: MiniShift -> `Minishift`

docs/kubernetes.adoc
```diff
@@ -240,6 +240,31 @@ a possible location of resources needed to be accessed during tests, usually wit
...
+use `current` or `existing` namespace configured using properties `namespace.use.current` and `namespace.use.existing`
+respectively if logged in as any other user.
+
+2. In case of remote cluster use `current` or `existing` namespace for the user authenticated by token using property
+`cube.auth.token` or by username and password using properties `cube.username` and `cube.password` respectively.
+
```
dipak-pawar: Can we add about user can overwrite this property by setting system  property while running tests? So they can avoid putting `cube.username`, `cube.password`, `cube.auth.token` in `arquillian.xml`

docs/kubernetes.adoc
bartoszmajsak: Fixed with d34340cb5f2a10dc349c2c94cf445efbcc796b5d

PR #1044, opened 2018-04-17
Make it possible to generate image streams

goldmann:

It is now possible to use the `@OpenShiftDynamicImageStreamResource`
annotation to generate an image stream. This image stream will be added
to the freshly generated project and removed afterwards.

This makes it easy to specify which images should be used for the test.

#### Background

Issue with testing OpenShift images is that we need to be able to specify which image should be tested. If you put tests in a pipeline you don't know what the image will be, but once it is executed, we know that information.

By definition tests should self-contained and prepare everything that is required to run the test. This includes image streams too IMHO. This is why generating an image stream beforehand and deploying it outside of the test is not really the way to go.

We can generate the image stream definition and it would be picked by the ARQ cube, but it looks pretty hacky to me.

Other option I investigated is to add another annotation, similar to the general one `@OpenShiftResource`, which would let me point to an existing IS definition and the processors inside Cube would just replace the image with the image provided, and possibly specify the insecurity of the repository. I implemented it that way, but I did not submit that idea because it looked a bit hacky (parsing JSON and YAML and replacing things).

- Add `@OpenShiftDynamicImageStreamResource` annotation

### DIFF
--- a/openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/resources/OpenShiftResourceFactory.java
+++ b/openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/resources/OpenShiftResourceFactory.java
@@ -25,6 +25,7 @@ package org.arquillian.cube.openshift.impl.resources;
 
 import org.arquillian.cube.kubernetes.impl.resolver.ResourceResolver;
 import org.arquillian.cube.openshift.api.AddRoleToServiceAccount;
+import org.arquillian.cube.openshift.api.OpenShiftDynamicImageStreamResource;
 import org.arquillian.cube.openshift.api.OpenShiftResource;
 import org.arquillian.cube.openshift.api.RoleBinding;
 import org.arquillian.cube.openshift.api.Template;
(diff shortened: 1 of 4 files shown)


goldmann:
FYI @rcernich @spolti @jwendell

goldmann:
@bartoszmajsak Can I ask for review?

bartoszmajsak:
Summoning @lordofthejars. I'm on a medical leave right now.

openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/resources/OpenShiftResourceFactory.java
```diff
@@ -126,6 +131,57 @@ private static void createOpenShiftResources(String resourcesKey, OpenShiftAdapt
...
+            // Add image definition to image stream
+            JSONObject spec = new JSONObject();
+            spec.put("tags", tags);
+
+            imageStream.put("kind", "ImageStream");
+            imageStream.put("apiVersion", "v1");
+            imageStream.put("metadata", metadata);
+            imageStream.put("spec", spec);
+
+            log.info(String.format("Creating new OpenShift image stream resource: '%s'", imageStream.toJSONString()));
```
lordofthejars: Can you add the Json creation inside a method? Maybe called like `createImageStreamRequest` or sth similar.

goldmann:
@lordofthejars Requested changes applied, tests added.

lordofthejars:
Sorry to not mention before but I have noticed that there are no docs. I think that since this is a new feature would be interesting to add in documentation as well.

goldmann:
@lordofthejars Will add it.

goldmann:
@lordofthejars  Done!

PR #1054, opened 2018-05-18
OpenShiftAssistant : add a scale method with applicationName

### LINKED ISSUE #1052: OpenShiftAssistant : add a scale method with applicationName
In my test I need to scale down (0) some pods (like postgresql or other) to test the behavior of my application when infrastructure is in failure.

Currently, the scale method use the applicationName used to deploy an application in the OpenShiftAssistant.

So I have tried to inject multiple OpenShiftAssistant in my test (using @ArquillianResource) and deploy for each OpenShiftAssistant the application and next call on the desired one the scale method. However it doesn't worked because the same object is injected. So the applicationName used by the scale method refer to the last application launched.

A workaround is to define this method in test:

```
public void scale(final int replicas, final String applicationName) {
		final OpenShiftClient client = openShiftAssistant.getClient();
		final String namespace = client.getNamespace();
		client
			.deploymentConfigs()
			.inNamespace(namespace)
			.withName(applicationName)
			.scale(replicas);

		// ideally, we'd look at deployment config's status.availableReplicas field,
		// but that's only available since OpenShift 3.5
		await().atMost(5, TimeUnit.MINUTES)
			.until(() -> {
				final List<Pod> pods = client
						.pods()
						.inNamespace(namespace)
						.withLabel("deploymentconfig", applicationName)
						.list()
						.getItems();
				try {
					return pods.size() == replicas && pods.stream()
							.allMatch(Readiness::isPodReady);
				} catch (final IllegalStateException e) {
					// the 'Ready' condition can be missing sometimes, in which case Readiness.isPodReady throws an exception
					// here, we'll swallow that exception in hope that the 'Ready' condition will appear later
					return false;
				}
			});
	}

```

So I guess, the better way would be to add in OpenShiftAssistant and KubernetesAssistant this kind of method:

```
public void scale(final int replicas, final String applicationName) {
...
}
```

Regards,
Damien

dcdh:

I have got a Wildfly Swarm application using jaxrs, and JPA. My application is linked with a PostgreSQL database.

I want to test my application when the database is down (scale defined to 0) to ensure that my rest api will respond as expected (custom json response with custom response code).

However, using OpenShiftAssistant, is not currently possible because the scale method used the last deployed application name.

To run my application I need to run PostgreSQL first and then my Wildfly application to avoid it to crash on startup.

On this pull request I have added a scale method with applicationName and desired scale parameters on both OpenShiftAssistant and KubernetesAssistant.

I have mutualized some methods like awaitApplicationReadinessOrFail.

Regards,
Damien

Fixes #1052

### DIFF
--- a/kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/KubernetesAssistant.java
+++ b/kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/KubernetesAssistant.java
@@ -397,21 +397,16 @@ public class KubernetesAssistant {
     }
 
     /**
-     * Awaits at most 5 minutes until all pods are running.
+     * Awaits at most 5 minutes until all pods of the application are running.
      */
     public void awaitApplicationReadinessOrFail() {
         await().atMost(5, TimeUnit.MINUTES).until(() -> {
-                List<Pod> list = client.pods().inNamespace(namespace).list().getItems();
-                return list.stream()
-                    .filter(pod -> pod.getMetadata().getName().startsWith(applicationName) && !pod.getMetadata().getName().endsWith("-deploy"))
-                    .filter(this::isRunning)
-                    .collect(Collectors.toList()).size() >= 1;
+                return client
+                    .replicationControllers()
+                    .inNamespace(this.namespace)
+                    .withName(this.applicationName).isReady();
             }
         );
-    }
-
-    private boolean isRunning(Pod pod) {
-        return "running".equalsIgnoreCase(pod.getStatus().getPhase());
     }
 
     public String project() {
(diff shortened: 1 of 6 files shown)


dcdh:
About the conflict on KubernetesAssistant
in **awaitApplicationReadinessOrFail(final String applicationName)** it must be **.withName(applicationName).isReady();** because applicationName referds to the method variable and not the class variable.

lordofthejars:
Thank you so much, LGTM!! Can you resolve the conflict? Then when I get a green from Travis I will merge.

lordofthejars:
ASAP we merge it, we can do it.

dipak-pawar:
Thank you @dcdh for your contribution.

PR #1059, opened 2018-06-20
Close all cubes, even on error

AndyGee:

A call to stop is not guaranteed to complete:

Exception in thread "Thread-4" org.arquillian.cube.spi.CubeControlException: Could not stop vnc
	at org.arquillian.cube.spi.CubeControlException.failedStop(CubeControlException.java:23)
	at org.arquillian.cube.docker.impl.model.DockerCube.stop(DockerCube.java:169)
	at org.arquillian.cube.impl.client.ForceStopDockerContainersShutdownHook.lambda$attachShutDownHookForceStopDockerContainers$0(ForceStopDockerContainersShutdownHook.java:17)
	at java.lang.Thread.run(Thread.java:748)

This prevents the call to other cubes from completing.

Catch and log errors, but continue to process.

Fixes #


AndyGee:
Not sure how to squash those commits? The last one is relevant, and only changes one file.

core/src/main/java/org/arquillian/cube/impl/client/ForceStopDockerContainersShutdownHook.java
```diff
@@ -1,21 +1,35 @@
...
+            for (Cube cube: cubes) {
 
+                try {
                     cube.stop();
-                    cube.destroy();
+                } catch (final Throwable e) {
+                    Logger.getLogger(ForceStopDockerContainersShutdownHook.class.getName()).log(Level.WARNING, "Falied to stop: " + cube.getId(), e);
+                } finally {
+                    try {
+                        cube.destroy();
```
dipak-pawar: @AndyGee Suppose cube failed to stop container, then you can destroy it but with `-f` option, to remove it forcefully. We are not using `-f` option in cube while destroying it

Just curious to know why cube failed to stop container(`vnc` from exception)?

AndyGee: Sure, the issue is here that cube starts several containers in a test. If the first container fails to stop for 'any' reason then stop is never called on the following containers (due to the exception).

The secondary issue is that vnc was not started properly (I think that is another cube/drone issue), which is why the call to stop fails. 

This PR ensures that the 'other' containers are stopped and destroyed even if one fails. Throwing an error and stacktrace on the command line is not the expectation, and is not a test failure - Hence the logged warning.

core/src/main/java/org/arquillian/cube/impl/client/ForceStopDockerContainersShutdownHook.java
```diff
@@ -1,21 +1,35 @@
...
         Runtime.getRuntime().addShutdownHook(new Thread(() -> {
             final List<Cube<?>> cubes = cubeRegistry.getCubes();
-            for (Cube cube : cubes) {
+            for (Cube cube: cubes) {
 
+                try {
                     cube.stop();
-                    cube.destroy();
+                } catch (final Throwable e) {
+                    Logger.getLogger(ForceStopDockerContainersShutdownHook.class.getName()).log(Level.WARNING, "Falied to stop: " + cube.getId(), e);
```
dipak-pawar: how about declaring logger as `private static Logger logger = Logger.getLogger(ForceStopDockerContainersShutdownHook.class.getName());` and using it here.

Also there is typo for `Falied` -> `Failed`

bartoszmajsak:
> Not sure how to squash those commits? The last one is relevant, and only changes one file.

We do `squash and merge` from GH, so don't worry about that @AndyGee

core/src/main/java/org/arquillian/cube/impl/client/ForceStopDockerContainersShutdownHook.java
AndyGee: I'd only declare a logger static if I intend to use it more than once. Here it's only ever going to be used on an error that we don't expect to get often.

PR #1060, opened 2018-06-26
Update Istio support to 0.8.0

geoand:
The latest version is needed in order to create tests that work with the new networking resources in Istio 0.8


geoand:
@metacosm I presume that it's completely safe to bump versions of Istio Java API isn't it?

lordofthejars:
Yeah, you are right. Can you also update the ftests?

geoand:
@lordofthejars Sure thing! I'll do that tomorrow most likely.
I just wanted to see initially if the tests would pass before doing anything else :)

lordofthejars:
Cool, in theory they should pass since routerules alpha1 are still valid (but deprecated)

geoand:
Seems like things have broken on the Java API side, I'll need to take a good look tomorrow.

metacosm:
Indeed, it looks like the client uber-jar has changed. I'll take a look tomorrow.

geoand:
@metacosm @lordofthejars The Istio Java API requires version `3.2` of the Kubernetes Client, while Arquillian currently uses `3.1`.
Should we bump Arquillian to use that version as well or perhaps release a version of the Istio Java API that uses `3.1` instead of `3.2`?

geoand:
I bumped the Kubernetes Client version just to see if the tests will pass :)

metacosm:
The weird thing, though, is that Arquillian attempts to load the 3.2 version and failing, despite not needing it. I'm not too familiar with versionned uber-jars but my understanding is that they're meant to avoid exactly that kind of issue… Since Arquillian references the v3_1 version of the Kubernetes client, which I presume should come from the uber-jar, it should be failing to find *that* version, not the v3_2. Also, I think it's probably a bad idea to bundle dependencies in an uber-jar and export these dependencies to be used by clients because, indeed, this would mean that the client breaks each time the versioned bundled dependency changes. Perharps, that's the goal but then I don't see the benefit from using an uber-jar as opposed to having a direct dependency. Perhaps @lordofthejars could elaborate?

metacosm:
Different error but still failing…

metacosm:
Actually, the uber-jar is just versioned, it doesn't bundle the k8s client, so it's normal that Arquillian was complaining about not finding v3_2.

lordofthejars:
yes if version is bumped (uber ones) then imports needs to be updates as well. This is the drawback of uber jars.

geoand:
So I will continue this PR :)

First step is to get the existing tests working, then to update the ftests

geoand:
It seems like we can't move forward with this until a new model/client is released that contains [this](https://github.com/fabric8io/kubernetes-model/pull/322) fix.

geoand:
@lordofthejars @iocanel I updated this PR to use version `4.0.0` of the Kubernetes  clients, but it won't compile currently since [DefaultNamespaceService.cleanup](https://github.com/geoand/arquillian-cube/blob/5b7c3667149822e44cabbce2dad58c17fb8c0509/kubernetes/kubernetes/src/main/java/org/arquillian/cube/kubernetes/impl/namespace/DefaultNamespaceService.java#L165) uses `client.securityContextConstraints()` which has been moved from the `KubernetesClient` to the `OpenshiftClient`.
Is the solution in this case to simply remove the statement, or try to adapt the client to OpenshiftClient and try to perform the delete?

Thanks!

geoand:
We still need to release a new version of the Istio Java API before continuing work on this

lordofthejars:
@geoand I think that the delete should be done in both cases if the cluster is a kubernetes or openshift. The only difference is that when it is an openshift then we should adapt and call the `securityContextConstraints` method.

geoand:
@lordofthejars Thanks, that's what I ended up doing :)

The tests are failing however... I am checking if the failures are specific to this branch or if the same happens in master as well

geoand:
@lordofthejars @iocanel @metacosm The inclusion of the new Kubernetes client (4.0.0) means that when Arquillian attempts to deploy a resource file containing something like:

```
{
    "apiVersion" : "extensions/v1beta1",
    "kind" : "Deployment",
....
}
```

will fail and should instead be changed to:

```
{
    "apiVersion" : "extensions/v1beta1",
    "kind" : "Deployment",
....
}
```

Is this the desired behavior or should the former case be handled in some way?

iocanel:
@geoand: This shouldn't be a problem, as we introduced a compatibility layer that will fallback to old urls if a 404 is received.

metacosm:
I thought that `extensions` was still supported?

geoand:
I checked again and all tests are passing except the ones that involve the use of `extensions/v1beta1`.
So once the intereceptor in the Kubernetes Client has been updated, we should be good to go

geoand:
[Here](https://github.com/fabric8io/kubernetes-client/pull/1132) is the PR that fixes the issue mentioned above.

geoand:
Waiting for Maven Central to get Kubernetes Client the `4.0.1` release. Once that happens I will update the PR and hopefully everything will be green

lordofthejars:
@geoand Thank you very very much. Then now we have the update to 0.8.0 but using old resources in ftest right? So maybe after merging this issue we can open a new one for updating the ftest.

geoand:
@lordofthejars You are welcome :)

lordofthejars:
@geoand Did you open an issue?

geoand:
@lordofthejars Unfortunately no... Sorry about that, would have been nice for the release notes :(

lordofthejars:
No worries I open right now

geoand:
Thanks!

PR #1066, opened 2018-07-20
feat(#1065): provide an option to repeat @AwaitRoute several times

### LINKED ISSUE #1065: provide an option to repeat @AwaitRoute several times
##### Issue Overview

In some OpenShift clusters, particularly OpenShift Online Starter, it takes a while for the route to "stabilize" and during that timeframe, the route intermittently succeeds and fails. I'd like to enhance the `@AwaitRoute` mechanism to support a configurable amount of repetitions -- the route must respond successfully several times _in a row_ to be considered available. Default number of repetitions would be 1, i.e. current behavior.

##### Expected Behaviour

Be able to specify that the route must respond successfully e.g. 20 times in a row before it's considered available by the `@AwaitRoute` mechanism.

##### Current Behaviour

The `@AwaitRoute` mechanism only waits for the first successful response.

##### Steps To Reproduce
1. `oc login` into an OpenShift Online Starter cluster
2. `mvn clean verify -Popenshift,openshift-it` in a local clone of e.g. https://github.com/openshiftio-vertx-boosters/vertx-http-booster

##### Additional Information

I will be submitting an issue report to the OpenShift Online team as well.

Ladicek:
Fixes #1065.

### DIFF
--- a/docs/kubernetes.adoc
+++ b/docs/kubernetes.adoc
@@ -142,12 +142,14 @@ start
 | definitionsFile                     | String         | Any | Definitions file path                                         
 | proxiedContainerPorts               | List           | Any | Comma Separated List following Pod:containerPort OR 
 Pod:MappedPort:ContainerPort
-| enableImageStreamDetection          | Bool (true)    | Any | Enable detecting ImageStream resources located at target/XXX-is.(json|yaml)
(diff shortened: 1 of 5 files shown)


Ladicek:
Sure. I'll have to think about precedence (I'd probably like arquillian.xml/system property to override the annotation value), but that should be doable.

Ladicek:
I just pushed the change. Sorry it took me so long, I was on vacation :-)

Issue #1090, opened 2018-10-11
Update to latest istio-java


lordofthejars:
Update to latest istio-java 1.0.0.beta.2

PR #1091, opened 2018-10-11
chore(#1090): Updates to Istio 1.0.0.Beta2

lordofthejars:


kubernetes/istio/istio/src/main/java/org/arquillian/cube/istio/impl/IstioClientCreator.java
```diff
@@ -16,9 +16,7 @@
 
     public void createIstioClient(@Observes final KubernetesClient client) {
 
-        KubernetesAdapter adapter = new KubernetesAdapter(client);
-        IstioClient istioClient = new IstioClient(adapter);
-
+        IstioClient istioClient = new DefaultIstioClient(client.getConfiguration());
```
iocanel: While this is perfectly fine, It's slightly better to do it via:

     client.adapt(IstioClient.class);

With this approach you decouple the code from the actual impl and also you use a clone of the http client instance instead of creating one from config.

PR #1099, opened 2018-10-24
#1098 VideoConversionLifecycleManager & VncRecorderLifecycleManager

### LINKED ISSUE #1098: VideoConversionLifecycleManager & VncRecorderLifecycleManager should not fail the suite
##### Issue Overview

VideoConversionLifecycleManager & VncRecorderLifecycleManager both use `@Observes` for lifecycle.

As `@Observes` events are handled directly in the event chain then extra care (try/catch) needs to be taken on event methods that may fail. Other observers must still be called even on an error, else cleanups are not run for example.

Generally, all `@Observes` methods should be safe to call (try/catch), with their own error handling, else failures break the event chain.

##### Expected Behaviour

At least an error message on failure, but not the failure of the test suite.

##### Current Behaviour

Any errors in `@Observes` methods cause the test suite to fail, eg:

`
org.arquillian.cube.spi.CubeControlException: Could not create flv2mp4
	at org.arquillian.cube.spi.CubeControlException.failedCreate(CubeControlException.java:15)
	at org.arquillian.cube.docker.impl.model.DockerCube.create(DockerCube.java:115)
	at org.arquillian.cube.docker.drone.VideoConversionLifecycleManager.startConversion(VideoConversionLifec

.....

Caused by: com.github.dockerjava.api.exception.InternalServerErrorException: {"message":"Get https://registry-1.docker.io/v2/: dial tcp: lookup registry-1.docker.io on 192.168.65.1:53: no such host"}
`

The above example is due to the fact the image is blocked by policy, but it should not break the suite.

AndyGee:

Catch and log errors in `@Observes` methods, but do not fail the suite and allow cleanups - This should actually be the default for any `@Observes` method, as the CDI events are synchronous - A failure prevents other observers from running, such as cleanups etc.

Also, the Cube internal images might need to be pulled from a user specified docker registry. So this PR also adds and new "dockerRegistry" property - I did think about using the _docker_ qualifier, but using _cubedrone_ allows for a different registry if required (actually have a use case for this). 

- Added try/catch/log to `@Observes` methods in above classes
- VolumeCreator.java should use a better path
- Added "dockerRegistry" property & documentation

Fixes #1098

### DIFF
--- a/docker/drone/src/test/java/org/arquillian/cube/docker/drone/SeleniumContainersTest.java
+++ b/docker/drone/src/test/java/org/arquillian/cube/docker/drone/SeleniumContainersTest.java
@@ -9,6 +9,8 @@ import org.junit.runner.RunWith;
 import org.mockito.Mock;
 import org.mockito.runners.MockitoJUnitRunner;
 
+import java.util.HashMap;
+
 import static org.hamcrest.CoreMatchers.both;
 import static org.hamcrest.CoreMatchers.hasItem;
 import static org.hamcrest.CoreMatchers.is;
@@ -95,6 +97,7 @@ public class SeleniumContainersTest {
         when(cubeDroneConfiguration.isBrowserDockerfileDirectorySet()).thenReturn(false);
         when(cubeDroneConfiguration.isBrowserImageSet()).thenReturn(false);
         when(cubeDroneConfiguration.getContainerNameStrategy()).thenReturn(ContainerNameStrategy.STATIC);
+        when(cubeDroneConfiguration.getDockerRegistry()).thenReturn("");
 
         final SeleniumContainers firefox = SeleniumContainers.create("firefox", cubeDroneConfiguration);
         assertThat(firefox.getBrowser(), is("firefox"));
@@ -109,6 +112,7 @@ public class SeleniumContainersTest {
         when(cubeDroneConfiguration.isBrowserDockerfileDirectorySet()).thenReturn(false);
         when(cubeDroneConfiguration.isBrowserImageSet()).thenReturn(false);
         when(cubeDroneConfiguration.getContainerNameStrategy()).thenReturn(ContainerNameStrategy.STATIC);
+        when(cubeDroneConfiguration.getDockerRegistry()).thenReturn("");
 
         final SeleniumContainers firefox = SeleniumContainers.create("chrome", cubeDroneConfiguration);
         assertThat(firefox.getBrowser(), is("chrome"));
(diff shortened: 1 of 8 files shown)


docs/drone.adoc
```diff
@@ -47,6 +47,7 @@ browserImage:: Docker image to be used as custom browser image instead of the of
 browserDockerfileLocation:: Dockerfile location to be used to built custom docker image instead of the official one. This property has preference over browserImage.
 containerNameStrategy:: Sets the strategy for generating the container name. Valid values are `STATIC, STATIC_PREFIX, RANDOM` with the default being `STATIC`. `STATIC` will always use the same name, `STATIC_PREFIX` lets you define a prefix and enables running different tests in parallel and `RANDOM` generates a unique container name on every test run enabling running the same test in parallel.
 containerNamePrefix:: The prefix for the `STATIC_PREFIX` container name strategy.
+dockerRegistry:: Use this to override the default docker registry with a user specified registry - Don't forget the trailing slash (i.e. _your.repo**/**_)
```
dipak-pawar: How about checking `dockerRegistry` defined in configuration and making sure it has trailing slash, if not then add it. There are a lot of chances user will miss this documentation. It'll be good if we can make cube more smart.

AndyGee: Done

docker/drone/src/test/java/org/arquillian/cube/docker/drone/SeleniumContainersTest.java
```diff
@@ -156,4 +160,44 @@ public void shouldCreateStaticContainerNameAndPort() {
...
+
+        final CubeDroneConfiguration conf = CubeDroneConfiguration.fromMap( new HashMap<String, String>() {{
+            put("containerNameStrategy",ContainerNameStrategy.STATIC_PREFIX.toString());
+            put("dockerRegistry","my.registry/\r\n ");
+        }});
+
+        checkRegistry(conf);
+    }
+
+    private void checkRegistry(CubeDroneConfiguration conf) {
```
dipak-pawar: How about `verifyRegistry`?

docs/drone.adoc
dipak-pawar: Do we need comment  `Don't forget the trailing slash (i.e. _your.repo**/**_)` now?

docker/drone/src/test/java/org/arquillian/cube/docker/drone/SeleniumContainersTest.java
AndyGee: Its private, don't see any value

docker/ftest-container-star-operator/pom.xml
```diff
@@ -68,6 +68,12 @@
       <artifactId>wildfly-arquillian-container-remote</artifactId>
       <version>${version.wildfly}</version>
       <scope>test</scope>
+      <exclusions>
```
AndyGee: This is not required, and causes a build failure on OpenJDK

dipak-pawar: I didn't get your point?

AndyGee: The exclusion

AndyGee: this = jconsole is not required

AndyGee: jconsole = causes a build failure on OpenJDK

AndyGee: So exluding it means that the build passes on OpenJDK

PR #1111, opened 2019-03-13
Execute commands after container start

tobiasstadler:
I want to copy files and execute commands after a container started (similar to the before stop commands)

Implemented AfterStart observer similar to BeforeStopObserver

Fixes #


bartoszmajsak:
@lordofthejars can you have a look?

lordofthejars:
Thank you very much for improving Arquillian Cube.

PR #1113, opened 2019-03-29
#1107 Allow override of selenium containers and use real DinD info

### LINKED ISSUE #1107: Enhancement - Allow override of selenium containers
Enhancement - Allow override of selenium containers. Currently, they're added after the cubeSpecificProperties are evaluated and applied. And also determine the real container hostname/IP for a cube.

I am working on a patch, so this ticket is a heads up and allows me to create a branch name. 

##### Issue Overview

The browser is not on DOCKER_HOST:14444 (hence dockerHost internally) in a CICD env.

When running in a CICD env it is common that the 'runners' are DIND (Docker in Docker) - This means the docker daemon is actually on the host, and the runner is mapped/mounted to use the hosts daemon.

When something like `docker run -p 14444:4444 ...` is executed, then the image is actually started on the host, and not in the runner. This means that the hostname/IP of the running image is NOT the host or the runner, ie. it gets its own name/ip. In this case, DOCKER_HOST is not useful.

$DOCKER_HOST:14444 <- This is not where the image service is available

172.18.0.4:14444 <- This is where the browser 'might' be.

We can inspect information from a container:

`docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' container_name_or_id`, so I'm guessing right now that the API will also be able to help us here to get the container IP dynamically. Another way is to configure the docker network and override the configuration.

The URL for the selenium server is then going to be something like `http://containerIP:14444` - Eiter way, we need to set it dynamically or allow the cubeSpecificProperties to define overrides for the await strategy and webdriver connection etc..

##### Expected Behaviour

Await polls the 'real' browser host
The WebDriver (and other selenium containers) use the 'real' container host

##### Current Behaviour

The internal dockerHost variable is evaluated to DOCKER_HOST, which is not where the container is available.

##### Steps To Reproduce
1. Run the 'browser' container on a remote machine
2. Cube tries to call http://dockerHost:14444

##### Additional Information

I'll be adding my findings here as I work on the patch.

AndyGee:

Adds better Docker in Docker (DinD) capability to the platform and allows enrichment of the static context of a test class. The selenium containers can also be overridden or modified.

- Static context enrichment
- DinD resolution of containers that cannot be referenced by dockerhost, hostname or localhost.
- Enhanced Await strategies for DinD
- Overriding of Selenium container attributes using cubeSpecificProperties
- Docker network cleanup on failures
- Tests for new features
- Fix breaking tests on Win OS
- Multiple tweaks along the way

Fixes #1107


AndyGee:
This is too big even for my liking - I'll close it and make a few smaller requests.

PR #1149, opened 2019-10-03
fix: avoid issuing invalid wait-for-it command

mstruk:
The default wait-for-it script is easily invoked without a HOST, because internalIP easily resolves to empty string.

An example invocation which can never succeed because it's syntactically incorrect:

    /tmp/wait-for-it.sh :8080 -s -- echo Service is Up

We can go an extra step, and make sure that a non-empty HOST is set. we may not default to a good value but at least it's one that sometimes works.

- if internalIP is empty, default to containerIP, if that one is also empty default to localhost


bartoszmajsak:
This LGTM but would like to hear an opinion from @lordofthejars.

Many thanks @mstruk for your contribution!

PR #1150, opened 2019-10-03
feat: adds docker-compose support for environment maps

mstruk:
This patch allows using the following syntax in docker-compose:
- Decimal value for version (e.g. version: '3.5')
- List for entrypoint (e.g. entrypoint: [], or entrypoint: [/bin/sh, -c, echo]
- Map for environment


bartoszmajsak:
This LGTM but would like to hear an opinion from @lordofthejars.

Many thanks @mstruk for your contribution!

PR #1156, opened 2019-10-22
Update OkHttp to 4.2.2

tobiasstadler:


bartoszmajsak:
@tobiasstadler I am a bit reluctant with upgrading to 4.x as it's been rewritten in Kotlin and thus became a significantly bigger dependency (almost double the size on its own). Do we have a strong reason to do that update?

tobiasstadler:
no, no strong reason

bartoszmajsak:
Thanks, let's keep it in 3.x unless @lordofthejars has other opinion

PR #1179, opened 2020-05-16
fix(#1178): Calculate and specify a wait-for-it timeout

### LINKED ISSUE #1178: PollingAwaitStrategy wait-for-it does fallback ignores parameters
##### Issue Overview

When using the polling await strategy on a slower machine with container images lacking the ss command, the strategy falls back to using wait-for-it script, but doesn't specify a timeout.  On slower machines autostarting multiple containers in parallel, the default timeout of 15 seconds isn't long enough.

##### Expected Behaviour

The polling await strategy should use the await parameters to specify a timeout to the wait-for-it script so that await times can be adjusted

##### Current Behaviour

Sometimes the containers start successfully, but on slower machines with multiple containers autostarting, container startups fail.

phillipross:

When a container is configured with polling await strategy, the parameters for the strategy are used to compute a timeout and passed to the wait-for-it command.

Fixes #1178


bartoszmajsak:
Thanks for your contribution @phillipross. I guess I have to first fix the build (general problem, not PR related) and then we can move forward.

phillipross:
Yes, I had to disable OpenShift tests to get my builds to work.  I don't have the expertise with OpenShift to determine the problem, but figured I should submit the PR anyway.  Thanks!

phillipross:
@bartoszmajsak any progress on this?  let me know if there's anything I can do to help move things alone.  Thanks!

bartoszmajsak:
> @bartoszmajsak any progress on this? let me know if there's anything I can do to help move things alone. Thanks!

Sorry @phillipross I didn't have time yet to look at build failures. If you have time and you are willing to dig into it that would be just awesome. I might be able to get into that later this week, but been pretty busy with other stuff lately :(

phillipross:
I understand, no problem 😉  I believe when I looked before it seemed to be issues with the openshift integration or something, but I'll take a closer look and see what I can uncover.

Issue #1270, opened 2023-11-22
[TASK]:Build with JDK11.


jimma:
This needs upgrade the jboss-parent to the latest one and bump other dependencies version.

rsearls:
Please assign this to me.  I am starting work on it now.

jimma:
@rsearls  We still didn't find the people who can assign issue.   Please directly start to work on this issue.

rsearls:
Thanks.  Started it yesterday after posting my comment in the issue.

gaol:
@rsearls With some discussions on https://github.com/arquillian/arquillian-cube/pull/1276, I think it will be great to address the following issues too with this issue:
* Enable JDK 11 in the github action workflow (if it gets merged before this one :) )
* [A sun.net.spi.nameservice.NameService implementation in openshift/openshift module](https://github.com/arquillian/arquillian-cube/blob/c7025efb2bda2f111af5864f2196255eff85d720/openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/dns/ArqCubeNameServiceDescriptor.java) should not work anymore with upgrading to JDK 11, as the [NameService got dropped from JDK9+](https://bugs.openjdk.org/browse/JDK-8134577),  we need to decide how to do with it.

Thanks :)

rsearls:
@gaol I recommend the changes for those 2 issues be submitted under separate PRs after code for this PR had been merged, so if anything needs to be backed out, it can be easily done.
 
In addressing this JDK-11 compile issue nearly all dependencies have had to be upgraded and many class files required corresponding changes.  I believe the better strategy is to get clean compiling code and use that as the base to refine the project moving forward.

fabiobrz:
This can be closed as resolved since the [only outstanding issue](https://github.com/arquillian/arquillian-cube/issues/1287) is not a  blocker here, see https://github.com/arquillian/arquillian-cube/issues/1287#issuecomment-2611155885

@rsearls, @gaol feel free to push back in case you'd have any objections.

PR #1276, opened 2023-11-26
[Issue:1272] Add github action workflow yml to build

### LINKED ISSUE #1272: [TASK]:Add github action workflow yml to build

gaol:
Fixes #1272

I would like to propose the following changes in this PR:

* based on Jim's initial github action setup from #1274
* Replace docker image of `jonmorehouse/ping-pong` to `hashicorp/http-echo`, because the former one is missing from docker hub, it will fail the test in a clean env.
* Specify `sun.net.spi.nameservice.provider` configure in the `maven surefile plugin` configuration instead of set it in the test code, because it would be too late if other tests have load the InetAddress already
* Fixes test in `fabric8-maven-plugin-build` to check if the namespece is set correctly, that depends on the current environment setup.
* Disable jdk 11 in the github action for the moment, it should be enabled again after issue #1270 got fixed
* Disable minikube env in the github action for the moment, as there are test failures, not sure if the upgrade of fabric8 kubernetes client can solve it. We need to address those tests when the test failures got fixed in minikube environment.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -1,0 +1,60 @@
+name: Arquillian-Cube Maven Build
+
+on:
+  push:
+    branches:
+      - '**'
+  pull_request:
+    branches:
+      - '**'
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 90
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+        java: ['8','11']
+
+    steps:
+      - name: Setup Minikube
+        uses: manusa/actions-setup-minikube@v2.9.0
+        with:
+          driver: docker
+          container runtime: containerd
+          minikube version: 'v1.30.1'
+          kubernetes version: 'v1.27.1'
+          github token: ${{ secrets.GITHUB_TOKEN }}
+          start args: "--memory='4gb' --cpus='2'"
+      - name: Enable minikube registry
+        run: |
+          minikube addons enable registry
+          kubectl port-forward --namespace kube-system service/registry 5000:80 &
+      - name: Checkout arquillian-cube
+        uses: actions/checkout@v4
+      - uses: actions/cache@v3
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+      - name: Set up JDK ${{ matrix.java }}
+        uses: actions/setup-java@v3
+        with:
+          java-version: ${{ matrix.java }}
+          distribution: 'adopt'
+      - name: Build with Maven Java ${{ matrix.java }}
+        run: |
+          mvn -fae clean install
+      - uses: actions/upload-artifact@v3
+        if: failure()
+        with:
+          name: surefire-reports-${{ matrix.os }}-${{ matrix.java }}
+          path: '${{ github.workspace }}/**/surefire-reports/*.*'
+      - uses: actions/upload-artifact@v3
+        if: failure()
+        with:
+          name: server-logs-${{ matrix.os }}-${{ matrix.java }}
+          path: '${{ github.workspace }}/**/*.log'
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -3,10 +3,17 @@ name: Arquillian-Cube Maven Build
 on:
   push:
     branches:
-      - '**'
+      - main
   pull_request:
     branches:
-      - '**'
+      - main
+    paths-ignore:
+      - 'doc/**'
+      - '.circleci/**'
+      - '.travis/**'
+env:
+  # failsafe.groups configuration depends on the env setup.
+  FAILSAFE_GROUPS: ''
 
 jobs:
   build:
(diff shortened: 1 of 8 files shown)


gaol:
@jimma @fabiobrz may I ask you to review ? thanks

jimma:
This looks good to me.

.github/workflows/maven.yml
```diff
@@ -0,0 +1,68 @@
+name: Arquillian-Cube Maven Build
+
+on:
+  push:
+    branches:
+      - '**'
```
fabiobrz, line 6: Shouldn't we trigger just on `master`?

.github/workflows/maven.yml
```diff
@@ -0,0 +1,68 @@
...
+  pull_request:
+    branches:
+      - '**'
```
fabiobrz, line 9: Have we considered skipping the checks if the PR is just about e.g.: docs? See for instance a WIP [here](https://github.com/Intersmash/intersmash/pull/66/files#diff-d978a3dadf11b38be2ec45facd9d1f5a3785546602451eeb09c2c3cf20a104aaR9)

.github/workflows/maven.yml
```diff
@@ -0,0 +1,68 @@
...
+    steps:
+      - name: Setup Minikube
+        uses: manusa/actions-setup-minikube@v2.9.0
+        with:
+          driver: docker
+          container runtime: containerd
+          minikube version: 'v1.30.1'
+          kubernetes version: 'v1.27.1'
+          github token: ${{ secrets.GITHUB_TOKEN }}
+          start args: "--memory='4gb' --cpus='2'"
```
fabiobrz, line 33: Any specific requirement that we might want to document with a comment here?

openshift/openshift/pom.xml
```diff
@@ -220,4 +220,19 @@
...
   </dependencies>
 
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <systemPropertyVariables>
+            <sun.net.spi.nameservice.provider.1>dns,ArquillianCubeNameService</sun.net.spi.nameservice.provider.1>
```
fabiobrz, line 230: Sorry if I am missing something but where is `ArquillianCubeNameService` taken from?

fabiobrz (commented):
Hi @gaol - and thanks for this PR. I dropped a few question about the workflow configuration. Feel free to let me know what you think

openshift/openshift/pom.xml
gaol: It comes from [ArqCubeNameServiceDescriptor implementation in openshift/openshift module](https://github.com/arquillian/arquillian-cube/blob/c7025efb2bda2f111af5864f2196255eff85d720/openshift/openshift/src/main/java/org/arquillian/cube/openshift/impl/dns/ArqCubeNameServiceDescriptor.java#L23).

.github/workflows/maven.yml
gaol: done

.github/workflows/maven.yml
gaol: done

.github/workflows/maven.yml
jimma: I am thinking if we need to create another branch(main branch?) to put all new updates, changes there.

.github/workflows/maven.yml
jimma: Do you mean the args to start the minikube ?   By default minikube is started with `2 cpus` and `2gb memory` and it's not enough for some heavy project, so I explicitly added these arguments to minikube.

fabiobrz: Perfect, thanks.

.github/workflows/maven.yml
fabiobrz: +1, we should rename to `main` but I think we can handle that issue separately, and maybe more in general.

openshift/openshift/pom.xml
fabiobrz: Ok, thanks @gaol. Now I am wondering now whether we should keep setting it in the [test class](https://github.com/arquillian/arquillian-cube/blob/c7025efb2bda2f111af5864f2196255eff85d720/openshift/openshift/src/test/java/org/arquillian/cube/openshift/impl/DnsServiceTest.java#L23). BTW - it is documented as well. Should we update and just keep passing the properties to the surefire plugin execution via `mvn`? WDYT?

gaol: I think it is fine to leave it settings in the test classes, because it will make the test pass when you trying to run it in the IDE without having additional setup, but I am also good to delete the settings in the test class. :)

And the [NameService has been dropped from JDK9](https://bugs.openjdk.org/browse/JDK-8134577), so we may need to think about how to do with this when https://github.com/arquillian/arquillian-cube/issues/1270 needs to be fixed.

fabiobrz: +1, I'd say we can leave it as is now, i.e. accept your changes and resolve this thread.

It would be useful to track the future changes with an issue.

fabiobrz (approved):
LGTM, thanks @gaol and @jimma !

gaol:
As the default branch will be renamed to `main`, I am closing this one, suppressed by #1304

PR #1304, opened 2024-04-23
GitHub action setup

### LINKED ISSUE #1272: [TASK]:Add github action workflow yml to build

gaol:
Fixes #1272

* Based on Jim's initial github action setup from Initial change to start the github action build  #1274
* Check changes to `main` branch only and ignore changes in non code paths
* Check unit tests only for now. `./mvnw clean package`

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -1,0 +1,60 @@
+name: Arquillian-Cube Maven Build
+
+on:
+  push:
+    branches:
+      - '**'
+  pull_request:
+    branches:
+      - '**'
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 90
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+        java: ['8','11']
+
+    steps:
+      - name: Setup Minikube
+        uses: manusa/actions-setup-minikube@v2.9.0
+        with:
+          driver: docker
+          container runtime: containerd
+          minikube version: 'v1.30.1'
+          kubernetes version: 'v1.27.1'
+          github token: ${{ secrets.GITHUB_TOKEN }}
+          start args: "--memory='4gb' --cpus='2'"
+      - name: Enable minikube registry
+        run: |
+          minikube addons enable registry
+          kubectl port-forward --namespace kube-system service/registry 5000:80 &
+      - name: Checkout arquillian-cube
+        uses: actions/checkout@v4
+      - uses: actions/cache@v3
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+      - name: Set up JDK ${{ matrix.java }}
+        uses: actions/setup-java@v3
+        with:
+          java-version: ${{ matrix.java }}
+          distribution: 'adopt'
+      - name: Build with Maven Java ${{ matrix.java }}
+        run: |
+          mvn -fae clean install
+      - uses: actions/upload-artifact@v3
+        if: failure()
+        with:
+          name: surefire-reports-${{ matrix.os }}-${{ matrix.java }}
+          path: '${{ github.workspace }}/**/surefire-reports/*.*'
+      - uses: actions/upload-artifact@v3
+        if: failure()
+        with:
+          name: server-logs-${{ matrix.os }}-${{ matrix.java }}
+          path: '${{ github.workspace }}/**/*.log'
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -3,10 +3,17 @@ name: Arquillian-Cube Maven Build
 on:
   push:
     branches:
-      - '**'
+      - main
   pull_request:
     branches:
-      - '**'
+      - main
+    paths-ignore:
+      - 'doc/**'
+      - '.circleci/**'
+      - '.travis/**'
+env:
+  # failsafe.groups configuration depends on the env setup.
+  FAILSAFE_GROUPS: ''
 
 jobs:
   build:
@@ -16,7 +23,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        java: ['8','11']
+        java: ['11']
 
     steps:
       - name: Setup Minikube
@@ -28,6 +35,8 @@ jobs:
           kubernetes version: 'v1.27.1'
           github token: ${{ secrets.GITHUB_TOKEN }}
           start args: "--memory='4gb' --cpus='2'"
+        env:
+          FAILSAFE_GROUPS: "-Dfailsafe.groups=org.arquillian.cube.docker.impl.requirement.RequiresDocker"
       - name: Enable minikube registry
         run: |
           minikube addons enable registry
@@ -45,9 +54,12 @@ jobs:
         with:
           java-version: ${{ matrix.java }}
           distribution: 'adopt'
+      - name: Maven pre-fetch dependencies ${{ matrix.java }}
+        run: |
+          ./mvnw verify -q -U -DskipTests # pre-fetch dependencies
       - name: Build with Maven Java ${{ matrix.java }}
         run: |
-          mvn -fae clean install
+          ./mvnw -fae clean package ${{ env.FAILSAFE_GROUPS }}
       - uses: actions/upload-artifact@v3
         if: failure()
         with:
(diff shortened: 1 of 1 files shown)


gaol:
@jimma @fabiobrz may I ask you to review ? thanks ! This will enable github action just like what .circleci does now for the unit tests. The integration tests will be fixed by following up issues, and once they are done, the github action will be updated to cover integration tests as well.

.github/workflows/maven.yml
```diff
@@ -3,10 +3,18 @@ name: Arquillian-Cube Maven Build
...
-      - '**'
+      - main
   pull_request:
     branches:
-      - '**'
+      - main
+    paths-ignore:
+      - 'doc/**'
+      - '.circleci/**'
+      - '.travis/**'
```
jimma, line 13: Should we remove the  travis setup ?

gaol: done

.github/workflows/maven.yml
```diff
@@ -0,0 +1,75 @@
...
+  DOCKER_HOST: 'unix:///var/run/docker.sock'
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 90
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
```
rhusar, line 25: I am not sure we need to create a platform matrix as this will require linux containers, so this won't ever work on windows and probably neither on macos.

rhusar: i.e. I would just do `runs-on: ubuntu-latest` and move on

.github/workflows/maven.yml
```diff
@@ -0,0 +1,75 @@
...
+        java: ['11']
+
+    steps:
+      - name: Setup Minikube
+        uses: manusa/actions-setup-minikube@v2.9.0
+        with:
+          driver: docker
+          container runtime: containerd
+          minikube version: 'v1.30.1'
+          kubernetes version: 'v1.27.1'
```
fabiobrz, line 35: Just asking whether we'd want to use the latest K8s minor, or at least micro? See https://kubernetes.io/releases/

Or is there a reason for referencing `1.27.1`?

.github/workflows/maven.yml
gaol: got it !

fabiobrz: I guess `matrix` aims at supporting more than just one JDK, as it was in the initial PR: https://github.com/arquillian/arquillian-cube/pull/1274/files#diff-5dbf1a803ecc13ff945a08ed3eb09149a83615e83f15320550af8e3a91976446R19

But just my take.
A question, though: wouldn't be better to have them working on linux and file a separate tracker for expanding? Short living (hopefully) and easy to review PRs would be the reason.
WDYT?

fabiobrz (commented):
Hi @gaol - and thanks for this PR.
I dropped some minor comments, and I'd like to know whether we do have some run reference for the action we are implementing: maybe here? https://github.com/gaol/arquillian-cube/actions

WDYT?

.github/workflows/maven.yml
gaol: Sure, we can use the latest version I guess. It was there from original PR. will update it to `1.30.0` which is the latest one I can see. also I will update the `actions-setup-minikube` to latest one to support it.

gaol:
> I'd like to know whether we do have some run reference for the action we are implementing: maybe here? https://github.com/gaol/arquillian-cube/actions
> 
> WDYT?

Yes, there is: https://github.com/gaol/arquillian-cube/tree/github_action_setup_own_test, which is the place I test the github action execution.

> wouldn't be better to have them working on linux and file a separate tracker for expanding? Short living (hopefully) and easy to review PRs would be the reason.

Sure, now the plan is to stay with `runs-on: ubuntu-latest`, and having the jdk matrix to `11` only, we can expand to more jdk versions later.

fabiobrz:
> the github action execution.



> > I'd like to know whether we do have some run reference for the action we are implementing: maybe here? https://github.com/gaol/arquillian-cube/actions
> > WDYT?
> 
> Yes, there is: https://github.com/gaol/arquillian-cube/tree/github_action_setup_own_test, which is the place I test the github action execution.
> 
> > wouldn't be better to have them working on linux and file a separate tracker for expanding? Short living (hopefully) and easy to review PRs would be the reason.
> 
> Sure, now the plan is to stay with `runs-on: ubuntu-latest`, and having the jdk matrix to `11` only, we can expand to more jdk versions later.

Thanks @gaol - Ok, so we'd have to wait for a green bar there, right? IIUC the last commit is failing the CI checks: https://github.com/gaol/arquillian-cube/actions/runs/8814197921/job/24193526839

gaol:
yes, @fabiobrz , I think there is a temporary failure (failed with `apt-get update -y`), will re-run it later and comment here once I get green result.

.github/workflows/maven.yml
rhusar: JDK matrix for sure - but os matrix not really useful.

gaol:
Updated according to the feed backs:
* only use runner: `ubuntu-latest`
* jdk matrix is `11` for now
* minikube version: `v1.32.0`  for `manusa/actions-setup-minikube@v2.10.0`
* kubernetes version: `v1.30.0` for `manusa/actions-setup-minikube@v2.10.0`


The reference github action execution is: https://github.com/gaol/arquillian-cube/actions/runs/8818141575/job/24206240475

@fabiobrz @rhusar

fabiobrz (dismissed):
Unit tests in the K8s category are executed and passing along with the Docker ones. I am approving, let's expand to ITs and OpenShift from here. Thanks @gaol !

I'll merge this once @rhusar will approve too.

.github/workflows/maven.yml
```diff
@@ -0,0 +1,74 @@
+name: Arquillian-Cube Maven Build
```
rhusar, line 1: Please drop the dash, the project name is "Arquillian Cube" (no dash).

Never the less I always struggle how to name these. The project name seems redundant. The maven build is also redundant, since that's the only build system we use here. Also, build is a bit of a misnomer as primarily this is used to testing...

.github/workflows/maven.yml
```diff
@@ -0,0 +1,74 @@
...
+          ./mvnw -fae clean package ${{ env.FAILSAFE_GROUPS }}
+      - uses: actions/upload-artifact@v3
+        if: failure()
+        with:
+          name: surefire-reports-${{ matrix.os }}-${{ matrix.java }}
+          path: '${{ github.workspace }}/**/surefire-reports/*.*'
+      - uses: actions/upload-artifact@v3
+        if: failure()
+        with:
+          name: server-logs-${{ matrix.os }}-${{ matrix.java }}
```
rhusar, line 70: The ${{ matrix.os }} would fail to resolve now.

.github/workflows/maven.yml
```diff
@@ -0,0 +1,74 @@
...
+      - name: Maven pre-fetch dependencies ${{ matrix.java }}
+        run: |
+          ./mvnw verify -q -U -DskipTests # pre-fetch dependencies
+      - name: Build with Maven Java ${{ matrix.java }}
+        run: |
+          ./mvnw -fae clean package ${{ env.FAILSAFE_GROUPS }}
+      - uses: actions/upload-artifact@v3
+        if: failure()
+        with:
+          name: surefire-reports-${{ matrix.os }}-${{ matrix.java }}
```
rhusar, line 65: The ${{ matrix.os }} would fail to resolve now.

.github/workflows/maven.yml
rhusar, line 70: Also any reason why this action is duplicated? I would just collect this in one zip with multiple paths.

rhusar (changes requested):
Few more things.

.github/workflows/maven.yml
gaol: right !

gaol: They are different artifacts, one is all from `target/urefire-reports/`, the other is from all `*.log` files, the 2 artifacts separate the information. but I think it is fine as well to zip all into one artifact.

gaol:
@rhusar thanks, fixed as the following:
* Rename the workflow name to `Arquillian Cube Test`, please feel free to propose new names if you like.
* Combine all log files and surefire-reports files into one artifact to upload on failure.
* Drop the use of `matrix.os`

The last execution is: https://github.com/gaol/arquillian-cube/actions/runs/8826949457/job/24233479032

rhusar (approved):
Looks good to me, thanks!

fabiobrz (approved):

PR #1314, opened 2024-09-23
Resume K8s integration tests

### LINKED ISSUE #1313: Make K8s integration tests green again, and resume the related CI execution
##### Issue Overview

Resume K8s integration tests.

##### Expected Behaviour

The tests that verify the Arquillian Cube Kubernetes integration should pass, and the CI script should be fixed in order to resume the execution of such tests.

##### Current Behaviour

The Arquillian Cube integration tests for Kubernetes are failing, and the CI doesn't execute them, see https://github.com/arquillian/arquillian-cube/blob/main/.circleci/config.yml#L26 and https://github.com/arquillian/arquillian-cube/blob/main/.circleci/config.yml#L77

##### Steps To Reproduce
1. 
```
mvn clean verify -Dfailsafe.groups=org.arquillian.cube.kubernetes.impl.requirement.RequiresKubernetes
``` 

##### Additional Information

N/A

fabiobrz:
Changes aim at resuming Arquillian Cube tests for Kubernetes integration.

- Fixing Kubernetes integration implementation
- Fixing test resources and tests
- Removing the CircleCI script which does nothing relevant to actual integration tests, and improving the GitHub actions script n order to run a matrix on the docker and kubernetes targets.

Fixes #1313

### DIFF
--- a/kubernetes/ftest-kubernetes-logs/pom.xml
+++ b/kubernetes/ftest-kubernetes-logs/pom.xml
@@ -25,6 +25,11 @@
       <artifactId>junit</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.fabric8</groupId>
+      <artifactId>kubernetes-client</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -11,8 +11,6 @@ on:
       - 'doc/**'
       - '.circleci/**'
 env:
-  # failsafe.groups configuration depends on the env setup.
-  FAILSAFE_GROUPS: ''
   DOCKER_HOST: 'unix:///var/run/docker.sock'
 
 jobs:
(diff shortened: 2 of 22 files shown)


kubernetes/ftest-kubernetes-logs/pom.xml
```diff
@@ -25,6 +25,11 @@
       <artifactId>junit</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.fabric8</groupId>
```
jimma, line 29: Just some minor thing. Can we add this dependency under kubernetes/pom.xml instead of adding to each sub modules ?

jimma (approved):

kubernetes/ftest-kubernetes-logs/pom.xml
fabiobrz: @jimma the dependency is needed by 5 out of 9 child projects, and we don't really want for others that don't need it to inherit it, e.g.: the `istio` module should not depend on this artifact as they don't need it. The dependency management is already defined by the root POM, so in the end I don't think we should bring it up.
Do you agree?

fabiobrz:
I am merging this, based on the approval and related discussion with @jimma about the pending comment, which has been agreed upon to be marked as resolved.

PR #1339, opened 2024-12-15
Update OpenShift integration and restore itegration tests

### LINKED ISSUE #1323: Make OpenShift integration tests green again, and resume the related CI execution
##### Issue Overview

Resume OpenShift integration tests.

##### Expected Behaviour

The tests that verify the Arquillian Cube OpenShift integration should pass, and the CI script should be fixed in order to resume the execution of such tests.

##### Current Behaviour

The Arquillian Cube integration tests for OpenShift are failing, and the CI doesn't execute them, see https://github.com/arquillian/arquillian-cube/blob/main/.github/workflows/maven.yml#L24

##### Steps To Reproduce
1. 
```
mvn clean package -Dfailsafe.groups=org.arquillian.cube.openshift.impl.requirement.RequiresOpenShift -Dcontainerless.skip.tests=true
``` 

##### Additional Information

As a first step, it can be evaluated whether it would be enough to start with a CI configuration that uses OpenShift 3, which can be executed by runners, see https://github.com/fabric8io/kubernetes-client/blob/main/.github/workflows/e2e-tests.yml#L127

fabiobrz:
Restore OpenShift integration tests

See the commit list

Fixes #1323


fabiobrz:
Hi @jimma - could you please review (once you're back, of course :slightly_smiling_face:... )

jkube-maven-plugin-build/src/test/resources/spring-boot/pom.xml
```diff
@@ -0,0 +1,224 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+    Inspired by https://github.com/eclipse-jkube/jkube/tree/master/quickstarts/maven/spring-boot, copyright notice follows:
+
+    Copyright (c) 2019 Red Hat, Inc.
```
gaol, line 5: it is 2025 now :) The similar for others as well.

jkube-maven-plugin-build/src/test/resources/spring-boot/src/main/java/org/eclipse/jkube/maven/sample/Application.java
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright (c) 2019 Red Hat, Inc.
```
gaol, line 2: it is 2025 now :) The similar for others as well.

gaol (approved):
I am good with the changes.

Some trivial comments on the year in file head. :)

fabiobrz:
> I am good with the changes.
> 
> Some trivial comments on the year in file head. :)

Thanks for your review @gaol! 
I've had a look to the copyright year for those files and I can confirm I left them untouched intentionally, since I copy/pasted most of them from https://github.com/eclipse-jkube/jkube/tree/master/quickstarts/maven/spring-boot
Anyway, I have pushed one latest change to duplicate the initial comment I added already to the POM, so it is in the Java classes as well, see

fabiobrz:
I am going to merge this, according to the approval received from @gaol, since we do need to have full CI checks running on future changes. 
Just [a couple of comment lines were added](https://github.com/arquillian/arquillian-cube/pull/1339/commits/2b223da1e07739325934cfde2b78a81205d79d37) after the approval, to follow up on the review comments, therefore no functional code has been changed.

PR #1340, opened 2025-01-04
Bump commons-compress to 1.21, to fix security issues, and related commons-* dependencies to align with docker-java:3.4.0

fabiobrz:
Duplicates #1325 but using 1.27.1, since the one proposed by dependantbot is throwing a NoClassFound exception. 

See #1325 

- commons-compress from 1.19 to 1.27.1

Fixes https://github.com/arquillian/arquillian-cube/security/dependabot?q=package%3Aorg.apache.commons%3Acommons-compress+manifest%3Acore%2Fpom.xml+has%3Apatch

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -107,17 +107,17 @@
     <dependency>
       <groupId>commons-io</groupId>
       <artifactId>commons-io</artifactId>
-      <version>2.14.0</version>
+      <version>2.13.0</version>
     </dependency>
     <dependency>
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-compress</artifactId>
-      <version>1.27.1</version>
+      <version>1.21</version>
     </dependency>
     <dependency>
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-lang3</artifactId>
-      <version>3.9</version>
+      <version>3.12.0</version>
     </dependency>
     <dependency>
       <groupId>org.arquillian.cube</groupId>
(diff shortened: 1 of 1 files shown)


fabiobrz:
> ~The CI looks like does not pick up the latest dependency of commons-lang3:3.16.0, might relate to the cache setup in the github action ?~
> 
> OK, please ignore above comment.
> 
> I clicked the `re-run jobs` button and from https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/re-running-workflows-and-jobs#about-re-running-workflows-and-jobs, it might use the same commit.

> It looks that the update of the pull request does not trigger the action, maybe add `types: [synchronize]` to the workflow ? see: https://github.com/orgs/community/discussions/25796

Hi @gaol - and thanks for your feedback. It seems to me the error is actually related to the fact that `docker-java:3.4.0` would use `commons-compress:1.19.0` and then `commons-lang3:3.12.0` transitively, which in turn conflicts with the latest updates here.
I'll try to experiment a bit and push some changes, so that we're sure the GitHub workflow will run on latest commit.
Let's see how it goes...

fabiobrz:
Hi @gaol - it seems this is the less invasive solution, i.e. to align with the set of `commons-*` that `docker-java:3.4.0` is transitively using as part of the `docker-java-core:3.4.0` direct dependencies.
CI checks are green, would you mind reviewing and merging in case you're fine with the changes?

PR #1342, opened 2025-01-08
Fix resources structure in kubernetes/fabric8

gaol:
Fix the resources structure in kubernetes/fabric8:
The `kubernetes/fabric8/src/main/java/resources` should be at `kubernetes/fabric8/src/main/resources`


fabiobrz (approved):
Thanks @gaol! This LGTM, merging.

PR #1353, opened 2025-02-06
Fix the build module (and submodule) POM coordinates

fabiobrz:
SSIA

- Fix the version in `build` module and submodule POMs

Fixes #


fabiobrz:
Merging right away since this is just an update to a module that builds documentation, and checks are still green.

PR #1363, opened 2025-04-03
Wildfly + JKube on Kubernetes example

### LINKED ISSUE #1365: Implement a test to demonstrate effective WildFly testing on Kubernetes
##### Issue Overview

Implement a test to demonstrate effective WildFly testing on Kubernetes, covering all the aspects of the process

##### Expected Behaviour

An integration test should demonstrate how a regular WildFly application can be deployed to Kubernetes, and tested. 

##### Current Behaviour

There are some WildFly related tests, but none of them is covering all the aspects of the Kubernetes deployment process.

##### Steps To Reproduce
N/A

##### Additional Information

The solution could leverage the JKube Maven plugin integration, and be similar to one [Spring Boot test application that is implemented already](https://github.com/arquillian/arquillian-cube/tree/main/jkube-maven-plugin-build/src/test/resources/spring-boot).

fabiobrz:
SSIA

- Add a WildFly test for Kubernetes, using the Arquillian Cube JKube integration

Fixes #1365

### DIFF
--- a/jkube-maven-plugin-build/src/main/java/org/eclipse/jkube/maven/plugin/build/JKubeMavenPluginResourceGeneratorBuilder.java
+++ b/jkube-maven-plugin-build/src/main/java/org/eclipse/jkube/maven/plugin/build/JKubeMavenPluginResourceGeneratorBuilder.java
@@ -18,7 +18,7 @@ public class JKubeMavenPluginResourceGeneratorBuilder {
     private static final Logger logger = Logger.getLogger(JKubeMavenPluginResourceGeneratorBuilder.class.getName());
 
     private Path pom;
-    private String[] k8sGoals = new String[] {"package", "k8s:build", "k8s:resource", "k8s:deploy"};
+    private String[] k8sGoals = new String[] {"package", "k8s:build", "k8s:resource", "k8s:push", "k8s:apply"};
     private String[] openshiftGoals = new String[] {"package", "oc:build", "oc:resource"};
     private String namespace;
     private boolean mvnDebugOutput;
@@ -35,8 +35,13 @@ public class JKubeMavenPluginResourceGeneratorBuilder {
         return this;
     }
 
-    public JKubeMavenPluginResourceGeneratorBuilder goals(String[] goals) {
+    public JKubeMavenPluginResourceGeneratorBuilder k8sGoals(String[] goals) {
         this.k8sGoals = goals;
+        return this;
+    }
+
+    public JKubeMavenPluginResourceGeneratorBuilder openshiftGoals(String[] goals) {
+        this.openshiftGoals = goals;
         return this;
     }
 
--- a/kubernetes/pom.xml
+++ b/kubernetes/pom.xml
@@ -27,7 +27,6 @@
     <module>ftest-kubernetes-reporter</module>
     <module>ftest-kubernetes-assistant</module>
     <module>ftest-kubernetes-resources</module>
-    <module>ftest-kubernetes-wildfly-bootable-jar</module>
   </modules>
 
   <profiles>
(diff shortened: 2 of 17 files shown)


fabiobrz:
@gaol - please don't care about failing OpenShift CI checks here, which are due to Ubuntu 20 deprecation in GitHub runners, and that we will address separately.
Plus these changes are specifically about k8s, so not related to OpenShift.

kubernetes/pom.xml
```diff
@@ -29,4 +29,25 @@
...
+          <module>fabric8</module>
+          <module>kubernetes</module>
+          <module>istio</module>
+          <module>reporter</module>
+          <module>ftest-kubernetes</module>
+          <module>ftest-kubernetes-logs</module>
+          <module>ftest-kubernetes-reporter</module>
+          <module>ftest-kubernetes-assistant</module>
+          <module>ftest-kubernetes-resources</module>
+          <module>ftest-kubernetes-wildfly-bootable-jar</module>
```
gaol, line 48: I think we can only have this new module here, the other modules are enabled by default in any profile.

jkube-maven-plugin-build/src/main/java/org/eclipse/jkube/maven/plugin/build/JKubeMavenPluginResourceGeneratorBuilder.java
```diff
@@ -18,7 +18,7 @@ public class JKubeMavenPluginResourceGeneratorBuilder {
     private static final Logger logger = Logger.getLogger(JKubeMavenPluginResourceGeneratorBuilder.class.getName());
 
     private Path pom;
-    private String[] k8sGoals = new String[] {"package", "k8s:build", "k8s:resource", "k8s:deploy"};
+    private String[] k8sGoals = new String[] {"package", "k8s:resource", "k8s:build", "k8s:push", "k8s:deploy"};
```
gaol, line 21: May I know the reason of changing the order of `k8s:resource` and `k8s:build` ?

Another comment may not relate to this PR, but not sure if `package` should be activated by default, I see the wildfly-jar-plugin generates the bootable jar 3 times in the log. The 3rd one may come from `k8s:deploy`, maybe consider to `k8s:apply` ?

gaol (approved):
@fabiobrz LGTM :+1:  with some trivial comments

I was checking the wildfly:35.0.1.Final usage last night which actually is requiring JDK17+, but I see the new `ftest-kubernetes-wildfly-bootable-jar` module is enabled only on `<jdk>[17,)</jdk>`, so it is good to me.

kubernetes/pom.xml
fabiobrz: Thanks, fixed!

jkube-maven-plugin-build/src/main/java/org/eclipse/jkube/maven/plugin/build/JKubeMavenPluginResourceGeneratorBuilder.java
fabiobrz: > May I know the reason of changing the order of k8s:resource and k8s:build ?

It seemed to me it was resembly the actual plugin workflow, but I've reverted it now.

> Another comment may not relate to this PR, but not sure if package should be activated by default, I see the wildfly-jar-plugin generates the bootable jar 3 times in the log. The 3rd one may come from k8s:deploy, maybe consider to k8s:apply ?

Yes, apply is actually the right choice if `k8s:build` and `k8s:resource` are executed previously, hence I reverted this change as well.
Regarding `package`, instead, IIUC it's better to have it in the defaults because it's not implied that the required build would happen for sure first.

fabiobrz:
> @fabiobrz LGTM 👍 with some trivial comments

Thanks @gaol - I'ìve pushed some fixes in order to address your comments. Feel free to let me know what you think

gaol (approved):

PR #1373, opened 2025-05-07
Adding a new WildFly Bootable JAR on OpenShift test

### LINKED ISSUE #1374: Create a WildFly Bootable JAR + OpenShift functional test
##### Issue Overview

Similar to what has been done for a WIldFly on K8s, it would be good to have an OpenShift integration test that verifies how a WildFly Bootable JAR can be easily deployed and tested with Arquillian Cube.

##### Expected Behaviour

Have a functional test which is run along with OpenShift related tests.

##### Current Behaviour

There are no WidlFly Bootable JAR tests which demonstrate automatic provisioning and test of a modern WildFly application.

##### Steps To Reproduce
1. e.g.: `mvn verify -B -Dfailsafe.groups=org.arquillian.cube.openshift.impl.requirement.RequiresOpenshift -Dcontainerless.skip.tests=true`, and ssee that no such tests exist

##### Additional Information

Use JKube integration and a similar approach to the [WildFly + K8s test](https://github.com/arquillian/arquillian-cube/pull/1363)

fabiobrz:

SSIA

- A new functional test to verify how a modern WildFly Bootable JAR can be deployed and tested on OpenShift 
- A fix to the WildFly Bootable JAR k8s test artifact id

Fixes #1374

### DIFF
--- a/openshift/ftest-openshift-wildfly-bootable-jar/src/test/java/org/arquillian/cube/kubernetes/WildflyBootableJarOpenShiftIT.java
+++ b/openshift/ftest-openshift-wildfly-bootable-jar/src/test/java/org/arquillian/cube/kubernetes/WildflyBootableJarOpenShiftIT.java
@@ -1,0 +1,77 @@
+package org.arquillian.cube.kubernetes;
+
+import io.fabric8.kubernetes.api.model.ObjectMeta;
+import io.fabric8.kubernetes.api.model.Service;
+import io.fabric8.kubernetes.api.model.ServiceList;
+import io.fabric8.openshift.client.OpenShiftClient;
+import io.restassured.RestAssured;
+import io.restassured.response.Response;
+import org.apache.http.HttpStatus;
+import org.arquillian.cube.openshift.impl.enricher.RouteURL;
+import org.arquillian.cube.openshift.impl.requirement.RequiresOpenshift;
+import org.arquillian.cube.requirement.ArquillianConditionalRunner;
+import org.assertj.core.api.Assertions;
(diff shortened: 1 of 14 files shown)


openshift/ftest-openshift-wildfly-bootable-jar/src/test/java/org/arquillian/cube/kubernetes/WildflyBootableJarOpenShiftIT.java
```diff
@@ -0,0 +1,75 @@
...
+            .extracting(Service::getMetadata)
+            .extracting(ObjectMeta::getName)
+            .containsExactlyInAnyOrder("hello-world-svc");
+    }
+
+    @Test
+    @InSequence(2)
+    public void helloEndpointShouldReplyWithHttp200() throws InterruptedException {
+        // wait until one route is actually ready, 60 seconds at most currently works for CI...
+        Thread.sleep(60 * 1000);
```
gaol, line 64: Maybe it would be better to use the `Awaitility.await()` API to check on conditions until 60 seconds timeout instead of sleeping for 60 seconds before the check.

gaol (approved):
LGTM with trivial comments. :)

fabiobrz:
> LGTM with trivial comments. :) 

Hi @gaol, I addressed your comments, and BTW added a couple of commits to fix two tests that started failing during the last two days, most likely due to GutHub runners specifics around networking. Not a big deal but feel free to check and let me know what you think. 
CI checks are all green now! 🙂

PR #1398, opened 2025-08-26
[issue 1397] - Adding support to mount volumes with free string configuration option

### LINKED ISSUE #1397: Docker: `ContainerDsl`, `ContainerDslRule` and `ContainerBuilder` don't support adding volumes with the SELinux labeling configuration option
**Describe the bug**
`ContainerDsl`, `ContainerDslRule` and `ContainerBuilder` provide a `withVolume` method which only accepts the BindMode enumeration to set a volume mount binding options. 
Such implementation does not allow using the `z` or `Z` options to configure SELinux labeling, see https://docs.docker.com/engine/storage/bind-mounts/#options-for---volume and ultimately results in a failure when trying to start a container which expects a volume mounted with such option, as for instance a PostgreSql container. 

**To Reproduce**
1. Inject a `@DockerContainer` which expects a volume to be mounted with the `Z` option, e.g.:
```
@DockerContainer
    Container postgres = Container.withContainerName("postgres")
            .fromImage("quay.io/centos7/postgresql-13-centos7:centos7")
            .withPortBinding("5432->5432")
            .withEnvironment("POSTGRESQL_DATABASE", POSTGRESQL_DATABASE) // creates POSTGRESQL_DATABASE database after 1st start
            .withEnvironment("POSTGRESQL_USER", POSTGRESQL_USER)
            .withEnvironment("POSTGRESQL_PASSWORD", POSTGRESQL_PASSWORD)
            .withVolume(postgresDataDir.getAbsolutePath(), "/var/lib/pgsql/data:Z")
            .withAwaitStrategy(postgresContainerAwaitStrategy())
            .build();
``` 
2. Run the test and see it fail with Arquillian Cube failing at starting the container because of a parsing error of the volume mount which basically turns out to be as follows:
`<host_volume_path>:/var/lib/pgsql/data:Z:rw`

**Expected behavior**
It should be possible to call `.withVolume(postgresDataDir.getAbsolutePath(), "/var/lib/pgsql/data:Z")` without having an additional option - i.e. `rw` - added by default and the container should start.

**Screenshots**
N/A

**Additional context**
This behavior is not aligned with what is done by  [DockerContainerObjectBuilder](https://github.com/arquillian/arquillian-cube/blob/main/docker/docker/src/main/java/org/arquillian/cube/docker/impl/client/containerobject/DockerContainerObjectBuilder.java#L440) which instead is adding the `:Z` option by default when binding volumes.

fabiobrz:
Changes allow to set a free form bind mount configuration option when creating a docker container so that - for example - `:Z` can be set for container that run on SELinux.

- Adding an overloaded version of `withVolume(String, String, String)` to ContainerBuilder, ContainerDsl and ContainerDslRule to allow setting free form option for volume mounts.

Fixes #1397

### DIFF
--- a/docker/ftest-docker-containerobject-dsl/src/test/java/org/arquillian/cube/containerobject/dsl/PostgreSqlIT.java
+++ b/docker/ftest-docker-containerobject-dsl/src/test/java/org/arquillian/cube/containerobject/dsl/PostgreSqlIT.java
@@ -1,0 +1,102 @@
+package org.arquillian.cube.containerobject.dsl;
+
+import org.arquillian.cube.docker.impl.await.StaticAwaitStrategy;
+import org.arquillian.cube.docker.impl.client.config.Await;
+import org.arquillian.cube.docker.impl.client.containerobject.dsl.Container;
+import org.arquillian.cube.docker.impl.client.containerobject.dsl.DockerContainer;
+import org.arquillian.cube.docker.impl.requirement.RequiresDocker;
+import org.arquillian.cube.docker.impl.util.OperatingSystemFamily;
+import org.arquillian.cube.docker.impl.util.OperatingSystemResolver;
+import org.arquillian.cube.requirement.ArquillianConditionalRunner;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.attribute.PosixFilePermissions;
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Verify that {@link DockerContainer} can be used to define a valid {@code PosatgreSql} Docker container.
+ * <p>
+ *     A {@link DockerContainer} annotation is used to define a PostgreSql Docker container.
+ *     A local directory is created on the host to hold the PostgreSql data and to be mounted as a volume on the
+ *     Docker container.
+ *     The data directory is created statically, so that it can be used by the {@link DockerContainer} definition later
+ *     in the test lifecycle.
+ *     The {@code Z} bind mode option is passed so that it can be used by SELinux enabled systems, see the related
+ *     <a href="https://docs.docker.com/engine/storage/bind-mounts/#configure-the-selinux-label">Docker documentation</a>,
+ *     otherwise the container would fail to start with the following error:<br>
+ *     {@code mkdir: cannot create directory '/var/lib/pgsql/data/userdata': Permission denied}
+ * </p>
+ */
+@Category({ RequiresDocker.class})
+@RunWith(ArquillianConditionalRunner.class)
+public class PostgreSqlIT {
+    static final String POSTGRESQL_USER = "user";
+    static final String POSTGRESQL_PASSWORD = "pass";
+    static final String POSTGRESQL_DATABASE = "test-database";
+    static final String POSTGRESQL_IMAGE = "quay.io/centos7/postgresql-13-centos7:centos7";
+    static final int POSTGRESQL_HOST_PORT_BINDING = 5432;
+    static final String POSTGRESQL_CONTAINER_DATA_DIRECTORY_PATH = "/var/lib/pgsql/data";
+    static final File POSTGRESQL_HOST_DATA_DIRECTORY;
+
+    static {
+        try {
+            POSTGRESQL_HOST_DATA_DIRECTORY = java.nio.file.Files.createTempDirectory(null).toFile();
+        } catch (IOException e) {
+            throw new IllegalStateException("Cannot create PostgreSql data dir", e);
+        }
+    }
+
+    @BeforeClass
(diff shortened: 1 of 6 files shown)


docker/ftest-docker-junit-rule/src/test/java/org/arquillian/cube/docker/junit/rule/PostgreSqlIT.java
```diff
@@ -0,0 +1,99 @@
...
+        }
+    }
+
+    private static File createPostgresDataDir() throws IOException {
+        // create local data dir for postgres database
+        File postgresDataDir = java.nio.file.Files.createTempDirectory(null).toFile();
+        // This is needed to run on SELinux enabled Docker hosts
+        // otherwise the following error will prevent the container from starting successfully:
+        //   mkdir: cannot create directory '/var/lib/pgsql/data/userdata': Permission denied
+        java.nio.file.Files.setPosixFilePermissions(Path.of(postgresDataDir.toURI()),
```
gaol, line 60: Shall we check the OS before calling `setPosixFilePermissions` in case running on Windows ?

docker/ftest-docker-containerobject-dsl/src/test/java/org/arquillian/cube/containerobject/dsl/PostgreSqlIT.java
```diff
@@ -0,0 +1,97 @@
...
+            throw new IllegalStateException("Cannot create PostgreSql data dir", e);
+        }
+    }
+
+    @BeforeClass
+    public static void configurePostgresDataDir() throws IOException {
+        // This is needed to run on SELinux enabled Docker hosts
+        // otherwise the following error will prevent the container from starting successfully:
+        //   mkdir: cannot create directory '/var/lib/pgsql/data/userdata': Permission denied
+        java.nio.file.Files.setPosixFilePermissions(Path.of(POSTGRESQL_HOST_DATA_DIRECTORY.toURI()),
```
gaol, line 60: Shall we check the OS before calling `setPosixFilePermissions` in case running on Windows  ?

docker/ftest-docker-junit-rule/src/test/java/org/arquillian/cube/docker/junit/rule/PostgreSqlIT.java
```diff
@@ -0,0 +1,99 @@
...
+ *     Docker container.
+ *     The data directory is created statically, so that it can be used by the {@link ContainerDslRule} definition later
+ *     in the test lifecycle.
+ *     The {@code Z} bind mode option is passed so that it can be used by SELinux enabled systems, see the related
+ *     <a href="https://docs.docker.com/engine/storage/bind-mounts/#configure-the-selinux-label">Docker documentation</a>,
+ *     otherwise the container would fail to start with the following error:<br>
+ *     {@code mkdir: cannot create directory '/var/lib/pgsql/data/userdata': Permission denied}
+ * </p>
+ */
+@Category({ RequiresDocker.class})
```
gaol, line 36: Shall we add `@RunWith(ArquillianConditionalRunner.class)` to make the `RequiresDocker` work as expected or am I missing something ?

gaol (commented):
Overall it looks good to me :+1: 

I have some comments though on the test codes just in case my concerns are valid :)

docker/ftest-docker-junit-rule/src/test/java/org/arquillian/cube/docker/junit/rule/PostgreSqlIT.java
fabiobrz: It is not always needed, as in this case, i.e. when using `@ClassRule` with `ContainerDslRule', see https://arquillian.org/arquillian-cube/#_junit_rule

docker/ftest-docker-containerobject-dsl/src/test/java/org/arquillian/cube/containerobject/dsl/PostgreSqlIT.java
fabiobrz: Thanks, fixed.

docker/ftest-docker-junit-rule/src/test/java/org/arquillian/cube/docker/junit/rule/PostgreSqlIT.java
fabiobrz: Thanks, fixed.

fabiobrz:
> Overall it looks good to me 👍
> 
> I have some comments though on the test codes just in case my concerns are valid :)

Thanks for your review @gaol, I've addressed your comments. Feel free to let me know if there's anything else you'd wish to discuss.

docker/ftest-docker-junit-rule/src/test/java/org/arquillian/cube/docker/junit/rule/PostgreSqlIT.java
gaol: Ah, yes. that is the thing I am missing :) thank you .

Issue #1409, opened 2025-09-30
Review GitServer implementation


fabiobrz:
##### Issue Overview

Arquillian Cube GitServer.java is now using Gitea. 
This causes issues with the project CI checks that use CRC and often fail at deploying the Gitea operator.

##### Expected Behaviour
The HelloPodDeploymentOpenShiftITCase should be executed by the project CI and pass

##### Current Behaviour
The HelloPodDeploymentOpenShiftITCase is skipped

##### Steps To Reproduce
1. Just have a look to one execution of e2e CI checks

##### Additional Information
The logic that the test uses could be fully revisited, thus removing the need for a remote Git service integration

PR #1411, opened 2025-10-02
Fix HelloPodDeploymentOpenShiftITCase by removing Gitea dependency

striver-24:
The `HelloPodDeploymentOpenShiftITCase` was being skipped due to Gitea operator deployment failures in CI environments using CRC (CodeReady Containers). The test was marked with `@Ignore("Needs the Gitea operator, which Arquillian Cube tries to provision, but it does not work on GitHub CI which uses CRC")`.

This PR eliminates the Gitea dependency by switching from build-from-source to a pre-built image approach while preserving all original testing functionality.

Fixes #
- Remove @Ignore annotation to allow test execution
- Replace build-from-source approach with pre-built WildFly image
- Remove RequiresRemoteResource dependency
- Add runtime admin user creation instead of build-time setup
- Enhance readiness probe to validate WildFly management API
- Expose management port (9990) alongside application port (8080)

### DIFF
--- a/openshift/ftest/src/test/java/org/arquillian/cube/openshift/ftest/HelloPodDeploymentPreBuiltOpenShiftITCase.java
+++ b/openshift/ftest/src/test/java/org/arquillian/cube/openshift/ftest/HelloPodDeploymentPreBuiltOpenShiftITCase.java
@@ -4,30 +4,29 @@ import java.net.URL;
 
 import org.arquillian.cube.olm.impl.requirement.RequiresOlm;
 import org.arquillian.cube.openshift.impl.requirement.RequiresOpenshift;
-import org.arquillian.cube.remote.requirement.RequiresRemoteResource;
 import org.arquillian.cube.requirement.ArquillianConditionalRunner;
 import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.container.test.api.TargetsContainer;
 import org.jboss.arquillian.test.api.ArquillianResource;
 import org.jboss.shrinkwrap.api.ShrinkWrap;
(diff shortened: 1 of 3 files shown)


fabiobrz:
Hi @striver-24 - and thanks for contributing to Arquillian Cube! :heart: 

Just one general note: this PR is not providing a different approach for Cube to deploy and use a Git service, which is still a valid use case. 
That is not an issue, given that #1409 suggests that the test can be fixed by removing the need for a Git service integration altogether.
Once we choose to do so, though, we need to keep a test that validates the build from source approach, so could you please duplicate the test case and apply your solution to the new one?

fabiobrz:
Awesome, thanks @striver-24!

fabiobrz:
Hi @striver-24 - I've tried running the test locally and it does not pass. Could you try as well and confirm?

fabiobrz:
> Hi @striver-24 - I've tried running the test locally and it does not pass. Could you try as well and confirm?

Basically what happens here is that Cube will not take the `container qualifier="hello-openshift-prebuilt"` into account, thus not start it, because it should match the name of the default container, i.e. `container qualifier="hello-openshift" default="true"`.
One valid configuration for your new test to run successfully would be to place it to a new, separate submodule, let's say `ftest-prebuilt`.

fabiobrz (changes requested):
Some minor comments.

fabiobrz:
Hi @striver-24 - doing some cleanup :slightly_smiling_face: 
Do you have any plans for the changes here to be wrapped up? 
Just feel free to let me know if that is not the case, and I'll finish the work you started.

